PR #15, opened 2015-10-12
Add IndexCounter

LuchenTan:
Add IndexCounter to generate a list of all indexed document Ids


LuchenTan:
Add IndexCounter, output all document IDs in an index

PR #29, opened 2015-10-24
Fixed breaks of DumpDocids

LuchenTan:


LuchenTan:
Added docIdName parameter, and modified to LOG.info and LOG.error

PR #35, opened 2015-10-30
Mock TwitterStream to read tweets from stored files

xeniaqian94:
A stable previous stored tweets collection might be good for random trial on https://github.com/lintool/Anserini/issues/31, which redirects to this issue https://github.com/lintool/Anserini/issues/20

Does this issue mean: either real-time tweets or obsolete tweets will become an option that user could configure?

For fluent API, Twitter4j's (interface) TwitterStream was used before as 
<pre><code>
TwitterStream twitterStream = new TwitterStreamFactory().getInstance();
RawStreamListener rawListener = new RawStreamListener(){
    ...
    }

twitterStream.addListener(rawListener);
twitterStream.sample();
</code></pre>

Might be desirable to have some class that encapsulate both TwitterStream and MockStream? Would like some ideas on fluent API tomorrow's hackathon if possible.


lintool:
@xeniaqian94 I'm working on this issue, but let's chat f2f.
This commit has a lot of clutter on it, e.g., adding and then removing file (they're preserved in the history). I'd like to keep the commit history as clean as possible... so it might be better if you just clean this up; easiest is to probably kill this branch and start over.

PR #37, opened 2015-10-31
Might help mock stream with Tweets 2011

xeniaqian94:
Might help with issue https://github.com/lintool/Anserini/pull/35. Now TweetStreamIndexer reads from stored tweets2011 Json file. Or could simply close this pull request if there's duplicate effort, this commit is still good for understanding & rebase git cmd practice purpose.


lintool:
@xeniaqian94 your master commit 056f6bb7003a05d213528625676bcd0c31ec1dff contains a 18 MB data file src/main/resources/storedTweets/20110123-001.json which doesn't really belong...

The issue with Git is that once you check something in, it's permanently in history, even if you remove it. You then merge master into the RESTful-API branch, which "pollutes" it...

So, I would suggest "winding back" to https://github.com/xeniaqian94/Anserini/commit/c6382fd06d6d7a25f61cfa5647d61d6a48d26b82 which is the last "good" commit. Or alternatively, you could start over with master, pull Adam's initial REST implementation, and go from there... make sense?

PR #43, opened 2015-11-03
RESTful Client API

xeniaqian94:
RESTful Client API. Issue https://github.com/lintool/Anserini/issues/24

Besides, TweetStreamIndexer can now be registered with its own IndexWriter, Using TweetSearcher's static field indexWriter(not instantiated) can cause problem when client runs alone.


lintool:
Closing out of date PR.

PR #56, opened 2016-02-17
Feature extractor for Gov2 collection

dyshi:
Added feature extractor for Gov2 and Twitter collection that will compute the feature vector for each query and document


dyshi:
merged into https://github.com/lintool/Anserini/pull/59

PR #58, opened 2016-02-19
Adding the cacm collection

dyshi:
Compressed the cacm indexer and searcher into one commit


dyshi:
merged into https://github.com/lintool/Anserini/pull/59

PR #64, opened 2016-06-01
Just changes to IndexWebCollection

dyshi:


src/main/java/io/anserini/index/IndexWebCollection.java
```diff
@@ -101,15 +117,21 @@ private int indexWarcRecord(WarcRecord warcRecord) throws IOException {
...
+      if (positions) {
+          // Important, lucene 5 no longer has simple setIndexed option
+          // set through index options
+          fieldType.setIndexOptions(IndexOptions.DOCS_AND_FREQS_AND_POSITIONS);
+          fieldType.setStoreTermVectorPositions(true);
+          document.add(new Field(FIELD_BODY, contents, fieldType));
+      } else {
+          fieldType.setIndexOptions(IndexOptions.DOCS_AND_FREQS);
+          document.add(new Field(FIELD_BODY, contents, fieldType));
+      }
```
lintool: I'm not sure I get the logic here. Previously, there were `-positions` and `-docvectors` options. Here, it seems that you're always storing the term vectors? Why can't we just use `TermVectorsTextField`?

dyshi: Is that also in lucene/document? it doesn't appear to be in 5.3. I remember reading that in 5.3 we need to specify to store term vectors

lintool: Okay, but the three options we want to support are:
+ no args (default): no positions, use `NoPositionsTextField`, no storing term vectors
+ `-positions`: store positions, but not term vectors
+ `-positions -docvectors`: store positions and term vectors

Can you please change the code to keep the old behavior?

iorixxx: > we cannot store positions without term vectors

Please see stock [TextField] (https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/document/TextField.java) that stores positions without term vectors.

dyshi: Oh thanks! Read the field options and thought that was all there was. Also found the NoPositionsTextField in the Anserini's Index package.
I remember though at the beginning when we were trying to use term vectors only specifying through the IndexOption actually stored the tf files. Am I mixing up some concepts?

I've updated the code.

iorixxx: > Am I mixing up some concepts?

There are two positions here. Positions required for the PhraseQuery, SpanNearQuery, proximity search as in the positional inverted index. The other is specific to term vectors: FieldType#setStoreTermVectorPositions

dyshi: Oh I see, so storing the position information in the index is separate from storing positions again in the term frequencies? Thanks!

lintool: @dyshi So, it seems like the original code should do what you need if you use `-positions -docvectors`?

This particular class has a lot of dependencies - in fact, all other experiments depend on it... I don't want changes to regress and break existing experiments...

lintool:
Hi @dyshi I still don't think your new patch implements the correct logic:

```
      if (positions && docVectors) {
          // Important, lucene 5 no longer has simple setIndexed option
          // set through index options
          fieldType.setStored(true);
          fieldType.setStoreTermVectors(true);
          fieldType.setIndexOptions(IndexOptions.DOCS_AND_FREQS_AND_POSITIONS);
          fieldType.setStoreTermVectorPositions(true);
          document.add(new Field(FIELD_BODY, contents, fieldType));
      } else if (!positions && docVectors){
        document.add(new NoPositionsTextField(FIELD_BODY, contents));
      } else if (positions && !docVectors) {
          document.add(new TextField(FIELD_BODY, contents, Field.Store.YES));
      } else {
          document.add(new TextField(FIELD_BODY, contents));
      }
```

For example, the `!positions && docVectors` doesn't actually store the doc vectors?
Can you please make the semantics of each of the branch match the options and build test indexes, say, on Gov2, for all conditions?

lintool:
Okay, thanks - I'm testing out your code right now.

PR #73, opened 2016-10-14
Add DumpIndex util to easily print index info

Peilin-Yang:
The purpose:

1. Let the user print the index info more easily. This could be used if the user wants to see the stats of the index in order to confirm the correctness of indexing process, e.g. number of docs in the index.
2. Indri has dumpindex util which is convenient. People would be more willing to migrate to Anserini if we have the same util.

The stats include:

- index stats
- term counts
- document dump (find id, find name, print text, print vector)

### DIFF
--- a/src/main/java/io/anserini/util/DumpIndex.java
+++ b/src/main/java/io/anserini/util/DumpIndex.java
@@ -1,0 +1,70 @@
+package io.anserini.util;
+
+import io.anserini.index.IndexWebCollection;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.lucene.index.DirectoryReader;
+import org.apache.lucene.store.FSDirectory;
+
+import java.io.File;
+import java.io.IOException;
+
+public class DumpIndex {
+  private static final Logger LOG = LogManager.getLogger(DumpIndex.class);
+
+  String print_repository_stats( String indexDir ) throws IOException {
+    FSDirectory dir = FSDirectory.open(new File(indexDir).toPath());
+    DirectoryReader reader = DirectoryReader.open(dir);
+
+    int docCount = reader.numDocs();
+    int contentsCount = reader.getDocCount(IndexWebCollection.FIELD_BODY);
+    long termCount = reader.getSumTotalTermFreq(IndexWebCollection.FIELD_BODY);
+
+    StringBuilder sb = new StringBuilder();
+    sb.append("Repository statistics:\n");
+    sb.append("documents:\t" + docCount + "\n");
+    sb.append("contentsCount(doc with contents):\t" + contentsCount + "\n");
+    //sb.append("unique terms:\t" + uniqueTermCount);
+    sb.append("total terms:\t" + termCount + "\n");
+    sb.append("fields:\t\t" + IndexWebCollection.FIELD_ID + " " + IndexWebCollection.FIELD_BODY);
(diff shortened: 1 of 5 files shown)


src/main/java/io/anserini/util/DumpIndex.java
```diff
@@ -0,0 +1,221 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+public class DumpIndex {
+  private static final Logger LOG = LogManager.getLogger(DumpIndex.class);
+
+  class RawDocNotStoredException extends Exception {
```
lintool: Would it make sense to make these top-level classes under a new package?

src/main/java/io/anserini/util/DumpIndex.java
```diff
@@ -0,0 +1,221 @@
...
+    Document d = reader.document(1);
+    List<IndexableField> fields = d.getFields();
+    for (IndexableField f : fields) {
+      sb.append(f.name()+" ");
+    }
+    sb.append("\n");
+    return sb.toString();
+  }
+
+  String print_term_counts( DirectoryReader reader, String termStr )
```
lintool: This and below also...

src/main/java/io/anserini/util/DumpIndex.java
```diff
@@ -0,0 +1,221 @@
...
+      .append("\n");
+    return sb.toString();
+  }
+
+  /*
+  * print the internal id
+  */
+  String print_document_id( DirectoryReader reader, ArrayList<String> externalIDs ) throws IOException {
+    StringBuilder sb = new StringBuilder();
+    HashMap<String, Integer> mapping = new HashMap<>();
```
lintool: Just `Map<String, Integer> mapping`: no need to specify implementation in declaration.

src/main/java/io/anserini/util/DumpIndex.java
```diff
@@ -0,0 +1,221 @@
...
+      .append("    term (t)             Term text      Print inverted list for a term\n")
+      .append("    documentid (di)      Original IDs   Print the internal document IDs of documents\n")
+      .append("    documentname (dn)    Internal Document ID    Print the text representation of a document ID\n")
+      .append("    documenttext (dt)    Internal Document ID    Print the text of a document\n")
+      .append("    documentvector (dv)  Internal Document ID    Print the document vector of a document\n")
+      .append("    stats (s)                           Print statistics for the Repository\n");
+    System.out.println(sb.toString());
+  }
+
+  public static void main(String[] clArgs) throws IOException{
```
lintool: Do you want to use args4j for cleaning arg parsing, like in the other classes?

src/main/java/io/anserini/util/DumpIndex.java
Peilin-Yang: yes u r right

src/main/java/io/anserini/util/DumpIndex.java
Peilin-Yang: At the beginning I tried to not to use args4j to make the code with less dependencies.
But I can migrate to args4j to make it more compatible with other codes in the repo.

lintool: Yes, please do - arg4j is pretty widely used and makes things a lot cleaner...

PR #76, opened 2016-10-17
add support for indexing trectext collections (disk1-5 + AQUAINT)

Peilin-Yang:
@lintool 
The code is mixed with IndexWebCollection. I think we'd better separate them some time in the future. 
Tested with disk1-5 and AQUAINT
**NOTE**: For AQUAINT, we only have the uncompressed version. I think the original files are gzipped? In the code I have this support, can you try that? 

Index Stats:

- DISK12

documents:      741687
contentsCount(doc with contents):       741686
total terms (with or without stop word):    311265071 / 222530202

- DISK45 (the CR dir is ignored in the code)

documents:      528030
contentsCount(doc with contents):       528030
total terms (with or without stop word):    251357430 / 174540587

- AQUAINT

documents:      1031455
contentsCount(doc with contents):       1031326
total terms (with or without stop word):    446995583 / 320157232

### DIFF
--- a/src/main/java/io/anserini/document/TrecTextRecord.java
+++ b/src/main/java/io/anserini/document/TrecTextRecord.java
@@ -1,0 +1,74 @@
+package io.anserini.document;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import io.anserini.index.IndexWebCollection;
+
+public final class TrecTextRecord {
+
+  public static final String DOCNO = "<DOCNO>";
+  public static final String TERMINATING_DOCNO = "</DOCNO>";
+
+  public static final String DOC = "<DOC>";
+  public static final String TERMINATING_DOC = "</DOC>";
+
+  public static final String[] startTags = {"<TEXT>", "<HEADLINE>", "<TITLE>", "<HL>", "<HEAD>",
+          "<TTL>", "<DD>", "<DATE>", "<LP>", "<LEADPARA>"
+  };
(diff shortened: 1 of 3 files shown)


src/main/java/io/anserini/document/TrecTextRecord.java
```diff
@@ -0,0 +1,74 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import io.anserini.index.IndexWebCollection;
+
+public final class TrecTextRecord {
```
lintool: Do you want to call this a `TrecDocument`?

src/main/java/io/anserini/document/TrecTextRecord.java
```diff
@@ -0,0 +1,74 @@
...
+  public static final String[] startTags = {"<TEXT>", "<HEADLINE>", "<TITLE>", "<HL>", "<HEAD>",
+          "<TTL>", "<DD>", "<DATE>", "<LP>", "<LEADPARA>"
+  };
+  public static final String[] endTags = {"</TEXT>", "</HEADLINE>", "</TITLE>", "</HL>", "</HEAD>",
+          "</TTL>", "</DD>", "</DATE>", "</LP>", "</LEADPARA>"
+  };
+
+  public static final int BUFFER_SIZE = 1 << 16; // 64K
+
+  public static WarcRecord parseTrecTextRecord(StringBuilder builder) {
```
lintool: I don't understand why you call this a WarcRecord... isn't it confusing with the actual web collections? These are newswire documents we're indexing, right?

lintool:
Do you have effectiveness numbers (AP, P30)?

src/main/java/io/anserini/document/TrecTextRecord.java
Peilin-Yang: I agree that this is confusing...
The reason probably is that indexing TrecRecord is exactly the same as indexing the WarcRecord after we get the contents. Trec record includes tags such as <TEXT> <HEADLINE> etc. We need jsoup to extract the pure text out of it. 
If I create another record -- TrecRecord, I will duplicate the indexWarcRecord for indexing the Trec record by only changing the function args. 
I think the better way to deal with this is we refactor the code so that different types of the records inherit a parent class. That means a lot of renaming though...
What do you think?

Peilin-Yang:
I've filed another issue for the performances.
Will do that soon.

src/main/java/io/anserini/document/TrecTextRecord.java
lintool: sgtm - I'm happy to merge in this as is, but please add a `// TODO ` about refactoring, and open up a new issue.

Probably call it something like `AnseriniDocument` (disambiguate from Lucene document), and have both `WarcRecord` and `TrecDocument` inherit it...

Peilin-Yang: sounds good.

Issue #78, opened 2016-10-18
Refactoring of the index and document


Peilin-Yang:
For now everything is based on Warc formatted records.
We'd have other types of records too, e.g. Trec text or maybe other types in the future.
It is better to have a base record and everything inherits it.

lintool:
@Peilin-Yang For the TREC documents, you might want to consider what we did for the IR-reproducibility experiments: https://github.com/lintool/IR-Reproducibility/tree/master/systems/lucene/ingester/src/main/java/luceneingester

IIRC, there the code uses an actual XML parser instead of string parsing...

Peilin-Yang:
I think you rely on some Lucene code to do the parsing (https://github.com/lintool/IR-Reproducibility/blob/master/systems/lucene/ingester/src/main/java/luceneingester/IndexThreads.java#L25-L27)?

Yes, Lucene has an arsenal of parsing different old TREC collections, e.g. GOV2, LATimes, FT, FBIS, etc.
https://github.com/apache/lucene-solr/tree/master/lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds

It will be interesting to look into it.

lintool:
Yes, we should copy that code and integrate/refactor in Anserini.

Peilin-Yang:
Merged

PR #84, opened 2016-10-25
refactor of indexing collection

Peilin-Yang:
See #78 

This commit mainly:

1. Splits the IndexWebCollection (renamed to IndexCollection for generality) into pieces: IndexThreads, DiscoverFiles, etc. For now the IndexCollection does everything which is hard to maintain. After the refactor IndexCollection only maintains the main function (the entrance); IndexThreads deals with the threads; DiscoverFiles collects the to-be-indexed files based the collection. 
2. Puts more work to each abcRecord: they go through the documents of each individual file, read documents and index them.

For a new record in the future add an entry to the IndexThreads and put the logic of reading/indexing documents into a new file in document package.


lintool:
Hi @Peilin-Yang thanks for your contribution.

I took a look at your code... here's some discussion:

There's still a lot of logic centralized in `IndexThreads`, where you have "if type foo do this, type bar do that, etc." As we add more collections to index, this would become more complex...

How's this for an alternative proposal:

+ Rename `ISimpleRecord` something like `Indexable` - anything that can be indexed.
+ Create an interface`Collection<D extends Indexable>` - an abstract collection of anything that can be indexed. Maybe it even implements `Iterator`.
+ The indexer takes a `Collection` as input - that's the abstraction.
+ We have concrete implementations - `Gov2Collection`, `ClueWeb09bCollection`, etc. Each of these implementations has the parsing of the records, walking of the files, etc. hidden inside it. But this is hidden from the actual indexer, which basically just needs to know how to call "next document".
+ You can use inheritance to group together "similar collections. E.g., `WarcCollection` implements `Collection`, `ClueWeb09bCollection` extends `WarcCollection`, etc.

Thoughts about this design?

Peilin-Yang:
@lintool 
Yes, this design makes the `Collection` more abstract.
I will try to make the changes soon.

src/main/java/io/anserini/index/collections/ClueWeb12Collection.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import io.anserini.document.ClueWeb12WarcRecord;
+import io.anserini.document.Indexable;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+public class ClueWeb12Collection<D extends ClueWeb12WarcRecord> extends WarcCollection {
+    public ClueWeb12Collection(Path inputDir) throws IOException {
```
lintool: two space indent, and everywhere else.

src/main/java/io/anserini/index/collections/Collection.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Set;
+
+public interface Collection<D extends Indexable> extends Iterator<Indexable> {
+    Set<String> skippedFilePrefix = new HashSet<>();
```
lintool: should this be an abstract class instead because you have instance variables?

src/main/java/io/anserini/index/collections/ClueWeb09Collection.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import io.anserini.document.ClueWeb09WarcRecord;
+import io.anserini.document.Indexable;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+public class ClueWeb09Collection<D extends ClueWeb09WarcRecord> extends WarcCollection {
+    public ClueWeb09Collection(Path inputDir) throws IOException {
```
Peilin-Yang: sure

src/main/java/io/anserini/index/collections/ClueWeb12Collection.java
Peilin-Yang: ok

src/main/java/io/anserini/index/collections/Collection.java
Peilin-Yang: you're right..should be an abstract class.
And `WarcCollection`, `ClueWeb09Collection`, `ClueWeb12Collection` should inherit this instead of `TrecCollection`

src/main/java/io/anserini/index/IndexThreads.java
```diff
@@ -139,111 +138,44 @@ private int indexSimpleRecord(ISimpleRecord ISimpleRecord) throws IOException {
...
-                    if (Collection.CW09.equals(collection)) {
-                        addCount = indexClueWeb09WarcFile();
-                    } else if (Collection.CW12.equals(collection)) {
-                        addCount = indexClueWeb12WarcFile();
-                    } else if (Collection.GOV2.equals(collection)) {
-                        addCount = indexGov2File();
-                    } else if (Collection.TrecText.equals(collection)) {
-                        addCount = indexTrecFile();
+                    Collection curC = null;
+                    if (CollectionClass.TrecText.equals(collectionClass)) {
```
Peilin-Yang: well..I am not aware of this and this is interesting, will try.

Peilin-Yang: It seems that currently there is no support for indexing tweets from files?

lintool:
Hi @Peilin-Yang can you go through all the files and make sure we have two space indentation to be consistent? Also, have you verified that everything still works after refactoring - i.e., effectiveness results still check out?

Peilin-Yang:
@lintool 

1. will do the indent.
2. I've tested the indexes using DumpIndex and the numbers are correct -- disk12345, AQUAINT, Gov2, CW09B. Did not test on CW09A and CW12A but it should work.
3. will verify the effectiveness numbers soon.

lintool:
Yes, we have a separate indexing path for tweets: let's leave that alone for now... we'll circle back and get that indexed in a separate issue.

Are we good to go for merging?

Peilin-Yang:
@lintool let me finish the refactor and the test first?

lintool:
sg

Peilin-Yang:
@lintool Successfully verified the effectiveness numbers on `Gov2`.
I do not discard `CollectionClass` since `SearchWebCollection` uses that and I do not want touch it for now.

lintool:
@Peilin-Yang Are we ready for merging?

Peilin-Yang:
@lintool Yes, if you do not have further questions.
We need a README for `IndexCollection` and another one for `DumpIndex`
I am going to file these issues

Issue #85, opened 2016-10-27
Reorganize docs


lintool:
I don't like GitHub's wiki feature because it's hard to give fine-grained write permission to people.

How about we just move `*.md` into a `docs/` folder and check them into the repo directly? The top-level README can just point to `foo.md` in the `docs/` folder?

That way we keep docs + code in the same repo, preserving history, etc.

Thoughts?

Peilin-Yang:
`docs/*.md` can solve the problem of write permission.
Maybe an up-to-date Wiki is more user friendly?
Is it easy to make linkages between `*.md`?

lintool:
@Peilin-Yang Yes, we can add links using relative paths. So this isn't any different from Wiki, especially since GitHub gives you in-place md updates now.

Check out: https://github.com/lintool/Anserini/tree/reorganize-docs

Peilin-Yang:
@lintool this is much better now!

lintool:
Okay, I will shipit then.

PR #86, opened 2016-10-27
Reorganize docs

lintool:
Breaking long README into smaller files in `docs/` subdirectory. cf. Issue #85


lintool:
@Peilin-Yang Agrees this is the path forward, re Issue #85 
Shipping this branch.

PR #93, opened 2016-11-12
Added example of how to access Lucene index structures

lintool:
I spent some time trying to understand the Lucene API for accessing index structures (e.g., postings lists). What I've learned is encoded in this test case, which should be helpful others that want to manipulate low-level Lucene data structures.


Peilin-Yang:
This is great!
I was thinking the same thing.
I think I can borrow some codes here and add them to the `DumpIndex` somehow.
LGTM, Ship It!

PR #94, opened 2016-11-12
Refactoring the representation of documents and collections

lintool:
+ Created io.anserini.collection to parallel io.anserini.document.
+ Renamed Indexable to SourceDocument to be more descriptive.


Peilin-Yang:
created PR #96 to resolve the issues

PR #98, opened 2016-11-13
Refactoring Twitter tokenizers

lintool:
Twitter tokenizers go into io.anserini.analysis package. Ported test case from https://github.com/lintool/twitter-tools


Peilin-Yang:
not very familiar with this, but LGTM

PR #100, opened 2016-11-14
Index refactoring

lintool:
Main changes:
- Abstracted how we generate a Lucene document for indexing from a source document into the notion of a generator.
- Cleaned up code on multithreaded indexing into a new class called MultithreadedIndexer. The previous implementation used ThreadPoolExecutor in a janky way - cleaned all of that up.
- Now we pass around IndexArgs to the various classes that need configuring - simplifies.


Peilin-Yang:
Tested and it works fine.
Only one concern: do you think it is better to put all the Generators to another folder?

lintool:
Addressed - moved to io.anserini.index.generator package.

PR #103, opened 2016-11-15
Dumpindex improve

Peilin-Yang:
- add more info to `-s` - now we can use `-s` to see the index options of all indexed fields.
- `-t` print posting of the term too.


lintool:
See #104 I believe everything has been fixed now. Closing.

PR #110, opened 2016-11-23
Migration to Lucene 6.3.0

Peilin-Yang:
The main changes are:
- Migration from 5.3.0 to 6.3.0.
- Changed `IntField` to `IntPoint`
- Changed `LongField` to `LongPoint`
- Changed `NumericRangeQuery` to `LongPoint.newRangeQuery`
- Changed `NumericRangeFilter` to a `BooleanQuery` with `BooleanClause.Occur.FILTER` 
e.g. 
```java
Query filter = LongPoint.newRangeQuery(FILTER_FIELD_NAME, 0L, topic.getQueryTweetTime());
Query query = AnalyzerUtils.buildBagOfWordsQuery(SEARCH_FIELD_NAME, TEST_ANALYZER, queryStr);
BooleanQuery.Builder builder = new BooleanQuery.Builder();
builder.add(filter, BooleanClause.Occur.FILTER);
builder.add(query, BooleanClause.Occur.MUST);
Query q = builder.build();
TopDocs hits  = searcher.search(q, 1000);
```
One thing needs to be noticed is `BooleanClause.Occur.MUST`.
The logic of the above code is first filter the documents based on `filter`: only the documents that within the range will be kept. Then we search for the documents with `query` and the documents match the query will be returned.
If we instead use `BooleanClause.Occur.SHOULD` after filtering all the left documents will be returned. I think this is not what we want?
The main changes are related to tweets search so I created `TweetSearch` test case. Please take a look at it.


lintool:
I suppose we need to bite the bullet and do this at some point in time... I'm generally in favor, but after this, no more major changes until after SIGIR deadlines?

Do indexes need to be rebuilt? I suppose we should do it for regression purposes anyway.

Peilin-Yang:
I've test on TREC collections (disk12, disk45, wt, gov2, cw09) and it works fine with the index built with 5.3.0 in terms of `DumpIndex`, `SearchWebCollection`.
For MicroBlog I added the `TweetSearch` test class in order to test the bullets listed above.
But no tests on the real MicroBlog collections.

Based on http://lucene.apache.org/core/6_3_0/MIGRATE.html I do not think the index build part has changed a lot. But yes, regression tests are always needed.

We do want to incorporate the recent changes on Lucene (https://git-wip-us.apache.org/repos/asf?p=lucene-solr.git;a=commitdiff;h=4236da2) as the Axiomatic models are now merged to the main trunk and will be available (maybe) in the next release.

lintool:
Congrats on getting your patch committed, that's awesome!

Let's pull in the axiomatic stuff (and include it in eval) once it's in an official Lucene release? I don't want to start depending on nightly SNAPSHOTs, etc.

PR #112, opened 2016-11-26
Eval update

lintool:
+  Moved topics and qrels document into docs/; other minor tweaks.
+  Added gdeval script from https://github.com/trec-web/trec-web-2013/blob/master/src/eval/gdeval.pl


Peilin-Yang:
ship it!

PR #115, opened 2016-11-28
Updated gov2 experiments

lintool:
+ Updated results from regression of gov2 experiments.
+ Added trec_eval and gdeval examples.


Peilin-Yang:
ship it

PR #118, opened 2016-12-08
Embeddings

rosequ:
- Supporting word2vec for Anserini
- Updated Java version from 1.7 to 1.8 for Lucene compatibility (pom.xml)


src/main/java/io/anserini/embeddings/search/SearchW2V_Args.java
```diff
@@ -0,0 +1,26 @@
+package io.anserini.embeddings.search;
```
lintool: Can you please move this as an inner class, like https://github.com/twitter/GraphJet/blob/master/graphjet-demo/src/main/java/com/twitter/graphjet/demo/PageRankCassovaryDemo.java

src/main/java/io/anserini/embeddings/search/SearchW2V.java
```diff
@@ -0,0 +1,70 @@
...
+import org.kohsuke.args4j.CmdLineParser;
+import org.kohsuke.args4j.OptionHandlerFilter;
+import org.kohsuke.args4j.ParserProperties;
+
+import java.io.*;
+import java.util.Arrays;
+import java.util.Collection;
+
+public class SearchW2V {
+    public static void main(String[] args) {
```
lintool: two space indent, here and everywhere

src/main/java/io/anserini/embeddings/train/TrainW2V.java
```diff
@@ -0,0 +1,63 @@
+package io.anserini.embeddings.train;
```
lintool: same, inner class.

rosequ:
I have made the necessary changes.

PR #120, opened 2016-12-11
add Eval module

Peilin-Yang:
- add the `eval` module which evaluates the ranking list based on judgments.
- for now `map`, `precision`, `ndcg` are available.
- deal with ties the same way as `trec_eval`: first based on score then based on docno, both in descending order.
- got _exactly_ the same results with `trec_eval`, even for the padding format...see below the `diff`
- got the same results with `gdeval.pl` (`gdeval` prefers `.5f` print format while I use `.4f`)
image
image
image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,6 +73,10 @@
               <name>SearchWebCollection</name>
             </program>
             <program>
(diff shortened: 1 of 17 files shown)


Peilin-Yang:
@lintool having time to look at this? any comments?

PR #127, opened 2017-03-01
Indexer refactoring: Merged MultithreadedIndexer into IndexCollection

lintool:
I don't see a compelling reason to have both MultithreadedIndexer and IndexCollection, so I merged into class.

As sanity check, I've re-indexed ClueWeb09b and everything seems to work fine.


Peilin-Yang:
ship it!

Issue #129, opened 2017-03-01
Indexer memory buffer size


lintool:
There's an option to set the size of the in-memory buffer given to the IndexWriter:

```
     final IndexWriterConfig config = new IndexWriterConfig(analyzer);
     config.setSimilarity(new BM25Similarity());
     config.setOpenMode(IndexWriterConfig.OpenMode.CREATE);
     config.setRAMBufferSizeMB(512);
     config.setUseCompoundFile(false);
     config.setMergeScheduler(new ConcurrentMergeScheduler());
```

Results playing with different settings, on ClueWeb09b:

```
 512mb 02:14:06
1024mb 01:55:54
2048mb 01:39:20
4096mb 01:42:27
```

2g seems to be a good setting - I'll set this as the default.

lintool:
Closing, fixed as PR #130

Issue #126, opened 2017-02-28
Play with storing documents directly in the index


lintool:
Experiment with storing documents directly in the index to see how much extra space it takes...

lintool:
I implemented an indexing option to the text directly in the corpus. Here are some experimental results on gov2:

```
 81G original gov2 collection
 11G index, cnt 
 37G index, positions
 82G index, positions + docvectors
150G index, positions + docvectors + doc text
```

Note that this is storing the cleaned doc text (Jsoup removal of HTML pages, etc.). If we wanted to store the raw text, that would be even *more* space.

minafarid:
It looks like the difference between the `82G` and the `150G` is aprpox. the original `81G` documents size, which makes sense since we are storing the original text again in the index.

To me, the size of the index doesn't matter that much. Performance of querying and retrieving documents is more important. If the query performance is hurt, then it is better not to store the document text in the index.

The benefits of storing document text together with the index (as I see it, **please feel free to correct me or add more**) are:

1. Makes the index portable and decoupled from the original data. I just copy the index directory to a new machine, and I have all documents with it.
2. When querying, it saves some time to retrieve the document text in the query results than fetching documents separately

lintool:
I will verify if storing the documents hurts query retrieval performance. I do not think so because IIRC, Lucene fields are stored in "columnar" format.

I will also add an option to store the *raw* document and see how that changes the size. The difference is things like HTML cleaning. Of course, we can always reclean the raw document on demand.

@minafarid Correct, storing the doc makes the index completely portable.

Here's my current thinking: for ClueWeb09b and ClueWeb12-B13, we can probably store the docs. Guesstimate, the entire index will be ~0.5 TB, which is not *that* bad. However, I really can't justify the same for ClueWeb12. That's 5TB of text alone, so estimated index w/ everything will be 10TB. Each.

lintool:
`-storedocs` option merged in as PR #130 
Leaving issue open for further discussion.

lintool:
Okay, here's the update:

```
 81G original gov2 collection
 11G index, cnt 
 37G index, positions
 82G index, positions + docvectors
150G index, positions + docvectors + doc text
219G index. positions + docvectors + raw docs
```

The 81G raw collection becomes 137G once you stuff it into Lucene... I think the size increases because the compression isn't as efficient.

Thoughts?

msci342prof:
One idea: Often significant compression is obtained by how documents are
[...]

PR #130, opened 2017-03-02
Bumped memory buffer, stores docs in index

lintool:
Issue #129 - increase memory buffer size
Issue #126 - option to store document directly in the index


src/main/java/io/anserini/index/IndexCollectionArgs.java
```diff
@@ -51,6 +54,9 @@
   @Option(name = "-docvectors", usage = "boolean switch to store document vectors")
   public boolean docvectors = false;
 
+  @Option(name = "-storedocs", usage = "boolean switch to store raw documen text")
```
minafarid: add to usage message: default is false

lintool: Args4j prints out default values, e.g.,

```
$ nohup sh target/appassembler/bin/IndexCollection
Option "-collection" is required
 -collection VAL   : collection class in io.anserini.collection
 -docvectors       : boolean switch to store document vectors (default: false)
 -generator VAL    : document generator in io.anserini.index.generator
 -index [Path]     : index path
 -input [Path]     : collection path
 -keepstopwords    : boolean switch to keep stopwords (default: false)
 -memorybuffer N   : memory buffer size (default: 2048)
 -optimize         : boolean switch to optimize index (force merge) (default: false)
 -positions        : boolean switch to index positions (default: false)
 -storedocs        : boolean switch to store raw documen text (default: false)
 -threads [Number] : Number of Threads
```

minafarid: okay, I didn't know that.

PR #131, opened 2017-03-03
Separate options to store raw docs and to store transformed docs

lintool:
-storeTransformedDocs: store transformed document text (e.g., HTML cleaning)
-storeRawDocs: store raw document text


HTAustin:
Ship it. Looks fine.

PR #133, opened 2017-03-04
IndexCollectionArgs -> IndexCollection.Args

lintool:
Refactored to make IndexCollectionArgs an inner class of IndexCollection to make package hierarchy cleaner.

Other minor code cleanup.


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -50,6 +47,49 @@
...
+    @Option(name = "-keepstopwords", usage = "boolean switch to keep stopwords")
+    public boolean keepstop = false;
+
+    @Option(name = "-positions", usage = "boolean switch to index positions")
+    public boolean positions = false;
+
+    @Option(name = "-docvectors", usage = "boolean switch to store document vectors")
+    public boolean docvectors = false;
+
+    @Option(name = "-storeTransformedDocs", usage = "boolean switch to store transformed document text")
```
minafarid: Some styling consistency problem, I don't like how some options like `keepstopwords` are all small case while others like `storeTransformedDocs`, `storeRawDocs` are camel case

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -97,23 +140,24 @@ public void run() {
...
     this.args = args;
 
     LOG.info("Index path: " + args.index);
     LOG.info("Threads: " + args.threads);
     LOG.info("Keep stopwords: " + args.keepstop);
     LOG.info("Positions: " + args.positions);
     LOG.info("Store docvectors: " + args.docvectors);
-    LOG.info("Store docs: " + args.storeTransformedDocs);
+    LOG.info("Store docs: " + (args.storeTransformedDocs ? "transformed" :
+        (args.storeRawDocs ? "raw" : "none")));
```
minafarid: if `storeTransformedDocs` and `storeRawDocs` are mutually exclusive, shouldn't this be explained in the Arguments description?

src/main/java/io/anserini/index/IndexCollection.java
lintool: Yea, these args were accumulated over time so they're inconsistent... we should refactor `-keepstopwords` to `-keepStopwords`.

Options like `-docvectors` should really be `-storeDocvector`, and '-stopPositions`?

Makes the options longer, but easier to understand.

Thoughts? Should I make consistent?

minafarid: Yes, I think they should be consistent. Anything we add later on should follow the same style.

src/main/java/io/anserini/index/IndexCollection.java
lintool: Actually, technically, no - so I should actually change this.
You can actually store both if you wish... saving both will save you from needing to do doc cleaning.

PR #135, opened 2017-03-04
Refactoring of index utils

lintool:
Refactoring of @Peilin-Yang index utils that used to be called `DumpIndex`.


Peilin-Yang:
ship it!

Issue #132, opened 2017-03-03
Integrate sentence splitter into Lucene


lintool:
Once we fetch a document, we want to be able to split the text into sentence. We should probably integrate the Stanford tools: http://stanfordnlp.github.io/CoreNLP/ssplit.html

lintool:
IndexUtils now has the sample code of how to get access the transformed document collection, see #135.

lintool:
Merged in with #136 
Closing.

PR #136, opened 2017-03-06
integrating sentence splitter (#132)

rosequ:
Integrated sentence splitter into Lucene  with Stanford CoreNLP.

### DIFF
--- a/src/main/java/io/anserini/index/IndexUtils.java
+++ b/src/main/java/io/anserini/index/IndexUtils.java
@@ -31,7 +31,9 @@ import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
 import org.kohsuke.args4j.ParserProperties;
+import edu.stanford.nlp.simple.*;
 
+import java.util.List;
 import java.io.File;
 import java.io.IOException;
 
@@ -56,6 +58,9 @@ public class IndexUtils {
 
     @Option(name = "-dumpTransformedDoc", metaVar = "docid", usage = "dumps transformed document (if stored in the index)")
     String transformedDoc;
+
+    @Option(name = "-dumpSentences", metaVar = "docid", usage = "splits the fetched document into sentences (if stored in the index)")
+    String sentDoc;
 
     @Option(name = "-convertDocidToLuceneDocid", metaVar = "docid", usage = "converts a collection lookupDocid to a Lucene internal lookupDocid")
     String lookupDocid;
@@ -95,7 +100,7 @@ public class IndexUtils {
     for (String fd : fields) {
       FieldInfo fi = fieldInfos.fieldInfo(fd);
       System.out.println("  " + fd + " (" + "indexOption: " + fi.getIndexOptions() +
-          ", hasVectors: " + fi.hasVectors() + ", hasPayloads: " + fi.hasPayloads() + ")");
+              ", hasVectors: " + fi.hasVectors() + ", hasPayloads: " + fi.hasPayloads() + ")");
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -31,7 +31,9 @@
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
 import org.kohsuke.args4j.ParserProperties;
+import edu.stanford.nlp.simple.*;
```
lintool: Explicit imports of each class.

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -95,7 +100,7 @@ void printIndexStats() throws IOException {
     for (String fd : fields) {
       FieldInfo fi = fieldInfos.fieldInfo(fd);
       System.out.println("  " + fd + " (" + "indexOption: " + fi.getIndexOptions() +
-          ", hasVectors: " + fi.hasVectors() + ", hasPayloads: " + fi.hasPayloads() + ")");
+              ", hasVectors: " + fi.hasVectors() + ", hasPayloads: " + fi.hasPayloads() + ")");
```
lintool: revert needless changes.

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -119,7 +124,7 @@ public void printTermCounts(String termStr) throws IOException, ParseException {
 
   public void printDocumentVector(String docid) throws IOException, NotStoredException {
     Terms terms = reader.getTermVector(convertDocidToLuceneDocid(docid),
-        LuceneDocumentGenerator.FIELD_BODY);
+            LuceneDocumentGenerator.FIELD_BODY);
```
lintool: revert needless changes.

lintool:
This is a bit unrelated, but @rosequ please implement the following as a part of this PR: Since Haotian's index only has raw docs, we have to transform it first. So, the code path should be: if we don't find a transformed doc, we try the raw doc and transform it on the fly. Make the transformer an command-line option also, but make Jsoup the default.

rosequ:
Resolved the issues. Can we ship this now?

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -57,11 +60,18 @@
...
+    @Option(name = "-dumpSentences", metaVar = "docid", usage = "splits the fetched document into sentences (if stored in the index)")
+    String sentDoc;
+
     @Option(name = "-convertDocidToLuceneDocid", metaVar = "docid", usage = "converts a collection lookupDocid to a Lucene internal lookupDocid")
     String lookupDocid;
 
     @Option(name = "-convertLuceneDocidToDocid", metaVar = "docid", usage = "converts to a Lucene internal lookupDocid to a collection lookupDocid ")
     int lookupLuceneDocid;
+
+    //ToDo: Generalize to any transformer
```
lintool: Also, remove this for now... since the generators aren't actually going to work and we'll need to refactor the APIs.

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -57,11 +60,15 @@
...
 
+    @Option(name = "-dumpSentences", metaVar = "docid", usage = "splits the fetched document into sentences (if stored in the index)")
+    String sentDoc;
+
     @Option(name = "-convertDocidToLuceneDocid", metaVar = "docid", usage = "converts a collection lookupDocid to a Lucene internal lookupDocid")
     String lookupDocid;
 
     @Option(name = "-convertLuceneDocidToDocid", metaVar = "docid", usage = "converts to a Lucene internal lookupDocid to a collection lookupDocid ")
     int lookupLuceneDocid;
+
```
lintool: remove

rosequ:
Now?

Issue #121, opened 2017-02-25
Using Anserini from Python?


lintool:
Given the popularity of Python, would make sense to explore how Anserini could be called from Python. One option might be [Py4J](https://www.py4j.org/index.html), which is what Spark uses.

lintool:
We should look at: http://lucene.apache.org/pylucene/index.html

lintool:
Based on meeting today, @rosequ is assigned the lead on this task.

salman1993:
To start, I have two functionalities in mind:

[4:58]  
(1) issue query from Python, get back ranked list.

[4:58]  
(2) fetch a particular document id and get back raw text.

lintool:
Initial try with py4j as part of #139

lintool:
We seem to have a workable solution, as part of #144 
Closing for now. Will reopen if we have issues with py4j.

rosequ:
@lintool please reopen this issue

lintool:
Assigning to @MeowFei

lintool:
If we can't get PyLucene to work, maybe we can try alternatives?

I found [pyjnius](https://github.com/kivy/pyjnius) - maybe give that a try?

Just to explicitly lay this out: the annoying thing about Py4J is this:

>Note that the Java program must be started before executing the Python code above. In other words, the Py4J does not start a JVM.

From [Py4J homepage](https://www.py4j.org/) - this means that we first have to start a Java program, then start Python to connect to it. It would be nice if we could directly embed the JVM inside Python to avoid this step.

MeowFei:
We tried to use Pyjnius to call RetrieveSentences, but we got the following exception:
`  File "src/main/python/RetrieveSentences.py", line 15, in <module>
    test = autoclass("org.apache.lucene.index.DirectoryReaders")
  File "/Users/mengfeiliu/anaconda/lib/python3.6/site-packages/jnius/reflect.py", line 158, in autoclass
    c = find_javaclass(clsname)
  File "jnius/jnius_export_func.pxi", line 25, in jnius.find_javaclass
jnius.JavaException: Class not found b'org/apache/lucene/index/DirectoryReaders'
`
We did some debugging but it didn't seem to help. We are still looking into this issue.

lintool:
Classpath issue? http://pyjnius.readthedocs.io/en/latest/api.html#jvm-options-and-the-class-path

MeowFei:
[...]

PR #139, opened 2017-03-07
 #121 - Using Anserini from Python 

salman1993:
- opened gateway to search with query string and hits to get back a list of document ids
- added docs file to describe the steps


salman1993:
Added functionality to Python program:
- fetch raw document given docid

lintool:
Can you revert `SearchWebCollection` and copy/paste any code into a new class in the `py4j` package?  

In the current implementation, the Python code directly reaches into multiple Anserini objects (`SearchWebCollection` and `IndexUtils`). I think this will make managing dependencies more complex later down the road. As an alternative, we can have a facade object in the `py4j` namespace that serves as the single point of entry for *all* Python code - even if the code just proxies to other Anserini classes.

I think this will make managing the APIs easier later on.

So it'd look something like this:

```
from py4j.java_gateway import JavaGateway

gateway = JavaGateway()
index = gateway.jvm.java.lang.String("path")
searcher = gateway.jvm.io.anserini.py4j.Pyserini(index)
docids = searcher.search(query_string, num_hits)
doc_text = searcher.getRawDocument(docid)
```

In this case, `Pyserini` is the facade object - and that's the only thing we change as we're building new Python-Java bridge methods.

Thoughts?

salman1993:
what about the search(String query, int hits) method in SearchWebCollection - i should keep that there?

lintool:
no... the reason is that `SearchWebCollection` is due for major refactoring, so I don't want to break anything at this point...

salman1993:
I duplicated the required bits from SearchWebCollection for search(...) and I could use IndexUtils for getRawDocument(...)

PR #144, opened 2017-03-14
Flask API, Pyserini additions + cleanup and docs

salman1993:
- Created the Flask API to get a HTTP POST request with question and respond with answer from the PyTorch model. Added documentation.
- Added Pyserini function to get ranked passages given a query.
- Cleaned up the code for Pyserini
- Jaccard similarity model to complete the pipeline for now - will replace this with PyTorch model


src/main/python/api.py
```diff
@@ -0,0 +1,33 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Start the Flask API at the specified host, port')
+    parser.add_argument('--host', help='host address', required=False, type=str, default='0.0.0.0')
+    parser.add_argument('--port', help='port', required=False, type=int, default=5546)
+    parser.add_argument("--debug", help="print debug info", action="store_true")
+    args = parser.parse_args()
+    print("Host: %s" % args.host)
+    print("Port: %s" % args.port)
+    print("Debug info: %r" % args.debug)
```
rosequ: use `logging` instead?

src/main/java/io/anserini/py4j/PyseriniEntryPoint.java
```diff
@@ -121,9 +127,29 @@ public String getRawDocument(String docid) throws Exception {
...
+      for (Sentence sent : sentences) {
+        sentencesList.add(sent.text());
+      }
+    }
+
+    PassageScorer passageScorer = new IdfPassageScorer(indexDir, k);
+    passageScorer.score(sentencesList, "");
+    List<String> topSentences = new ArrayList<>();
+    List<ScoredPassage> topPassages = passageScorer.extractTopPassages();
+    for (ScoredPassage s: topPassages) {
```
tuzhucheng: Space after `s`?

src/main/python/api.py
```diff
@@ -0,0 +1,44 @@
+from flask import Flask, jsonify, request
+import argparse
```
tuzhucheng: For style, standard library imports should go before third party library imports. See https://www.python.org/dev/peps/pep-0008/#imports.

So we should have:

```python
import argparse

from flask import Flask, jsonify, request
...
```

src/main/python/api.py
```diff
@@ -0,0 +1,33 @@
...
+
+app = Flask(__name__)
+
+@app.route("/", methods=['GET'])
+def hello():
+    return "Hello! The server is working properly... :)"
+
```
tuzhucheng: This is the root route. I think it's ok to leave it here to make sure the server works. :)

src/main/python/api.py
```diff
@@ -0,0 +1,44 @@
...
+@app.route('/answer', methods=['POST'])
+def answer():
+    try:
+        req = request.get_json(force=True)
+        question = req["question"]
+        print("Question: %s" % (question))
```
tuzhucheng: Python doesn't recommend using the `%` operator anymore. Should use string.format instead, see https://docs.python.org/3.0/whatsnew/2.6.html#pep-3101-advanced-string-formatting.

```python
print("Question: {}".format(question))
```

src/main/python/api.py
```diff
@@ -0,0 +1,44 @@
...
+        # FIXME: get the answer from the PyTorch model here      
+        answer = get_answer(question)
+        answer_dict = {"answer": answer}
+        return jsonify(answer_dict)
+    except Exception as e:
+        error_dict = {"error": "ERROR - could not parse the question or get answer."}
+        return jsonify(error_dict)
+
+def get_answer(question):
+    index_path = "/home/s43moham/indexes/lucene-index.TrecQA.pos+docvectors+rawdocs"
```
tuzhucheng: Probably shouldn't hardcode user-specific path here? Instead, can read from a config file or argument or environment variable.

src/main/python/api.py
```diff
@@ -0,0 +1,44 @@
...
+    answer = jaccard.most_similar_passage(question, candidate_passages)
+    return answer
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Start the Flask API at the specified host, port')
+    parser.add_argument('--host', help='host address', required=False, type=str, default='0.0.0.0')
+    parser.add_argument('--port', help='port', required=False, type=int, default=5546)
+    parser.add_argument("--debug", help="print debug info", action="store_true")
+    args = parser.parse_args()
+    print("Host: %s" % args.host)
```
tuzhucheng: Ditto about `%` operator.

src/main/python/jaccard.py
```diff
@@ -0,0 +1,46 @@
...
+        for passage in candidate_passages:
+            sim = Jaccard.jaccard_overlap(query, passage)
+            if sim > best_sim:
+                best_sim = sim
+                best_passage = passage
+        return best_passage
+
+
+if __name__ == "__main__":
+    """Test out the Jaccard class."""
```
tuzhucheng: Since this is a test, should we be using unittest? https://docs.python.org/3/library/unittest.html

.gitignore
```diff
@@ -7,4 +7,3 @@ eval/trec_eval.9.0/
 bin/
 *.iml
 .idea
-resources/
```
tuzhucheng: Also ignore *.pyc and remove pyserini.pyc from this PR.

src/main/python/pyserini.py
```diff
@@ -0,0 +1,103 @@
...
+           ----------
+           index_path : str
+               The directory path for the Lucene index.
+        """
+        self.gateway = JavaGateway()
+        self.index = self.gateway.jvm.java.lang.String(index_path)
+        self.pyserini = self.gateway.jvm.io.anserini.py4j.PyseriniEntryPoint()
+        self.pyserini.initializeWithIndex(index_path)
+
+    def search(self, query_string, num_hits=20):
```
tuzhucheng: Since this is a default argument I thought the standard is to have no space around `=`. See http://stackoverflow.com/questions/8853063/pep-8-why-no-spaces-around-in-keyword-argument-or-a-default-parameter-value.

src/main/python/pyserini.py
```diff
@@ -0,0 +1,103 @@
...
+           -------
+            :obj:`list` of :obj:`str`
+               A list of top k passages that matched the query.
+        """
+        passages = self.pyserini.getRankedPassages(query_string, num_hits, k)
+        return passages
+
+if __name__ == "__main__":
+    """Test out the Pyserini class."""
+    index_path = "/home/s43moham/indexes/lucene-index.TrecQA.pos+docvectors+rawdocs"
```
tuzhucheng: Ditto about avoiding hard coding paths in code.

src/main/python/api.py
salman1993: yup, let's leave it there. it simplifies debugging for others using it.

src/main/python/api.py
salman1993: this API is very simple - it's just an endpoint for the UI to pass the question and get the answer back so i wanted to avoid importing the logging module.

src/main/python/pyserini.py
salman1993: good catch. actually, this index path has to be supplied in the arguments. fixing this up.

src/main/java/io/anserini/py4j/PyseriniEntryPoint.java
salman1993: good catch.

src/main/python/api.py
salman1993: oh nice! (had no idea about this one)

src/main/python/jaccard.py
salman1993: we won't be keeping this class so let's not unit test it separately. it'll get replaced by the PyTorch model anyway - just a placeholder for now.

.gitignore
salman1993: oh yeah for sure. didn't see that.

src/main/python/pyserini.py
salman1993: actually, the index_path should be a command line argument - great catch!

src/main/python/api.py
```diff
@@ -0,0 +1,33 @@
...
+        question = req["question"]
+        print("Question: %s" % (question))
+        # get the answer from the PyTorch model here
+        answer = "Albert Einstein (14 March 1879 – 18 April 1955) was a German-born theoretical physicist."
+        answer_dict = {"answer": answer}
+        return jsonify(answer_dict)
+    except Exception as e:
+        error_dict = {"error": "ERROR - could not parse the question or get answer."}
+        return jsonify(error_dict)
+
```
salman1993: The method works but as long as the API is followed. The Exception is there so that the API returns an error msg on some weird failure rather than exiting.

src/main/python/api.py
```diff
@@ -0,0 +1,48 @@
...
+        answer = get_answer(question)
+        answer_dict = {"answer": answer}
+        return jsonify(answer_dict)
+    except Exception as e:
+        error_dict = {"error": "ERROR - could not parse the question or get answer."}
+        return jsonify(error_dict)
+
+# FIXME: separate this out to a classifier class where we can switch out the models
+def get_answer(question):
+    pyserini = Pyserini(args.index)
```
tuzhucheng: `args` is not in scope.

src/main/python/jaccard.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        set1 = set(Jaccard.text2ngrams(text1, n=3))
+        set2 = set(Jaccard.text2ngrams(text2, n=3))
+        intersection = set1.intersection(set2)
+        union = set1.union(set2)
+        return (len(intersection) / float(len(union)))
+
+    @staticmethod
+    def most_similar_passage(query, candidate_passages):
+        """Returns the passage with the highest Jaccard overlap with the query."""
```
tuzhucheng: I think the implementation of this method can be a one-liner.

```python
return max(candidate_passages, key=lambda p: Jaccard.jaccard_overlap(query, p), default="")
```

src/main/python/pyserini.py
tuzhucheng: Should it be changed to a command line argument then?

src/main/python/pyserini.py
```diff
@@ -0,0 +1,103 @@
...
+    pyserini = Pyserini(index_path)
+    # gateway.help(pyserini)
+
+    search_results = pyserini.search(query_string="Airline Subsidies", num_hits=30)
+    print("Search Results:\n%s\n" % search_results)
```
tuzhucheng: Ditto for print and `%`. Same for line 99 and 102.

src/main/python/jaccard.py
salman1993: looks good.

salman1993: i didn't fix this because Python2 doesn't accept default argument plus this is placeholder model for now

tuzhucheng: Ah cool, I thought you are using Python 3 since you are calling print() like a function rather than using it like a statement.

PR #147, opened 2017-03-17
Fixing IdfPassageScorer

rosequ:
Fixed the bugs in `IdfPassageScorer` with the correct implementation.
Removed stopwords from the both the question and the answers.


src/main/java/io/anserini/qa/passage/IdfPassageScorer.java
```diff
@@ -1,60 +1,86 @@
...
     for (String sent: sentences) {
       double idf = 0.0;
-      String[] terms = sent.split(" ");
+      String[] terms = sent.split("\\s+");
       for (String term: terms) {
         try {
           TermQuery q = (TermQuery) qp.parse(term);
           Term t = q.getTerm();
-          idf += similarity.idf(reader.docFreq(t), reader.numDocs());
+          if (questionTerms.contains(t.toString().trim())) {
```
lintool: why don't you stick questionTerms in a hash to make this faster?
Also, pre-trim terms so you're not doing it each time in the loop.

lintool:
Remove `src/main/scala/WmdPassageScorer.scala` from commit?

Issue #128, opened 2017-03-01
More collection refactoring: Collection and CollectionFile


lintool:
@Peilin-Yang I'm looking at the Collection API, and I still don't think we have it quite right. So, we have a `Collection` object representing, say, CW09, right? When we actually perform indexing on a particular WARC file, we do something like:

```
Collection collection = (Collection) collectionClass.newInstance();
collection.prepareInput(inputFile);
... // actual indexing here
collection.finishInput();
```

This means that, for each file, we initialize a separate `Collection` object... but that doesn't make sense, because each file *isn't* a collection.

It seems to be that what we need is a `CollectionFile` interface. A `Collection` gives you a bunch of `CollectionFile` objects, and each one is an iterator over source objects.

Thoughts?

Peilin-Yang:
Yes. I think you're right...
Will refactor once I have time.

lintool:
Thanks. Can you do this CR:
https://github.com/lintool/Anserini/pull/127

I'll merge and you can branch off master, so we don't have merge conflicts...

lintool:
Closing, fixed per #148

PR #148, opened 2017-03-17
Index refactoring: Collection and FileSegments

lintool:
Per #128, refactoring the object modeling of "collections" into Collection and FileSegments. A (static) collection is comprised of multiple FileSegments (i.e., files), each of which contains source documents (to be indexed). Refactored indexer accordingly.

Tested on TREC disks 4-5, gov2, wt10g, ClueWeb09b, and ClueWeb12-B13.


Peilin-Yang:
LGTM

PR #152, opened 2017-03-20
QA System: Show Top k Passages

### LINKED ISSUE #151: QA system: return top k sentences
@salman1993 Can we add a parameter in the Flash server to return the top *k* passages (sentences)

@tuzhucheng Can we display those *k* in the API?

tuzhucheng:
Update Flask API to return top k passages and display them along with the Jaccard similarity between the passage and the query. Resolves #151 

image

@salman1993


salman1993:
Looks good!

PR #155, opened 2017-03-20
Updated docs for Flask service

lintool:
Added documentation on how to use curl to access the Flask service.


salman1993:
Looks good but we also need to changed the example Request and Response since we now pass in num_hits and k as parameters as well.

lintool:
Branches have diverged. @salman1993 or @tuzhucheng can you just roll in a sample `curl` command in your new patch?

Issue #154, opened 2017-03-20
QA System: voice frontend


lintool:
@tuzhucheng Can you make `Authorization: window.WITAI_API_SECRET` a command-line arg  or something like that in startup up I don't have to either open up the console or modify code directly?

tuzhucheng:
Yes, just thought of a good way. Will do today.

________________________________

lintool:
Closing.

PR #156, opened 2017-03-20
Read secrets from config file

tuzhucheng:
Addresses #154. Can place the key in `config.cfg` as a one-time change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ bin/
 *.pyc
 *node_modules/
 src/main/js/SpeechDemo-darwin-x64/
+src/main/python/config.cfg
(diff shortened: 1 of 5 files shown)


salman1993:
That makes sense. Let's show the API key then. But in that case, why not just hard code the Wit API key in the configuration file?

tuzhucheng:
Ah so when one uses the config he/she copies `config.cfg.example` to `config.cfg` (I've documented this) and changes the key from a placeholder to a real one. The `config.cfg` is gitignored so everyone can use different configs for their own index path, keys, etc.. Since this is a public repo it is probably not a good idea to make an API key publicly available.

salman1993:
I see. All of this looks good! :)  Just move the index, host port to a configuration file and make a PR.

PR #157, opened 2017-03-20
fixed #150: QA system bug: what is the capital of canada

### LINKED ISSUE #150: QA system bug: what is the capital of canada
Here's what I'm getting:

```
vpn-019:js jimmylin$ curl -H "Content-Type: application/json" -X POST -d '{"question": "what is the capital of canada"}' http://localhost:5546/answer
{
  "error": "ERROR - could not parse the question or get answer."
}
```

@salman1993 @rosequ any ideas?

salman1993:
- fixed the QA bug by adding query to sentences list 
- fixed another API bug while reading from config file - port must be int, not str

### DIFF
--- a/src/main/python/api.py
+++ b/src/main/python/api.py
@@ -57,4 +57,4 @@ if __name__ == "__main__":
     print("Port: {}".format(app.config['port']))
     print("Debug info: {}".format(args.debug))
 
-    app.run(debug=args.debug, host=app.config['host'], port=app.config['port'])
+    app.run(debug=args.debug, host=app.config['host'], port=int(app.config['port']))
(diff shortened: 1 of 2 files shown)


src/main/python/api.py
```diff
@@ -57,4 +57,4 @@ def get_answers(question, num_hits, k):
     print("Port: {}".format(app.config['port']))
     print("Debug info: {}".format(args.debug))
 
-    app.run(debug=args.debug, host=app.config['host'], port=app.config['port'])
+    app.run(debug=args.debug, host=app.config['host'], port=int(app.config['port']))
```
tuzhucheng: Interesting, I was aware of this, didn't do this conversion, and the server started fine.

Issue #145, opened 2017-03-15
Word Embeddings with Lucene


rosequ:
Loading pre-trained word2vec models is extremely slow with Deeplearning4j. We'll be using a Lucene index instead, with a word as the index and the embedding as the corresponding document.

lintool:
Initial implementation completed.

PR #159, opened 2017-03-22
Fixing #145

rosequ:
Used Lucene to access word embeddings faster.
- Indexes in 434.308s
- Final index size: 3G


pom.xml
```diff
@@ -409,5 +414,16 @@
       <artifactId>py4j</artifactId>
       <version>0.10.4</version>
     </dependency>
+
```
lintool: Why do we need Spark?

src/main/java/io/anserini/embeddings/search/IndexW2V.java
```diff
@@ -0,0 +1,177 @@
...
+    public String input;
+
+    @Option(name = "-index", metaVar = "[Path]", required = true, usage = "index path")
+    public String index;
+
+    @Option(name = "-term", metaVar = "String", usage = "get the embeddings corresponding to the term")
+    public String term = "";
+
+    //optional arguments
+    @Option(name = "--build", usage = "builds index if specified")
```
lintool: What does this option mean? Why wouldn't you build the index? And why double-dash?

pom.xml
rosequ: I think I copied the old pom.xml. Let me change it.

src/main/java/io/anserini/embeddings/search/IndexW2V.java
```diff
@@ -0,0 +1,177 @@
...
+
+    if (!Files.exists(this.indexPath)) {
+      Files.createDirectories(this.indexPath);
+    }
+
+    this.directory = FSDirectory.open(indexPath);
+  }
+
+  public float[] getEmbeddings(String term) throws IOException {
+    DirectoryReader reader = DirectoryReader.open(directory);
```
lintool: Hrm... you don't want to do initialize a directory reader each time...
Also, conceptually, the lookup functionality doesn't belong in the indexer.
Also, rename to getEmbeddingVector to be clear.

src/main/java/io/anserini/embeddings/search/IndexW2V.java
rosequ: This code is both builds and searches the index. An index is built if `--build` is specified. Double dash because the option doesn't take arguments.

lintool: Oh, I get it - you can either index or lookup.

lintool: I see. Lookup should be another class. Don't mix with indexer.

src/main/java/io/anserini/embeddings/search/IndexW2V.java
rosequ: Should I create yet another file, maybe? Just for obtaining the embeddings? 
Also, I could initialize the reader in a constructor, but the index may not exist... Should separate the logic by creating two different files?

lintool: Yes, a separate class. Call it WordEmbeddingDictionary or something like that.
The constructor takes an index path.

pom.xml
```diff
@@ -13,6 +13,7 @@
     <JETTY_VERSION>8.1.17.v20150415</JETTY_VERSION>
     <MUSTACHE_VERSION>0.8.18</MUSTACHE_VERSION>
     <JERSEY_VERSION>2.7</JERSEY_VERSION>
+    <spark.version>1.6.0</spark.version>
```
lintool: I think you want to remove this also?

src/main/java/io/anserini/embeddings/WordEmbeddingDictionary.java
```diff
@@ -0,0 +1,110 @@
...
+    @Option(name = "-term", metaVar = "String", usage = "get the embeddings corresponding to the term")
+    public String term = "";
+  }
+
+  private final Path indexPath;
+  private final FSDirectory directory;
+  private final DirectoryReader reader;
+
+  public WordEmbeddingDictionary(String indexPath) throws Exception {
+    this.indexPath = Paths.get(indexPath);
```
lintool: Be more specific about the Exception - `IOException`?

src/main/java/io/anserini/embeddings/WordEmbeddingDictionary.java
```diff
@@ -0,0 +1,110 @@
...
+
+    if (!Files.exists(this.indexPath)) {
+      Files.createDirectories(this.indexPath);
+    }
```
lintool: We should probably throw exception if the index doesn't exist? Doesn't make sense to create empty?

src/main/java/io/anserini/embeddings/WordEmbeddingDictionary.java
```diff
@@ -0,0 +1,110 @@
...
+
+    this.directory = FSDirectory.open(this.indexPath);
+    this.reader = DirectoryReader.open(directory);
+  }
+
+  public float[] getEmbeddingVector(String term) throws IOException {
+    IndexSearcher searcher = new IndexSearcher(reader);
+    Query query = AnalyzerUtils.buildBagOfWordsQuery(FIELD_ID, new WhitespaceAnalyzer(), term);
```
lintool: create a singleton - static final instance so we're not creating a new object every time.

src/main/java/io/anserini/embeddings/WordEmbeddingDictionary.java
```diff
@@ -0,0 +1,110 @@
...
+    TopDocs rs = searcher.search(query, 1);
+    ScoredDocuments docs = ScoredDocuments.fromTopDocs(rs, searcher);
+
+    byte[] val = docs.documents[0].getField(FIELD_BODY).binaryValue().bytes;
+    FloatBuffer floatBuffer = ByteBuffer.wrap(val).asFloatBuffer();
+    float[] floatArray = new float[floatBuffer.limit()];
+    floatBuffer.get(floatArray);
```
lintool: This seems dangerous somewhere here... is there potential for NPE that we need to catch?

src/main/java/io/anserini/embeddings/search/IndexW2V.java
```diff
@@ -0,0 +1,177 @@
+/**
```
lintool: Why are there two different IndexW2V files?

Issue #149, opened 2017-03-17
Indexer for Wikipedia articles


lintool:
Add Anserini indexing support for Wikipedia articles.

lintool:
This is a duplicate of #10.

lintool:
Fixed.

PR #160, opened 2017-03-25
Collection API for Wikipedia

lintool:
So that we can index Wikipedia, see #149


src/main/java/io/anserini/document/WikipediaArticle.java
```diff
@@ -0,0 +1,43 @@
...
+    return title;
+  }
+
+  @Override
+  public String content() {
+    return contents;
+  }
+
+  @Override
+  public boolean indexable() { return true; }
```
rosequ: should this be on a single line?

lintool: Oh yea, I'll fix.

PR #161, opened 2017-03-26
Indexer refactoring: creation of StringTransform

lintool:
Refinement of the object model for the indexing pipeline. A StringTransform is a string-to-string transformation. In the context of the indexing pipeline, it can be used to, for example, clean HTML documents into plain text documents.

Previously, this functionality was tucked away inside the LuceneDocumentGenerator - so that if we wanted it explore different HTML parsers, for example, we would have needed to duplicate a lot of code. Now, LuceneDocumentGenerator has pluggable StringTransforms.

This allowed us to easily run a long-standing experiment I've always wondered about: Jsoup vs. Lucene's use of the NekoHTML parser.


lintool:
Transformer, as in?
Transformer

Transformer sounds too odd to me... that's why I went with transform. I mean it as a noun, not a verb, if it helps..

Seeing no further comments, I'll merge and close.

Issue #158, opened 2017-03-21
QA system: display and config bugs


lintool:
I appear to be getting some display issues:

<img width="472" alt="screen shot 2017-03-21 at 8 11 31 am" src="https://cloud.githubusercontent.com/assets/313837/24146930/5b6f4e14-0e0e-11e7-86f0-eb57fcebcb80.png">

Can't scroll back up to the waveform visualization...

Also, `config.cfg` should be in base directory, not as the doc currently suggests:

> Copy src/main/python/config.cfg.example to src/main/python/config.cfg and make necessary changes, such as setting the index path and API keys.

Finally, the config content is something like:

> [Flask]
> host=0.0.0.0
> port=5546
> index=lucene-index.TrecQA.pos+docvectors+rawdocs
> witai_api_secret=XXXX

But this is a bit confusing since the API key is part of the frontend, not Flask, correct? IIRC, the js frontend only sends the QA string to the Flask server?

lintool:
While we're at it, can we just roll in a sample `curl` command with this issue?

tuzhucheng:
Will take a look at the scrolling issue.

The Flask app reads the config from the root of the Python code in `src/main/python/`. Can change it to read the config from the base directory.

The Flask app doesn't need the witai_api_secret. It can instead be supplied when the voice frontend is run. That will require a separate config file for the voice frontend or command line argument when the voice frontend is loaded. I wasn't aware of the option to pass custom command line options to `electron` which is used to run the voice frontend, but it does have that option.

lintool:
According to https://github.com/lintool/Anserini/blob/master/docs/flask-api.md

The instructions are:

```
sh target/appassembler/bin/PyseriniEntryPoint
python src/main/python/api.py [--debug]
```

As a result, the Python code reads from the base directory.

Concrete proposal:

+ Change docs to copy config to base directory.
+ Have one config, but separate sections for both Flask and the frontend.

Thoughts?

tuzhucheng:
I can move the example config to the base directory as well, so it will be a copy in the same directory.

I like the concrete proposal.

lintool:
One more request - can we collapse all docs into a single file? It's allowing having to get startup instructions from three different files...

tuzhucheng:
Just collapse these files right (all docs for QA system)?

* py4j-gateway.md
* flask-api.md
* speech-to-text.md

salman1993:
i agree. the docs should be collapsed.

PR #166, opened 2017-03-30
ported API code to Python3, collapsed UI, API, PythonGateway docs

salman1993:
- changed API code to run on Python3
- collapsed UI, API, PythonGateway docs to one docs file
- updated API docs and added steps to start the system


docs/speech-ui-api-docs.md
```diff
@@ -0,0 +1,83 @@
+# Commands to get the UI and server running once configured
+```
+sh target/appassembler/bin/PyseriniEntryPoint
+python3 src/main/python/api.py
```
tuzhucheng: Starting it with the root directory as the working directory doesn't actually work with the config - I'll fix it as a part of #158.

docs/speech-ui-api-docs.md
```diff
@@ -0,0 +1,83 @@
...
+# Setting up Speech to Text Demo
+
+`cd` into `src/main/js`.
+
+```sh
+npm install
+npm start
+```
+
+You should see a new icon in the toolbar. Click it and a small window will open. Use the keyboard shortcut for Chrome developer tools to open up the console. You'll need to type in the following command to enter the API key before using it:
```
tuzhucheng: I'll update this later.

src/main/python/api.py
```diff
@@ -52,7 +53,11 @@ def get_answers(question, num_hits, k):
     parser.add_argument("--debug", help="print debug info", action="store_true")
     args = parser.parse_args()
 
-    config = ConfigParser.ConfigParser()
+    if not os.path.isfile(args.config):
+        print("The configuration file ({}) does not exist!".format(args.config))
+        exit(1)
```
tuzhucheng: Should use `sys.exit` instead I think. See http://stackoverflow.com/a/6501134/2514889 for more details.

PR #167, opened 2017-04-02
added docs and Bash script

salman1993:
- added docs for setting up e2e qa system
-  added a Bash script to start the Gateway server, Flask API and the JavaScript app

### DIFF
--- a/docs/e2e-qa-system.md
+++ b/docs/e2e-qa-system.md
@@ -1,0 +1,52 @@
+# Get all the repos in castorini and arrange them in this format
+- castorini
+    -- Anserini
+    -- Castor
+    -- data
+    -- models
+    
+# Configure the speech-UI in Anserini
+- make the config file and add in the secret key for Wit API
+​
+# Getting the S&M model to src/main/python
+- Copy the directory Castor/sm_model to Anserini/src/main/python/sm_model
+- Copy the file Castor/castorini_smmodel_bridge.py to Anserini/src/main/python/
+​
+# Files for the model, w2v, word2dfs, stopwords
+- Train the model or use the pre-trained one in models/sm_model/sm_model.TrecQA.TRAIN-ALL.2017-04-02.castor
+- Training the model also produces word2vec file that is also needed later
+- Build the TrecQA Dataset - https://github.com/castorini/data/tree/master/TrecQA
+- This produces the word2dfs.p file that is also needed later
+- The stopwords.txt file is located under data/TrecQA/
+​
+# Making changes in Flask api
+- Open up the file Anserini/src/main/python/api.py and make these changes:
+```
+from castorini_smmodel_bridge import SMModelBridge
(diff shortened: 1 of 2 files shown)


docs/e2e-qa-system.md
```diff
@@ -0,0 +1,52 @@
...
+        ps_split = ps.split('\t')
+        candidate_passages.append(ps_split[0])
+​
+    answers_list = model.rerank_candidate_answers(question, candidate_passages)
+    answers = []
+    for score, sent in answers_list:
+        answers.append({'passage': sent, 'score': score})
+​
+    return answers
+```
```
rosequ: @salman1993 @lintool shouldn't we make this a part of `api.py` code instead of asking the user to change it?

salman1993: I agree but we need the model which in turn needs copying the sm_model directory which exists in a different repo and has to be copied over.
So... if we make this a part of the api.py code at this moment, the whole thing will break so I documented it. The ideal solution (that I propose) will be to have an interface for the Model or Bridge or whatever and all future models will obey that contract. So all we will need to do is import and switch out the model and we can have a pre-built IdentityModel that does nothing but returns the candidate passages.

rosequ: Or, we can just import across directories. Then the user wouldn't have to copy anything...

salman1993: So sth like 'from Castor.sm_model.sm_model_bridge import SMModel'? That will also include exporting Castor as a package (pretty good solution) or putting __init__.py in castorini (ugly!)

rosequ: Why is `__init__.py` ugly? I have seen standard packages using it... Also, couldn't find anything in PEP8 about it... @tuzhucheng any suggestions?

rosequ: Why is __init__.py ugly? I have seen standard packages using it... Also, couldn't find anything in PEP8 about it... @tuzhucheng any suggestions?

salman1993: The current directory format is this:
```
- castorini
  ---> Anserini/src/main/python/api.py
  ---> Castor/sm_model/*  -> contains the model
  ---> Castor/sm_model_bridge.py -> contains the bridge file
```
So if we want to use __init__.py here, imagine how many places we have to use it to directly import that model in Anserini/src/main/python from Castor. Wouldn't it be better to just export it as a package and import that?

tuzhucheng: Following the suggestion of http://stackoverflow.com/a/14509415/2514889, maybe we should create each project as a discrete installable package? This way we can import another project as a module.

As an example, I made some modifications to an open-source python package called uap-python and hosted it on GitHub: https://github.com/tuzhucheng/uap-python. I can just install it from pip using `pip install -e git+https://github.com/tuzhucheng/uap-python.git#egg=ua-parser`. I can then use this as a module from another Python project.

salman1993:
closing this PR - covered by commit #170

PR #168, opened 2017-04-07
Evaluating TrecQA systems

rosequ:
- made the QueryParser more forgiving by ignoring the Wildcard characters in the query
- switched to `compareTo()` to handle ordering of -0.0 and Nan
- an evaluation script for evaluating TrecQA systems


src/main/python/evaluate.py
```diff
@@ -0,0 +1,195 @@
...
+import os
+import re
+
+from nltk.tokenize import TreebankWordTokenizer
+from pyserini import Pyserini
+from castorini_smmodel_bridge import SMModelBridge
+
+
+path_to_castorini = "/Users/royalsequeira/Castorini"  # change this path
+model = SMModelBridge(path_to_castorini + '/models/sm_model/sm_model.no_ext_feats',
```
lintool: don't hard code this.

lintool:
Someone else please look at Python code.
Please write a README on how to user this - we need to start up the bridge, for example?

src/main/python/evaluate.py
rosequ: oh sorry, forgot to change it

rosequ:
Can someone do a CR for this one, please...

rosequ:
@gauravbaruah @salman1993 @HTAustin please do a CR on this and let's have it shipped.

PR #170, opened 2017-04-16
end2end qa pipeline (#167)

rosequ:
- Pipeline for running end2end QA system
- Managed model dependencies through python packaging
- Added documentation and shared the additional required files.

### DIFF
--- a/docs/setup_qa_pipeline.md
+++ b/docs/setup_qa_pipeline.md
@@ -1,14 +1,24 @@
 # Steps to setup an end2end QA pipeline 
-Clone the models and data repo:
+Clone the Anserini, models, and data repo:
 ```
+git clone https://github.com/rosequ/Anserini.git
 git clone https://github.com/castorini/data.git
 git clone https://github.com/castorini/models.git
 ```
 
-Clone and build Anserini:
+
+You should have the following directory structure after cloning:
+```
+.
+├── Anserini
+├── data
+└── models
 
 ```
-git clone https://github.com/rosequ/Anserini.git && cd Anserini
+
+Build Anserini:
+```
+cd Anserini
 mvn clean package appassembler:assemble
 ```
 
@@ -40,3 +50,8 @@ Start the Gateway server, Flask API, and the javascript app:
 chmod +x run_ui.sh
 ./run_ui.sh
 ```
+
+
+__NB:__  The speech UI cannot be run in Ubuntu. To test the pipeline in Ubuntu, make the following changes: 
+- Comment out the JavaScript part and run the Bash script
+- Make a REST API query to the endpoint using Postman, Curl etc.
--- a/docs/setup_qa_pipeline.md
+++ b/docs/setup_qa_pipeline.md
@@ -1,0 +1,42 @@
+# Steps to setup an end2end QA pipeline 
+Clone the models and data repo:
+```
+git clone https://github.com/castorini/data.git
+git clone https://github.com/castorini/models.git
+```
+
+Clone and build Anserini:
+
+```
+git clone https://github.com/rosequ/Anserini.git && cd Anserini
+mvn clean package appassembler:assemble
+```
+
+
+Install the dependency packages:
+
+```
+pip3 install -r requirements.txt 
+```
+
+Make sure that you have PyTorch installed. For more help, follow [these](https://github.com/castorini/Castor) steps.
+
+### Additional files for pipeline:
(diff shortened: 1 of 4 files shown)


rosequ:
@nims11, @gauravbaruah please do a CR

docs/setup_qa_pipeline.md
```diff
@@ -0,0 +1,42 @@
+# Steps to setup an end2end QA pipeline
```
salman1993: Can you please add this directory structure to the docs as well? Some may clone the directories in different places.

Get all the repos in castorini and arrange them in this format
- castorini
    -- Anserini
    -- Castor
    -- data
    -- models

docs/setup_qa_pipeline.md
```diff
@@ -0,0 +1,42 @@
...
+
+Make sure that you have correctly set your `config` file. For help, refer 
+[this](https://github.com/castorini/Anserini/blob/master/docs/speech-ui-api-docs.md).
+
+Start the Gateway server, Flask API, and the javascript app:
+
+```
+chmod +x run_ui.sh
+./run_ui.sh
+```
```
salman1993: Mention that this won't work in Ubuntu:
- comment out the npm from the script 
- use curl or Postman to test the endpoint

src/main/python/api.py
```diff
@@ -6,7 +6,13 @@
 from flask import Flask, jsonify, request
 # FIXME: separate this out to a classifier class where we can switch out the models
 from pyserini import Pyserini
-from jaccard import Jaccard
+from sm_model.bridge import SMModelBridge
+
+path_to_castorini = os.getcwd() + "/../../../.."
```
salman1993: This requires the directory structure to be maintained so it's important that they clone the repos in the same place.

rosequ: Yes, adding a typical structure in the doc should solve this problem...

rosequ:
@salman1993 have another look and see if we can ship this...?

salman1993:
ship it!

src/main/python/api.py
tuzhucheng: This is highly fragile. It's ok for a temporary solution but I have an alternative proposal: #172

src/main/python/api.py
```diff
@@ -6,7 +6,13 @@
 from flask import Flask, jsonify, request
 # FIXME: separate this out to a classifier class where we can switch out the models
 from pyserini import Pyserini
-from jaccard import Jaccard
+from sm_model.bridge import SMModelBridge
```
tuzhucheng: How does this work? I get ModuleNotFoundError: No module named 'sm_model'? sm_model is a part of Castor.

rosequ: @tuzhucheng did you follow these steps: https://github.com/castorini/Anserini/blob/master/docs/setup_qa_pipeline.md ?

docs/setup_qa_pipeline.md
```diff
@@ -0,0 +1,57 @@
+# Steps to setup an end2end QA pipeline 
+Clone the Anserini, models, and data repo:
+```
+git clone https://github.com/rosequ/Anserini.git
```
nims11: Fix the url?

docs/setup_qa_pipeline.md
```diff
@@ -0,0 +1,57 @@
...
+git clone https://github.com/castorini/data.git
+git clone https://github.com/castorini/models.git
```
nims11: Make these two installable by pip and add them to requirements.txt

rosequ: @nims11 only Anserini has python code and needs to be pip installed. @tuzhucheng clone the original Anserini repo and check if the issue repeats. I'll have the docs updated.

nims11: You don't need to have python code to make it a python package. We are just using pip and requirements.txt here as a module management tool.

nims11:
I also suggest to enforce `virtualenv` usage through the docs. `requirements.txt` has static version number (which is how it should be) defined and many people like to keep their global python environment updated. `virtualenv` is highly recommended specially when multiple people are working on a python project.

tuzhucheng:
I'm using `virtualenv` myself, need to document it sometime soon :)

rosequ:
I use it too but left the details the users. We, should probably mention it in docs. @tuzhucheng is your issue resolved?

tuzhucheng:
Yes @rosequ, I missed the install from requirements.txt - can get it working now!

PR #174, opened 2017-04-20
Fix Speech Demo UI/Config Issues

tuzhucheng:
Addresses issues in #158 

* Move config to root directory with section headers. **Please change the old config format to follow the new config example**
* Answer box no longer overflows to hide the speech button
* Minor doc updates

@rosequ @salman1993


docs/speech-ui-api-docs.md
```diff
@@ -16,19 +16,10 @@ npm install
...
-```javascript
-window.WITAI_API_SECRET = "Bearer XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX";
-```
-
-Ask for the API key or get a free one from wit.ai.
-
-
 # Flask API
 
 - Install Flask with the command: pip install flask
```
rosequ: Should we just use `pip3 install -r requirements.txt` instead? The new requirements file has Flask already...https://github.com/castorini/Anserini/blob/master/requirements.txt

docs/speech-ui-api-docs.md
```diff
@@ -16,19 +16,10 @@ npm install
...
-- Copy `src/main/python/config.cfg.example` to `src/main/python/config.cfg` and make necessary changes, such as setting the index path and API keys.
+- Copy `config.cfg.example` to `config.cfg` and make necessary changes, such as setting the index path and API keys.
 - The Flask API can be started with the following command. Default host is 0.0.0.0 and port is 5546 without debugging info.
 - Make sure to start the PyseriniEntryPoint (gateway) from Java to open up a socket for communication for Pyserini.
```
rosequ: This line can be removed. `run_ui.sh` looks after this case.

rosequ: `run_ui.sh` looks after this case. We may have to update the rest of the docs to avoid confusion....

tuzhucheng:
@rosequ I updated the docs. How does it look now?

rosequ:
@lintool let's merge this...

Issue #134, opened 2017-03-04
Integrate APIs for Freebase and other knowledge graphs


lintool:
Starting this issue as a placeholder for discussion with @minafarid et al.

We need to be able to access Freebase and knowledge graphs from Anserini - what form that takes is up for discussion, but this is needed for entity linking, relational extraction, etc.

minafarid:
I already have Freebase  RDF data uploaded to one of our servers (latest version, before Freebase retired). The data is imported into Virtuoso (RDF store) and exposed through a RESTful API. Querying Virtuoso is done using SPARQL. The same endpoint also contains DBPedia RDF (from 2014 or 2015 I think).

lintool:
How big is the Freebase RDF data? Is there an embedded option, e.g., using Apache Jena?
Or could we even throw the triples into Lucene?
I think in this context we really just want point lookups of triples, right?
This could reduce external dependencies.

lintool:
@minafarid Additional thought on this - at some point in time we need to index the entities in the KB itself so we can do entity linking (e.g., partial matches on names)... which we'll need Lucene for anyways. So if we can squeeze everything into Lucene, that will simplify our dependencies.

lintool:
We should take a look at this:
http://www.isi.edu/isd/LOOM/kres/freebase-tools/

lintool:
Assigning this task to @salman1993

minafarid:
I indexed freebase using http://www.isi.edu/isd/LOOM/kres/freebase-tools/

**Summary**
The tool performs indexing on two steps: 
1. Compress the URIs and Properties in the RDF triples (e.g., compress `http://rdf.freebase.com/` to `f_`). Kind of prefix compression. This is done using a `bash` script
2. Index the compressed triples using Lucene. This is done using `java`.

**Results**
* The compression script is very buggy and doesn't work on ubuntu.
* I ignored the compression script and ran the indexing directly on the full data.
* Indexing took ~ 2.5 hours on a 64-core machine with 256 GB ram, but not all resources were utilized during the indexing process. The documentation says that the compressed version would take 70 minutes to compress.
* Index size ~ 37 GB. Compressed version should be 13 GB

minafarid:
**Query expressiveness**

- It is easy to retrieve all or a subset of properties for a particular resource/entity.
- More complex queries is claimed to be available through coding
- Freebase-tools is not a replacement for sparql interfaces

**Placeholder for performance comparison against virtuoso**

PR #177, opened 2017-04-28
Indexing Freebase and Generating training data

minafarid:
Issue #134 - Implemented a class to facilitate generating training data.

**Implemented functionalities:**
* Index Freebase RDF dump
* Index entity mentions in ClueWeb09 using ClueWeb09FACC (http://lemurproject.org/clueweb09/FACC1/)
* Generate training data for properties. Currently implemented properties:
  - Birthdate
* Training data generation produces two files:
  - `[property_name].tsv` contains Freebase entity ids, their labels, and their property values.
  - `entity_mentions.tsv` contains freebase entity ids, labels as they appear in the documents, and ClueWeb09 documents in which they are mentioned.
* Also supports querying the indexes (`freebase` and `entity-mentions`) to find triples of a particular subject or mentions of a particular freebase entity in the corpus.

**Notes:**
I) In order to retrieve training data for other properties, the following needs to be done:

1. Re-indexing Freebase with the target property as indexable field. This takes around 2.5 hours.
2. Modifying the `TrainingDataGenerator` class to add another function to generate training data files, similar to the `birthdate()` function. The function should retrieve triples appropriately from the new freebase index, performing all the required joins that might be necessary (e.g., in the case of retrieving relationship examples like spouses)

II) Extending to other datasets might not be straightforward due to the following assumptions:

1. The KB indexing assumes that the triples are ordered by the subject (similar to the Freebase dump). This allows efficient indexing as we index all triples with the same subject as one document. Otherwise, we would need another representation of the indexed data/
2. We may need to re-write the training data generation methods because querying the new indexes might not be compatible with how freebase represents information. For example, Wikidata represents all data in RDF statements. Therefore, a different retrieval strategies might be needed to obtain the same kind of training data.


lintool:
Yes, we've adopted two-space indentation instead of four space.

I'm wondering why you've implemented everything in a single class. Why don't we implement a standalone indexer as one class, and then another class to generate training examples?

minafarid:
* OK will change to 2 spaces.

Originally, I wanted to extend `Collection` and let `IndexCollection` take care of indexing... However, Freebase dump has triples ordered by the subject, which means that we can sequentially scan the dump to create a single lucene document for every subject (and its predicates and object values).  `IndexCollection` was running multiple threads to index multiple `FileSegment`  (which I didn't fully understand at first), so I thought freebase won't fit this model.

I have two possible solutions in mind to really integrate what I wrote into Anserini:

Option 1: 
a. Initialize a single `FileSegment` for the whole rdf dump. 
b. linearly scan triples in that file segment, and add each subject and its predicates and object as a `SourceDocument` object and send this object for indexing.

Option 2:
a. split the `.gz` rdf dump into multiple `FileSegment` objects (if possible?) 
b. construct a `SourceDocument` from every triples. This will change the way data is represented in the index, which may require some minor modifications to the retrieval code.

Which (if any) sounds more interesting?

lintool:
I think you're waaay over thinking this. How about a class called `IndexFreebase`? How about another class called `SearchFreebase`. How about a third class for generating the training data?

minafarid:
I did major refactoring, tried to make the indexing and search process similar to other collections in Anserini by creating:
* `RDFDocument` extends `SourceDocument` to hold predicates and objects of a single entity
* `RDFCollection` extends `Collection` to hold RDF triple datasets and read `.gz` files to extract multiple triples of the same subject and builds an `RDFDocument`
* `LuceneRDFDocumentGenerator` to convert `RDFDocument` to lucene `Document`
* `IndexRDFCollection` similar to `IndexCollection` with main function to index RDF collections
* `SearchRDF` with a main function to search in the build index on RDF collection
* Some unit tests in `LuceneRDFDocumentGeneratorTest.java`

src/main/java/io/anserini/document/RDFDocument.java
```diff
@@ -0,0 +1,137 @@
...
+
+  /**
+   * Subject of the triples doc, also the RDFDocument id
+   */
+  private String subject;
+
+  /**
+   * The predicates and values of the subject entity
+   */
+  private Map<String, List<String>> predicateValues = new TreeMap<>();;
```
rosequ: remove the extra ;

src/main/java/io/anserini/document/RDFDocument.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    predicateValues.forEach((predicate, values) -> {
+      for (String value : values) {
+        sb.append(subject).append(TRIPLE_SPLITTER)
+                .append(predicate).append(TRIPLE_SPLITTER)
+                .append(value).append(TRIPLE_SPLITTER).append(".\n");
```
rosequ: why is the period necessary here?

src/main/java/io/anserini/index/IndexRDFCollection.java
```diff
@@ -0,0 +1,224 @@
...
+
+  /**
+   * Default index analyzer (uses StandardAnalyzer)
+   */
+  public final Analyzer ANALYZER;
+
+  /**
+   * Program arguments to hold parameters and properties.
+   */
+  Args args;
```
rosequ: Make the ANALYZER and args local?

src/main/java/io/anserini/index/IndexRDFCollection.java
```diff
@@ -0,0 +1,224 @@
...
+    } catch (CmdLineException e) {
+      System.err.println(e.getMessage());
+      parser.printUsage(System.err);
+      System.err.println("Example command: "+ IndexRDFCollection.class.getSimpleName() +
+              parser.printExample(OptionHandlerFilter.REQUIRED));
+      return;
+    }
+
+    new IndexRDFCollection(indexRDFCollectionArgs).run();
+  }
```
rosequ: Put the `main()` method at the end of the file for consistency...

src/main/java/io/anserini/index/generator/LuceneRDFDocumentGenerator.java
```diff
@@ -0,0 +1,163 @@
+package io.anserini.index.generator;
+
+import io.anserini.document.RDFDocument;
+import io.anserini.document.SourceDocument;
+import io.anserini.index.IndexRDFCollection;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.lucene.document.*;
```
rosequ: Avoid wildcard imports.

src/main/java/io/anserini/search/SearchRDF.java
```diff
@@ -0,0 +1,135 @@
...
+    public String index;
+
+    @Option(name = "-subject", metaVar = "[Subject URI]", required =  true, usage = "entity subject URI")
+    public String subject;
+
+    // Optional arguments
+
+    @Option(name = "-property",
+            metaVar = "[Property name]",
+            usage = "property name to search")
```
rosequ: write this in a single line.

src/test/java/io/anserini/index/generator/LuceneRDFDocumentGeneratorTest.java
```diff
@@ -0,0 +1,43 @@
+package io.anserini.index.generator;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
rosequ: avoid wildcard imports

src/main/java/io/anserini/search/SearchRDF.java
```diff
@@ -0,0 +1,135 @@
...
+    } catch (CmdLineException e) {
+      System.err.println(e.getMessage());
+      parser.printUsage(System.err);
+      System.err.println("Example command: "+ SearchRDF.class.getSimpleName() +
+              parser.printExample(OptionHandlerFilter.REQUIRED));
+      return;
+    }
+
+    new SearchRDF(searchArgs.index).search(searchArgs.subject, searchArgs.property);
+  }
```
rosequ: Put `main` at the end for consistency with the other code.

src/main/java/io/anserini/collection/RDFCollection.java
```diff
@@ -0,0 +1,131 @@
+package io.anserini.collection;
+
+import io.anserini.document.RDFDocument;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.*;
```
rosequ: Avoid wildcard imports

src/main/java/io/anserini/document/RDFDocument.java
minafarid: NTriples are in the format `s p o .`, so I am rendering triples in the same way they were read.

rosequ:
Also, can you add the corresponding documentation?

src/main/java/io/anserini/document/RDFDocument.java
rosequ: oh okay. I wanted to confirm that it wasn't a typo :)

src/main/java/io/anserini/document/RDFDocument.java
minafarid: ok

src/main/java/io/anserini/index/generator/LuceneRDFDocumentGenerator.java
minafarid: OK... intellij automatically did that. will fix it

src/main/java/io/anserini/search/SearchRDF.java
minafarid: ok good point.

minafarid:
Where do we add documentation? can you send me a reference/example?

rosequ:
You can add it [here](https://github.com/castorini/Anserini/tree/master/docs). You can follow [this](https://github.com/castorini/Anserini/blob/master/docs/experiments-gov2.md) format.

minafarid:
Awesome, will do this.

src/main/java/io/anserini/collection/RDFCollection.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Class representing an instance of an RDF collection.
+ * RDF datasets can represent knowledge bases such as Freebase.
+ * They often come in a single .gz file.
+ */
+public class RDFCollection extends Collection<RDFDocument> {
```
lintool: By strict Java conventions, this should be RdfCollection, but I don't care that much...

src/main/java/io/anserini/index/IndexRDFCollection.java
```diff
@@ -0,0 +1,224 @@
...
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.analysis.en.EnglishAnalyzer;
+import org.apache.lucene.codecs.lucene50.Lucene50StoredFieldsFormat;
+import org.apache.lucene.codecs.lucene62.Lucene62Codec;
+import org.apache.lucene.document.Document;
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.FSDirectory;
+import org.kohsuke.args4j.*;
```
lintool: another wildcard import - search for all others...

src/main/java/io/anserini/index/IndexRDFCollection.java
```diff
@@ -0,0 +1,224 @@
...
+   * @param args program arguments
+   * @throws Exception
+   */
+  public IndexRDFCollection(Args args) throws Exception {
+
+    // Copy arguments
+    this.args = args;
+
+    // Initialize analyzer
+    String defaultAnalyzerName = "org.apache.lucene.analysis.standard.StandardAnalyzer";
```
lintool: why Class.forName? why not just create a static final instance?

lintool:
Nice, this is exactly what I wanted as a start. Let's get this merged in once minor issues have been addressed.

src/main/java/io/anserini/index/IndexRDFCollection.java
minafarid: done

src/main/java/io/anserini/index/IndexRDFCollection.java
minafarid: I wanted to allow opportunity for other analyzers, but I will fix it for now.

minafarid: Actually, I'll remove it since I am storing RDF triple components as-is, not analysis is needed for now.

PR #179, opened 2017-05-04
updated demo

rosequ:
- simplified the demo to be compatible with the new bridge
- option to choose to output from the `PassageScorer` or `sm_cnn`

### DIFF
--- a/src/main/python/api.py
+++ b/src/main/python/api.py
@@ -6,15 +6,10 @@ import sys
 from flask import Flask, jsonify, request
 # FIXME: separate this out to a classifier class where we can switch out the models
 from pyserini import Pyserini
-from sm_model.bridge import SMModelBridge
-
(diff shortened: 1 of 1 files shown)


rosequ:
@tuzhucheng please review

src/main/python/api.py
```diff
@@ -44,23 +40,40 @@ def wit_ai_config():
...
+    else:
+        # the re-ranking model chosen is idf
+        sorted_answers = list(candidate_sent_scores)
+
+    print("in idf:{}".format(sorted_answers))
     answers = []
-
-    for score, sent in answers_list:
+    for score, sent in sorted_answers:
+        print(sent)
```
tuzhucheng: Should this really be printed or is it for debugging purposes only?

rosequ: oh yeah, it's just for debugging. Let me remove it...

PR #181, opened 2017-05-06
we can now specify a particular model file for e2e

gauravbaruah:
as part of the sourcing-IDF-from-index and e2e experiments.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,5 @@ pandas==0.19.2
 Flask==0.12.1
 nltk==3.2.2
 py4j==0.10.4
--e git+https://github.com/castorini/Castor.git#egg=sm-cnn-1.0.0
+#-e git+https://github.com/castorini/Castor.git#egg=sm-cnn-1.0.0
+-e git+https://github.com/gauravbaruah/Castor.git@corpus-idf#egg=sm_cnn-1.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,4 @@ pandas==0.19.2
 Flask==0.12.1
 nltk==3.2.2
 py4j==0.10.4
-#-e git+https://github.com/castorini/Castor.git#egg=sm-cnn-1.0.0
--e git+https://github.com/gauravbaruah/Castor.git@corpus-idf#egg=sm_cnn-1.0.0
+-e git+https://github.com/castorini/Castor.git#egg=sm-cnn-1.0.0
--- a/src/main/python/run_trecqa.py
+++ b/src/main/python/run_trecqa.py
@@ -11,11 +11,10 @@ class QAModel:
   def __init__(self, model_choice, index_path, qa_model_file=None):
     if not QAModel.instance:
        path = os.getcwd() + '/..'
-       if model_choice == "sm":
-         model_file = path + '/models/sm_model/sm_model.fixed_ext_feats_paper.puncts_stay'
-         if qa_model_file:
-           model_file = qa_model_file
-         QAModel.instance = SMModelBridge(model_file, 
+       if model_choice == "sm":         
+         if not qa_model_file:
+           qa_model_file = path + '/models/sm_model/sm_model.fixed_ext_feats_paper.puncts_stay'
+         QAModel.instance = SMModelBridge(qa_model_file, 
                              path + '/data/word2vec/aquaint+wiki.txt.gz.ndim=50.cache',
                              index_path)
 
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -4,4 +4,5 @@ pandas==0.19.2
 Flask==0.12.1
 nltk==3.2.2
 py4j==0.10.4
--e git+https://github.com/castorini/Castor.git#egg=sm-cnn-1.0.0
+#-e git+https://github.com/castorini/Castor.git#egg=sm-cnn-1.0.0
+-e git+https://github.com/gauravbaruah/Castor.git@corpus-idf#egg=sm_cnn-1.0.0
```
rosequ: Uncomment line 7 and delete line 8 because once the PR is merged we want the latest `sm_cnn` module to be same as the one found at https://github.com/castorini/Castor.git.

src/main/python/run_trecqa.py
```diff
@@ -150,6 +161,7 @@ def get_precision(actual, predicted, k):
   parser.add_argument("-k", help="top-k passages", default=10)
   parser.add_argument("-model", help="[idf|sm]", default='idf')
   parser.add_argument('-index', help="path of the index", required=True)
+  parser.add_argument('--qa-model-file', help="the path to the model file")
```
rosequ: use a single `-` for consistency.

src/main/python/run_trecqa.py
```diff
@@ -6,24 +6,35 @@
...
   idf_json = pyserini.get_term_idf_json()
   candidate_passages = []
   tokeninzed_answer = []
 
   if model_choice == 'sm':
-    path = os.getcwd() + '/..'
 
-    model = SMModelBridge( path +'/models/sm_model/sm_model.fixed_ext_feats_paper.puncts_stay',
-                          path + '/data/word2vec/aquaint+wiki.txt.gz.ndim=50.cache',index_path)
+    qa_DL_model = QAModel(model_choice, index_path, qa_model_file).instance
```
rosequ: nice one! should we rename this to `qa_dl_model`?

src/main/python/run_trecqa.py
```diff
@@ -6,24 +6,35 @@
...
+  instance = None
+  def __init__(self, model_choice, index_path, qa_model_file=None):
+    if not QAModel.instance:
+       path = os.getcwd() + '/..'
+       if model_choice == "sm":
+         model_file = path + '/models/sm_model/sm_model.fixed_ext_feats_paper.puncts_stay'
+         if qa_model_file:
+           model_file = qa_model_file
+         QAModel.instance = SMModelBridge(model_file, 
+                             path + '/data/word2vec/aquaint+wiki.txt.gz.ndim=50.cache',
```
rosequ: I like this one better: 
```python
if model_choice == "sm":
  if not qa_model_file:
    qa_model_file = path + '/models/sm_model/sm_model.fixed_ext_feats_paper.puncts_stay'
  QAModel.instance = SMModelBridge(qa_model_file, ...
```

rosequ:
Also, please update the README

src/main/python/run_trecqa.py
gauravbaruah: I'd like to think that "DL" enforces that it represents an acronym (deep learning). For instance we could have qa_IDF_model, qa_BM25_model, qa_PRM_model (positional relevance)....
Do let me know your views.

src/main/python/run_trecqa.py
gauravbaruah: AFAIK the https://docs.python.org/3/library/argparse.html convention is:
- no dashes are for required positional arguments. (script.py arg1 arg2)
- double dashes "--" are for switches that make the program change its operation. (script.py arg1 arg2 --run-it-this-way). Switches could have parameters  as well. (--num-iterations  4 or --use-this-file filename)
- single dash "-" is usually prescribed as a "shortcut" for switches. (-h for --help)

However I see that you have used the required=True flag. Note that the --qa-model-file is a switch, i.e. the program can run without it. 
Both approaches to arguments can work IMO. However, note that there now exist different conventions in Anserini and Castor. Do we want to keep them same? or, Do we want to follow respective conventions?

rosequ: @gauravbaruah yes it's true that `--` is used for switches. But, we have been using single `-` with default values for optional arguments. To be consistent, I think we should use `-` and use default values maybe?

gauravbaruah: Just to confirm, do you want the -qa-model-file to be required argument?
This means that run_trecqa.py will not pick up the hardcoded model from castorini/models/.

rosequ: No, let's not make it `required`. This is just to follow the existing conventions throughout Anserini and Castor.

Issue #182, opened 2017-05-08
Generate training data from Freebase index


minafarid:
* We need to retrieve training examples from Freebase index.
* The training examples can be used in machine learning applications.
* Depending on the target property, the retrieval queries will change to reflect how data is represented in Freebase RDF triples.

Peilin-Yang:
@lintool Is this still relevant?

lintool:
Closing, Freebase code removed in #329 - no longer relevant

PR #183, opened 2017-05-09
Issue #182 - Generate training examples

minafarid:
* Created `TrainingDataGenerator` to retrieve training examples from
  Freebase index.
* Implemented training data retrieval for `birthdate` property.
* Unit tests
* Added relevant docs.

Signed-off-by: Mina Farid <mina.farid@uwaterloo.ca>


src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -0,0 +1,360 @@
...
+ * Generate training data (positive examples) of a particular
+ * relationship from an indexed RDF dataset, e.g., Freebase.
+ */
+public class TrainingDataGenerator {
+
+  private static final Logger LOG = LogManager.getLogger(TrainingDataGenerator.class);
+  private final Logger TRAINING_DATA_OUTPUT_FILE_EXAMPLES;
+
+  private static final String[] supportedProperties = new String[]{
+          "birthdate"
```
rosequ: space after `[]`

src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -0,0 +1,360 @@
...
+    }
+  }
+
+  /**
+   * Generate training data for property birth date
+   * <p>
+   * Note: this function might need some refactoring when we add more properties
+   */
+  void birthdate() throws Exception {
+    QueryParser queryParser = new QueryParser(FIELD_BIRTHDATE, getKbIndexAnalyzer());
```
rosequ: can we throw a specific exception

src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -0,0 +1,360 @@
...
+    queryParser.setAllowLeadingWildcard(true);
+
+    Query q = queryParser.parse("*");
+
+    LOG.info("Starting the search using query: {}", q.toString());
+
+    // Collect all matching documents in a set of matching doc ids
+    Set<Integer> matchingDocIds = new HashSet<>(5);
+    getKbIndexSearcher().search(q, new CheckHits.SetCollector(matchingDocIds));
```
rosequ: why 5?

src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -0,0 +1,360 @@
...
+  private void generateTrainingData() throws Exception {
+    switch (args.property.toLowerCase()) {
+      case "birthdate":
+        birthdate();
+        break;
+
+      default:
+        LOG.error("Cannot generate training data for property: {}", args.property);
+        throw new IllegalArgumentException("Cannot generate training data for property: " + args.property + ". Supported properties are: " + supportedPropertiesStr);
+    }
```
rosequ: break into 2 lines

src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -0,0 +1,360 @@
...
+    l.info(sb.toString());
+  }
+
+  /**
+   * Determine which properties to retrieve and runs the training data retrieval pipeline.
+   *
+   * @throws Exception on error
+   */
+  private void generateTrainingData() throws Exception {
+    switch (args.property.toLowerCase()) {
```
rosequ: throw specific exception

src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  /**
+   * Writes an array as a line separated by TRAINING_DATA_SEPARATOR.
+   *
+   * @param l    the logger to write to (the training data file)
+   * @param data array of data
+   */
+  void writeToTrainingFiles(Logger l, String... data) {
+    StringBuilder sb = new StringBuilder();
```
rosequ: Did you mean `String[]`?

src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -0,0 +1,360 @@
...
+  IndexSearcher kbIndexSearcher = null;
+
+  /**
+   * Index analyzer to search
+   */
+  Analyzer kbIndexAnalyzer;
+
+  /**
+   * Simple value factory to parse literals using Sesame library.
+   */
```
rosequ: make these private variables and initialize them in a constructor?

src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -0,0 +1,360 @@
...
+      for (String birthdate : birthdates) {
+        // Get string value
+        String birthdateVal = extractValueFromTypedLiteralString(birthdate);
+
+        // Write property value as training data
+        writeToTrainingFiles(TRAINING_DATA_OUTPUT_FILE_EXAMPLES, freebaseId, englishLabel, birthdateVal);
+      }
+
+      // TODO - After building an index for the mentions of Freebase entities in ClueWeb, we need to get the ClueWeb mentions of this freebase entity and write them to a separate file
+    });
```
rosequ: break into 2 lines

src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -0,0 +1,360 @@
...
+   * Get the current default index analyzer or create it.
+   * Use EnglishAnalyzer as the default.
+   *
+   * @return kb index analyzer
+   * @throws Exception
+   */
+  Analyzer getKbIndexAnalyzer() throws Exception {
+    if (kbIndexAnalyzer == null) {
+      kbIndexAnalyzer = new EnglishAnalyzer();
+    }
```
rosequ: Using `EnglishAnalyzer()` without any parameters assumes a default stopword list. If that's not okay, use `EnglishAnalyzer(CharArraySet.EMPTY_SET)` instead.

src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -0,0 +1,360 @@
...
+
+  /**
+   * Determine which properties to retrieve and runs the training data retrieval pipeline.
+   *
+   * @throws Exception on error
+   */
+  private void generateTrainingData() throws Exception {
+    switch (args.property.toLowerCase()) {
+      case "birthdate":
+        birthdate();
```
rosequ: I prefer Java reflection to decide what method to be called at runtime.  Let's use [getMethod()](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getMethod-java.lang.String-java.lang.Class...-)  to decide what method to call based on the property selected by the user?

src/main/java/io/anserini/util/TrainingDataGenerator.java
minafarid: OK, I will change it as you suggested.

src/main/java/io/anserini/util/TrainingDataGenerator.java
minafarid: No, `String...` allows for variable number of parameters without the need to create an array explicitly from the calling function.
It is translated to String[] anyway.

src/main/java/io/anserini/util/TrainingDataGenerator.java
minafarid: OK, checking it out

src/main/java/io/anserini/util/TrainingDataGenerator.java
minafarid: No specific reason, default initial capacity is 16 and load factor is 0.75. I removed the `initialCapacity` anyway

minafarid:
Done as suggested by Royal, minus the `String...` comment.

src/main/java/io/anserini/util/TrainingDataGenerator.java
```diff
@@ -13,27 +13,27 @@
...
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.store.FSDirectory;
-import org.kohsuke.args4j.CmdLineException;
-import org.kohsuke.args4j.CmdLineParser;
-import org.kohsuke.args4j.Option;
-import org.kohsuke.args4j.OptionHandlerFilter;
-import org.kohsuke.args4j.ParserProperties;
+import org.kohsuke.args4j.*;
```
rosequ: remove wildcard imports. IntelliJ does it!

src/main/java/io/anserini/util/TrainingDataGenerator.java
rosequ: Oh nice, I learned something new today :)

src/main/java/io/anserini/util/TrainingDataGenerator.java
rosequ: What I meant was, can we initialize the necessary member variables such  `kbIndexDirectory`, `kbIndexReader` etc. in a constructor instead of initializing them with individual methods? something like [this](https://github.com/rosequ/Anserini/blob/master/src/main/java/io/anserini/index/IndexUtils.java#L82)

src/main/java/io/anserini/util/TrainingDataGenerator.java
minafarid: yup, fixed

PR #191, opened 2017-05-30
Freebase Entity Index

salman1993:
- builds up the index for freebase entities
- indexes three fields: 1) wikipedia_en_title, 2) rdf#label, 3) wikipedia_en (may contain variants)
- search class to look up an entity ID given a query


lintool:
You inadvertently checked-in a whole bunch of things that should be there, e.g., `freebase.sample.index/`, `freebase-subject-obama.gz`

src/main/java/io/anserini/index/generator/LuceneFreebaseEntityDocumentGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+    // Index subject as a StringField to allow searching
+    Field entityField = new StringField(FIELD_ENTITY, cleanUri(tripleDoc.getEntityId()), Field.Store.YES);
+    doc.add(entityField);
+
+    FieldType fieldType = new FieldType();
+    fieldType.setStored(true);
+    fieldType.setStoreTermVectors(true);
+    fieldType.setIndexOptions(IndexOptions.DOCS_AND_FREQS);
+
+    Field titleField = new TextField(FIELD_TITLE, tripleDoc.getTitle(), Field.Store.YES);
```
lintool: What's the difference between the `Field_TITLE` and `FIELD_LABEL`?

src/main/java/io/anserini/search/SearchFreebaseEntity.java
```diff
@@ -0,0 +1,135 @@
...
+import org.kohsuke.args4j.*;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
+
+import static io.anserini.index.generator.LuceneDocumentGenerator.FIELD_BODY;
```
lintool: I don't think you need these?

src/main/java/io/anserini/index/generator/LuceneFreebaseEntityDocumentGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+import io.anserini.index.IndexFreebaseEntityCollection;
+import org.apache.lucene.document.*;
+import org.apache.lucene.index.IndexOptions;
+
+/**
+ * Converts a {@link RDFDocument} into
+ * a Lucene {@link Document}.
+ */
+public class LuceneFreebaseEntityDocumentGenerator {
+  public static final String FIELD_ENTITY = "entity";
```
lintool: document what each of these fields are?

lintool:
Additional comment - I think we should use the Freebase terminology more precisely. I believe "entities" are called "topics" - so this should be called "FreebaseTopicIndex". We should also use `mid`, etc.

src/main/java/io/anserini/index/generator/LuceneFreebaseEntityDocumentGenerator.java
salman1993: I'll put in comments and in the docs

salman1993:
@minafarid - yeah i'm sure about the dots format. This is what it says in https://developers.google.com/freebase/:
"Note: In Freebase, objects have MIDs that look like /m/012rkqx. In RDF those MIDs become m.012rkqx."

minafarid:
@salman1993 my problem was in the prefix of the URI, not the dot notation. But it is okay if the data is represented this way.

PR #193, opened 2017-06-03
Freebase Topic index

salman1993:
- added code to index topics
- added an example lookup code (still needs to be fixed)
- added docs


src/main/java/io/anserini/kg/freebase/Topic.java
```diff
@@ -0,0 +1,169 @@
...
+   * @param o object value
+   */
+  public void addPredicateAndValue(String p, String o) {
+    p = cleanUri(p);
+
+    if (p.startsWith(WIKI_EN_URI)) {
+      if (p.startsWith(WIKI_EN_TILE_URI)) {
+        this.title = removeQuotes(o);
+      }
+      else {
```
minafarid: According to the coding style we are using, I think you should move the `else` next to the closing bracket of the `if`. Same for `catch` phrases.

src/main/java/io/anserini/kg/freebase/Topic.java
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Splitter that describes how s,p,o are split in a triple line
+   */
+  public static final String TRIPLE_SPLITTER = "\t";
+
+  /**
+   * Predicates for which the literals should be stored
+   */
+  public static final String WIKI_EN_URI = "http://rdf.freebase.com/key/wikipedia.en";
+  public static final String WIKI_EN_TILE_URI = "http://rdf.freebase.com/key/wikipedia.en_title";
```
minafarid: The `if` statements in lines 88 and 89 assumes that `WIKI_EN_URI` is the base string for `WIKI_EN_TILE_URI` (which has a typo). You should reflect that in the creation of the strings.

This should be changed to 
```java
WIKI_EN_URI = "htp://rdf..../wikipedia.en";
WIKI_EN_TITLE_URI = WIKI_EN_URI + "_title";
```

src/main/java/io/anserini/kg/freebase/Topic.java
```diff
@@ -0,0 +1,169 @@
...
+    else
+      return uri;
+  }
+
+  /**
+   * Removes quotes from the literal value in object field
+   */
+  private String removeQuotes(String literal) {
+    if (literal.charAt(0) == '\"' && literal.charAt(literal.length()-1) == '\"')
+      return literal.substring(1, literal.length()-1);
```
minafarid: I always add opening and closing brackets, `{` and `}` for if statements, even if they have one line body. This improves the readability imo.

src/main/java/io/anserini/kg/freebase/TopicLuceneDocumentGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+    Document doc = new Document();
+
+    // Index subject as a StringField to allow searching
+    Field topicMidField = new StringField(FIELD_TOPIC_MID, cleanUri(src.getTopicMid()), Field.Store.YES);
+    doc.add(topicMidField);
+
+    FieldType fieldType = new FieldType();
+    fieldType.setStored(true);
+    fieldType.setStoreTermVectors(true);
+    fieldType.setIndexOptions(IndexOptions.DOCS_AND_FREQS);
```
minafarid: `fieldType` is not a used variable, remove

src/main/java/io/anserini/kg/freebase/IndexTopic.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Builds a triples lookup index from a Freebase dump in N-Triples RDF format. Each
+ * {@link ObjectTriples} object, which represents a group of triples that share the same subject,
+ * is treated as a Lucene "document". This class builds an index primarily for lookup by
+ * <code>mid</code>.
+ */
+public class IndexTopic {
```
lintool: Should be IndexTopics (plural)

src/main/java/io/anserini/kg/freebase/IndexTopic.java
```diff
@@ -0,0 +1,184 @@
...
+    transformer.config(args);
+    transformer.setCounters(counters);
+
+    int cnt = 0;
+    ObjectTriplesIterator iter = new ObjectTriplesIterator(inputFile);
+    while (iter.hasNext()) {
+      ObjectTriples d = iter.next();
+      // make a Topic from the ObjectTriples
+      String subject = d.getSubject();
+      Topic topic = new Topic(subject);
```
lintool: Can you explain why we need a separate "Topic" object here?
Can't we just take an ObjectTriples and generate the proper Lucene document?
In other words, does it make sense that the "Object" needs to have an independent existence?

salman1993: ObjectTriples does not have (subject, predicate, object). It has 
(subject, map<predicates, list<objects>>) so I have to create a topic for each subject and get the required predicate values from the triples and put it in the topic.
What will be a better way to do this?

src/main/java/io/anserini/kg/freebase/IndexTopics.java
```diff
@@ -0,0 +1,184 @@
...
+    transformer.config(args);
+    transformer.setCounters(counters);
+
+    int cnt = 0;
+    ObjectTriplesIterator iter = new ObjectTriplesIterator(inputFile);
+    while (iter.hasNext()) {
+      ObjectTriples d = iter.next();
+      // make a Topic from the ObjectTriples
+      String subject = d.getSubject();
+      Topic topic = new Topic(subject);
```
lintool: I guess my suggestion was to fold this into `TopicLuceneDocumentGenerator`. Topic is always transient, right? I.e., we get out ObjectTriples, we convert it into Topic, and then throw it away? Wondering if we can remove a class altogether?

minafarid: I think we should remove ObjectTriples... Originally, I created it as a `Map`. Perhaps we can modify `ObjectTriplesIterator` to `TopicIterator` that returns a `Topic` object when `.next()` is called.

lintool: To be precise, in Freebase terminology, each "node" in the graph is called an object. Objects can either be topics or CVTs. So renaming `ObjectTriples` to `Topic` isn't accurate. I don't know if there's a better name. `FreebaseNode`?

minafarid: I believe @salman1993 is using `Topic` to capture _all_ information about a particular subject (i.e., entity/topic/concept). Information is encoded in triples and is referred to as _facts_.

salman1993: @lintool i put it in TopicLuceneDocumentGenerator in the new commit.

lintool:
@salman1993 I still see `src/main/java/io/anserini/kg/freebase/Topic.java` as part of the current commit?

salman1993:
removed Topic.java and moved all the logic to lucene generator

Issue #188, opened 2017-05-22
Refactoring Freebase API


lintool:
The current Freebase API written by @minafarid tries to mimic the collection API for indexing document collections by creating the notion of the RDF collection. I think this is a mistake and trying to jam disparate things together (RDF and text documents). Furthermore, I doubt that we have the general abstraction of what an "RDF collection" looks like anyway.

I propose that we give up this pretense and make the code explicitly about Freebase, and then generalize later as appropriate. In concrete terms, lets move all the Freebase code into a package called `io.anserini.kg.freebase` - `kg` for knowledge graph. Then we can develop `io.anserini.kg.yago` for Yago when we need do, and other knowledge graphs, etc.

minafarid:
Our definition of a document looks like the following:

* Each document contains information about **one** single entity or concept.
* The `docid` is the entity/subject URI, `subject_uri`. This makes sense because each entity or concept in RDF has a unique URI.
* We group all the predicates and their values of a particular entity in one document.
* Each predicate (sometimes referred to as property) forms a field name, and the field value is the object value of this predicate.

I believe this abstraction is standard across multiple RDF datasets. It is generic enough to capture how different datasets might represent/capture information. However, the querying functions (e.g., finding alma mater of people) will be done differently according to how this information is encoded.

The only freebase-specific implementation that we have is the indexing process. The indexing process assumes that triples are ordered by the subject URI, i.e., all the triples of a particular entity URI appear in sequence. This allows us to read a list of triples that belong to the same entity and index them in one step as one document.

A more generic implementation is to index every single triple as a separate lucene document. I tried it, but this won't work because we might have billions of triples and it would exceed lucene's limit of the number of documents in an index.

lintool:
@minafarid I think I agree with your comment. However, my main complaint was about jamming the "RDF document" into the same set of APIs as text documents... I don't think that's productive, and actually quite confusing...

minafarid:
Yes, it might be confusing for sure since most of the search is just lookups and emulating conjunctive queries in code. We don't really "search" in the documents, but we rather do direct lookups.

lintool:
Summarizing discussion on Slack, the API we converged on was `Freebase implements Iterable<FreebaseNode>`. A `FreebaseNode` is a node in the Freebase knowledge graph - in Freebase terminology, it's called an object: https://developers.google.com/freebase/guide/basic_concepts

However, "object" is too overloaded. Note that "objects" can be topics or CVTs.

[...]

PR #194, opened 2017-06-07
Freebase refactoring: Freebase implements Iterable<FreebaseNode>

lintool:
Per #188

Revised API: Freebase implements Iterable<FreebaseNode>. A FreebaseNode is a node in the Freebase knowledge graph. The API allows us to iterate over nodes - each node contains all the outgoing edges, i.e., all facts about that object. FreebaseNode replaces what we're calling ObjectTriples currently.


salman1993:
Under docs/ there are two files "index-freebase-topic.md" and "index-freebase-topics.md". Can you please remove one of them? We can merge this in after that. All the changes look good to me!

PR #196, opened 2017-06-15
Freebase IndexNodes refactoring

lintool:
Refactoring of IndexNodes:
+ Simplified code into more readable, functional style.
+ Moved common DocumentGenerator functionality into FreebaseNode


src/main/java/io/anserini/kg/freebase/FreebaseNode.java
```diff
@@ -125,4 +129,30 @@ public void clear() {
...
+
+  public static RdfObjectType getObjectType(String objectValue) {
+    // Determine the type of this N-Triples 'value'.
+    switch (objectValue.charAt(0)) {
+      case '<':
+        // e.g., <http://rdf.freebase.com/ns/m.02mjmr>
+        return RdfObjectType.URI;
+      case '"':
+        if (objectValue.charAt(objectValue.length() - 1) == '"') {
+          // e.g., "Hanna Bieluszko"@en";
```
salman1993: is this RdfObjectType.STRING or RdfObjectType.TEXT? 
the test shows contradictory examples.

src/test/java/io/anserini/kg/freebase/FreebaseNodeTest.java
```diff
@@ -0,0 +1,28 @@
...
+        FreebaseNode.getObjectType("<http://rdf.freebase.com/ns/m.02mjmr>"));
+
+    assertEquals(FreebaseNode.RdfObjectType.TEXT,
+        FreebaseNode.getObjectType("\"1954-10-28\"^^<http://www.w3.org/2001/XMLSchema#date>"));
+
+    assertEquals(FreebaseNode.RdfObjectType.TEXT,
+        FreebaseNode.getObjectType("\"Hanna Bieluszko\"@en"));
+
+    assertEquals(FreebaseNode.RdfObjectType.STRING,
+        FreebaseNode.getObjectType("\"Hanna Bieluszko\""));
```
salman1993: test contradicts with comment above.

src/main/java/io/anserini/kg/freebase/FreebaseNode.java
salman1993: are we getting rid of the "@en" at the end? i think we should since we are only keeping English labels anyway.

lintool: Good call.

Re: getting rid of "@en" - I'm okay with that but let's do it in another CR with proper test cases... we should also figure out a consistent way to deal with multi-lingual stuff...

PR #198, opened 2017-06-18
Freebase test case, FreebaseNode refactoring

lintool:
+ Added Freebase test case for reading N-Triples format. Refreshed code comments.
+ Refactored FreebaseNode - note that the nodes are identified by URIs. URIs are not necessarily MIDs, as the previous API (falsely) suggested.


src/test/java/io/anserini/kg/freebase/FreebaseTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class FreebaseTest {
+  @Test(expected = NoSuchElementException.class)
+  public void test1() throws Exception {
+    Freebase freebase = new Freebase(Paths.get("src/test/resources/freebase-rdf-head100.gz"));
+
+    FreebaseNode node;
+    Iterator<FreebaseNode> iter = freebase.iterator();
+    assertTrue(iter.hasNext());
+    node = iter.next();
```
minafarid: This is fine for testing that the functionality works, but I'm not a big fan of this kind of tests. It is like we look at the results first and then write the tests accordingly. I'd rather we look up a node by it's MID then continue with the tests.

lintool: But I believe than any test is better than no test...
You're welcome to contribute tests also... :)

PR #202, opened 2017-06-29
Pattern based evaluation (lintool/fourdrinier#18)

rosequ:
- Evaluation of E2E QA with TREC QA patterns 
- Updated the pipeline
- Updated the docs
@snapbug please have a look.


src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+    idf_json = pyserini.get_term_idf_json()
+    candidate_passages = []
+    tokeninzed_answer = []
+
+    if model_choice == 'sm':
+
+        qa_DL_model = QAModel(model_choice, index_path, w2v_cache, qa_model_file).instance
+
+        for ps in candidate_passages_scores:
+            ps_split = ps.split('\t')
```
snapbug: You should be consistent with your strings, are they `'` strings, or `"` strings? Comment equally applies to other places with strings used.

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+        for score, candidate in sorted_answers:
+            tokens = TreebankWordTokenizer().tokenize(candidate.lower().split("\t")[0])
+            tokeninzed_answer.append((tokens, score))
+        return tokeninzed_answer
+
+    elif model_choice == 'idf':
+        for candidate in candidate_passages_scores:
+            candidate_sent, score = candidate.lower().split("\t")
+            tokens = TreebankWordTokenizer().tokenize(candidate_sent.lower())
+            tokeninzed_answer.append((tokens, score))
```
snapbug: These few lines look suspiciously similar to the last few lines of the prior `if` condition ... perhaps some refactoring would make them only written once.

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+        return tokeninzed_answer
 
 def jaccard(set1, set2):
-  intersection = set1.intersection(set2)
-  union = set1.union(set2)
-  return len(intersection) / float(len(union))
+    intersection = set1.intersection(set2)
+    union = set1.union(set2)
+    return len(intersection) / float(len(union))
```
snapbug: If `set1` and `set2` are python `set` objects, you can `return len(set1 & set2) / float(len(set1 | set2))`, it does the same thing, but is (marginally) clearer in my opinion. Also if this is python3 (which I think it is?) then you don't have to worry about integer division and so can remove the `float(...)`.

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+            label = re.match('^<(positive|negative)>', prev)
 
-      if label:
-        label = label.group(1)
-        label = 1 if label == 'positive' else 0
-        answer = line.split('\t')
-        answer_count += 1
+            if label:
+                label = label.group(1)
+                label = 1 if label == 'positive' else 0
```
snapbug: can be replaced by `label = label == 'positive'`

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
 
-def get_precision(actual, predicted, k):
-  total_count = 0.0
-  hits = 0.0
+    for i, cand in enumerate(candidates):
+        this_candidate = " ".join(cand[0])
 
-  for qid in predicted.keys():
-    run_labels = predicted.copy()
+        hash_value = hashlib.md5(this_candidate.encode()).hexdigest()
```
snapbug: Add a comment about why you're using the md5 hash, I can think of a reason, but I don't know, and it looks funky.

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
-    for p in run_labels:
-      total_count += 1
+        out.write('{} Q0 {} {} {} TrecQA-pattern\n'.format(qid, answer_id, i+1, 1.0/(i + 1)))
 
-      if not actual[qid]:
-        continue
+# evaluate the run file at different depths
+def evaluate_at_k(fname, eval_depth):
+    for k in eval_depth:
+        with open(fname) as f,  open("{}.k{}.txt".format(fname, k), 'w') as output_file:
```
snapbug: extra spacing after `f,`

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+    for i, cand in enumerate(all_sentences):
+        this_candidate = cand.lower().split("\t")[0]
+        result = re.findall(r'{}'.format(pattern.lower()), this_candidate)
+        hash_value = hashlib.md5(this_candidate.encode()).hexdigest()
+        answer_id = "QA{}.{}".format(qid, hash_value)
+        if answer_id in seen:
+            continue
+        seen.add(answer_id)
+
+        if result:
```
snapbug: You can collapse these, using `1 if result else 0` inside the format ... I believe it's more pythonic, it certainly reduces the possibility of changing one format, but not the other. I'd also suggest using `len(result)` just to be a bit more explicit that result is not a bool

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+            cleaned_sentences.append(sent)
+
+    return cleaned_sentences
+
+# load the stopwords from the resource file
+def get_stopwords():
+    stopwords = set([])
+    with open("src/main/resources/io/anserini/qa/english-stoplist.txt") as stop:
+        for line in stop:
+            if "#" not in line:
```
snapbug: `# not in line` or `not line.startswith("#")`? There's a subtle difference and even if it doesn't matter in this case (which I suspect it doesn't), the intent should be made clearer.

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+                for line in pattern_file:
+                    det = line.strip().split(' ', 1)
+
+                    if det[0] not in pattern_dict:
+                        pattern_dict[det[0]] = det[1]
+
+    pyserini = Pyserini(args.index)
+    questions, answers, labels_actual = load_data(args.input)
+    threshold = 0.7
+    # TODO: what is this ^^ magic number
```
snapbug: You said it 😄

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+
+    qrel_file = ""
+
+    if args.qrel:
+        qrel_file = args.qrel
+    elif args.eval == 'jaccard':
+        qrel_file = '{}/qrel.jaccard.txt'.format(args.output)
+        create_qrel_jaccard(qrel_file, answers)
+    elif args.eval == 'pattern':
+        qrel_file = '{}/qrel.pattern.txt'.format(args.output)
```
snapbug: Double check that the description of `args.output` in the earlier file matches how it's used here -- my recollection is that it isn't.

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+                    print(e)
+                    print("pattern not found")
+
+    output_file = '{}/run.qa.{}.{}.h0_{}.h1_{}'.format(args.output, args.model,  args.eval, args.h0, args.h1)
+    seen_docs = []
+
+    with open(output_file, 'w') as out:
+        for qid, question in zip(answers.keys(), questions):
+            # sentence re-ranking after ad-hoc retrieval
+            candidates = get_answers(pyserini, question, int(args.h0), int(args.h1), args.model, args.index, args.w2v_cache,
```
snapbug: Without measuring, I feel like this line should be wrapped earlier?

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+                                   args.qa_model_file)
+
+            if args.eval == 'jaccard':
+                scored_candidates = score_candidates(candidates, answers[qid])
+                i, unjudged_count = 0, 0
```
snapbug: Is there a better name than `i`?

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
 from nltk.tokenize import TreebankWordTokenizer
 from pyserini import Pyserini
 from sm_cnn.bridge import SMModelBridge
+import hashlib
```
snapbug: As you're only using `md5` from `hashlib`, I suggest `from hashlib import md5`.

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
 
 def jaccard(set1, set2):
-  intersection = set1.intersection(set2)
-  union = set1.union(set2)
-  return len(intersection) / float(len(union))
+    intersection = set1.intersection(set2)
+    union = set1.union(set2)
+    return len(intersection) / float(len(union))
 
 def score_candidates(candidates, answers):
```
snapbug: There are quite a few "magic" `0` and `1` indices here, a `namedtuple` for example would fix this quite easily.

src/main/python/run_trecqa.py
```diff
@@ -5,219 +5,294 @@
...
+                if not args.pattern:
+                    print("Path to the pattern file should be included if the method of evaluation is pattern based")
+                    args.print_help()
+                    exit()
+                try:
+                    eval_by_pattern(qid, candidates, pattern_dict[qid], out)
+                except KeyError as e:
+                    print("Pattern not found for question: {}".format(e))
+
+        eval_depth = [int(i) for i in args.k.strip().split()]
```
snapbug: Could be replaced by `map(int, args.k.strip().split())` ... also I'm not sure how argparse deals with space separated arguments? I assume that this works as intended.

src/main/java/io/anserini/qa/passage/IdfPassageScorer.java
```diff
@@ -101,14 +118,19 @@ public void score(String query, Map<String, Float> sentences) throws Exception {
...
             seenTerms.add(t.toString());
           } else {
             idf += 0.0;
           }
+
+          TermQuery q2 = (TermQuery) qp2.parse(term);
+          Term t2 = q2.getTerm();
+          double termIDFwithStop = similarity.idf(reader.docFreq(t2), reader.numDocs());
+
+          termIdfMap.put(term, String.valueOf(termIDFwithStop));
```
snapbug: _Why_ is `termIdfMap` storing the `String.valueOf` and not the double directly?

src/main/java/io/anserini/qa/passage/IdfPassageScorer.java
```diff
@@ -101,14 +118,19 @@ public void score(String query, Map<String, Float> sentences) throws Exception {
           TermQuery q = (TermQuery) qp.parse(term);
           Term t = q.getTerm();
           double termIDF = similarity.idf(reader.docFreq(t), reader.numDocs());
-          termIdfMap.put(term, String.valueOf(termIDF));
 
           if (questionTerms.contains(t.toString()) && !seenTerms.contains(t.toString())) {
             idf += termIDF;
             seenTerms.add(t.toString());
           } else {
             idf += 0.0;
```
snapbug: Just remove the `else` block entirely if it's doing nothing.

src/main/java/io/anserini/py4j/PyseriniEntryPoint.java
```diff
@@ -131,6 +131,27 @@ public String getRawDocument(String docid) throws Exception {
...
+    TokenizerFactory<CoreLabel> tokenizerFactory = PTBTokenizer.factory(new CoreLabelTokenFactory(), "");
+    List<String> allSentences = new ArrayList<String>();
+
+    for (Map.Entry<String, Float> doc : docScore.entrySet()) {
+      List<Sentence> sentences = indexUtils.getSentDocument(doc.getKey());
+      for (Sentence thisSent : sentences) {
+        List<CoreLabel> tokens = tokenizerFactory.getTokenizer(new StringReader(thisSent.text())).tokenize();
+        String tokenizedAnswer = tokens.stream()
+                .map(CoreLabel::toString)
+                .collect(Collectors.joining(" "));
```
snapbug: But java is perfectly capable of storing a `List<List>` ... this would also stop having the re-split the terms on the other end. If this is because of a limitation in the python/java interop, then *document* it, so that dumb people like me don't get confused.

docs/end2end-TrecQA-eval.md
```diff
@@ -13,7 +13,7 @@ if you need help setting up the gateway.
 Run the following command to evaluate the end2end system:
 
 ```
-python3 src/main/python/run_trecqa.py
+python src/main/python/run_trecqa.py
```
rosequ: @snapbug we decided on using Python3 for all our systems. Therefore, I updated this line. It's part of the requirements list.

docs/end2end-TrecQA-eval.md
```diff
@@ -22,8 +22,7 @@ Possible parameters are:
 -input (required)
 ```
 
-Path of a TrecQA file
-
+Path of the TrecQA test file. It can be downloaded from [here](http://cs.jhu.edu/~xuchen/packages/jacana-qa-naacl2013-data-results.tar.bz2)
```
rosequ: The link points to a compressed folder that has dev, test, and train files. We need all of them!

docs/end2end-TrecQA-eval.md
```diff
@@ -33,16 +32,22 @@ Path of the index
...
-Number of passages to be retrieved per question (default: 20)
+Top-h1 passages to be re-ranked by the answer extractor (default: 10)
 
 ```
--k (optional: positive integer)
+-k (optional: list of positive integers)
 ```
 
-Top-k passages presented by the system (default: 10)
+Evaluate at depth k (default: 10)
```
rosequ: Sorry for the terminology, it's supposed to mean create run files that have at most 5 answers. I'll have this updated.

src/main/java/io/anserini/qa/passage/IdfPassageScorer.java
```diff
@@ -101,14 +118,19 @@ public void score(String query, Map<String, Float> sentences) throws Exception {
           TermQuery q = (TermQuery) qp.parse(term);
           Term t = q.getTerm();
           double termIDF = similarity.idf(reader.docFreq(t), reader.numDocs());
-          termIdfMap.put(term, String.valueOf(termIDF));
 
           if (questionTerms.contains(t.toString()) && !seenTerms.contains(t.toString())) {
```
rosequ: The intuition is that the answer that have rare question terms are important. Earlier, while working on the E2E demo, we realized that considering the term frequency of the question terms lowered the quality of the output. This, perhaps, is due to the repetition of common question terms in the answers.

src/main/python/e2e_experimental_settings.py
```diff
@@ -31,34 +33,33 @@ def __init__(self, qa_data_xml, qrel, word_embeddings_file, index_path, model_fi
...
         self.settings[setting.label] = setting
         self._setup_combinations()
 
-
     def _setup_combinations(self):
         all_settings = []
         for setting in self.settings.values():
             all_settings.append(setting.get_options())
 
         self.combinations = []
```
rosequ: e2e_experimental_settings.py belongs to @gauravbaruah but I'll adopt these changes in the updated PR.

src/main/java/io/anserini/qa/passage/IdfPassageScorer.java
rosequ: But, we decided to do this based only on the quality of our answers in the demo and we didn't measure the effect quantitatively. I will look into separately.

docs/end2end-TrecQA-eval.md
snapbug: Right, you need them all at different times. The note there says "Path of **the** TrecQA **test** file", so perhaps "Path to the {dev,train,test} TrecQA file (from here) depending".

src/main/python/e2e_experimental_settings.py
```diff
@@ -105,85 +105,83 @@ def run(self, indices):
...
+        # ap.add_argument("--punctuation", choices=["keep", "remove"], default="keep")
+        # ap.add_argument("--dash-words", choices=["keep", "split"], default="keep")
+        # ap.add_argument("models_root", help="path to folder containing models")
+
+        args = ap.parse_args()
+
+        experiments = Experiments(args.qa_data_xml, args.qrel_file, args.word_embeddings_file,
+                                  args.index_path, args.model_file)
+
+        # experiments.add_setting(Setting('idf_source', {
```
tuzhucheng: It is best if we can control all the experiment settings from the command line arguments rather than comment/uncommenting code.

src/main/python/run_trecqa.py
rosequ: Nice suggestion!

src/main/python/run_trecqa.py
rosequ: No, it exceeds the limit, I checked :P

src/main/java/io/anserini/qa/passage/IdfPassageScorer.java
rosequ: JSONObject works only on Map<String, String>

src/main/java/io/anserini/qa/passage/IdfPassageScorer.java
```diff
@@ -75,10 +85,17 @@ public void score(String query, Map<String, Float> sentences) throws Exception {
...
+    HashSet<String> questionTermsIDF = new HashSet<>(Arrays.asList(questionWithStopWords.toString()
+            .trim().toLowerCase().split("\\s+")));
+
     // add the question terms to the termIDF Map
-    for (String questionTerm : questionTerms) {
+    for (String questionTerm : questionTermsIDF) {
       try {
-        TermQuery q = (TermQuery) qp.parse(questionTerm);
+        TermQuery q = (TermQuery) qp2.parse(questionTerm);
         Term t = q.getTerm();
```
rosequ: No, `questionTerm` would just be a string. The reason each term is parsed is because `similarity.idf()` expects the term to be an instance of `Term`

src/main/python/run_trecqa.py
```diff
@@ -293,6 +292,6 @@ def get_stopwords():
                 except KeyError as e:
                     print("Pattern not found for question: {}".format(e))
 
-        eval_depth = [int(i) for i in args.k.strip().split()]
+        eval_depth = map(int, args.k.strip().split())
```
tuzhucheng: In Python 3, map returns a generator and not a list, make sure this is what you want.

docs/end2end-TrecQA-eval.md
```diff
@@ -54,22 +59,44 @@ Model to re-rank the passages (default: idf)
 -output (optional: file path)
 ```
 
-Path of the run file to be created
+Path of the run file, qrel file to be created (default:.)
```
snapbug: This still needs addressing

docs/end2end-TrecQA-eval.md
```diff
@@ -54,22 +59,51 @@ Model to re-rank the passages (default: idf)
...
+This path is a required argument if the SM model is used as an answer extractor.
 
+```
+-w2v-cache (Optional: word2vec cache file path)
+```
+The path of the cache file is a required argument if the SM model is used as an answer extractor.
 
-The above command will create a run file in the `trec_eval` format and a qrel file
-called qrels.txt
+### Sample commands:
```
snapbug: Is there any reason that the example commands background the task? It might be surprising to the user if they just blindly copy-paste without examining what the command actually is. (More common behaviour than it should be)

src/main/java/io/anserini/py4j/PyseriniEntryPoint.java
snapbug: I think you might have addressed this in a reply to another of my comments, but the code still needs documenting here as well.

src/main/python/e2e_experimental_settings.py
```diff
@@ -5,23 +5,23 @@
...
         return self.choice_flags.keys()
 
     def get_choice(self, setting):
         return self.choice_flags[setting]
 
     def get_options(self):
-        options = []
-        for key in self.choice_flags.keys():
-            options.append("{}:{}".format(self.label, key))
+        options = [options.append("{}:{}".format(self.label, key)) for key in self.choice_flags.keys()]
```
snapbug: Did you check that this works? I don't think you did.

`list.append` returns `None`, which would mean that `options` ends up as `[None, None,...]`. Also, given that you're just then going to return the list, I'd suggest replacing by:
`return ["{}:{}".format(self.label, key) for key in self.choice_flags.keys()]`
Also if this doesn't need to be a list, then changing to a generator (by swapping the outermost `[]` with `()`) will be marginally more efficient

src/main/python/e2e_experimental_settings.py
```diff
@@ -31,34 +33,33 @@ def __init__(self, qa_data_xml, qrel, word_embeddings_file, index_path, model_fi
         self.w2v_cache = word_embeddings_file
         self.index_path = index_path
         self.model_file = model_file
-        self.cmd_root = "python src/main/python/run_trecqa.py -input {} -output e2e.run -qrel {} -index {} -w2v-cache {} -qa-model-file {} -model sm".\
+        self.cmd_root = "python src/main/python/run_trecqa.py -input {} -output e2e.run -qrel {} -index {} -w2v-cache {} -qa-model-file {} -model idf". \
```
snapbug: Again, this should be parameterized.

src/main/python/e2e_experimental_settings.py
```diff
@@ -31,34 +33,33 @@ def __init__(self, qa_data_xml, qrel, word_embeddings_file, index_path, model_fi
...
             format(self.qa_data_xml, self.qrel, self.index_path, self.w2v_cache, self.model_file)
         self.eval_cmd_root = "./eval/trec_eval.9.0/trec_eval -m map -m recip_rank -m bpref {}".format(self.qrel)
-        self.rbp_cmd_root = "rbp_eval {}".format(self.qrel)
+        self.rbp_cmd_root = "rbp_eval/rbp_eval {}".format(self.qrel)
```
snapbug: Needs addressing

src/main/python/e2e_experimental_settings.py
```diff
@@ -105,85 +97,83 @@ def run(self, indices):
...
+            """
+            for c in enumerate(self.combinations):
+                print(c)
+            print("--run X Y Z to run combinations number X Y Z")
+
+    if __name__ == "__main__":
+        ap = argparse.ArgumentParser(description="Lists exerimental settings and runs experiments")
+        ap.add_argument("--list", help="lists all available experimental settings combinations",
+                        action="store_true")
+        ap.add_argument("--run", help='runs experimenal setting combination NUMBER(s)',
```
snapbug: "experimenal" -> "experimental"

src/main/python/run_trecqa.py
```diff
@@ -1,223 +1,297 @@
...
-  for candidate in candidates:
-    this_candidate = " ".join(candidate[0])
+    for candidate in candidates:
+        this_candidate = Candidate(retrieved_set=set(candidate[0]), retrieved_str=" ".join(candidate[0]),
+                                   retrieved_score=candidate[1], in_dataset=set(answer[1]))
 
-    # xml file doesn't contain answers
-    if not answers:
-      scored_candidates[this_candidate] = ("empty answer", 0.0, candidate[1])
+        # xml file doesn"t contain answers
```
snapbug: Ah, the downsides of global search and replace. Be careful with it 😄

src/main/python/run_trecqa.py
```diff
@@ -1,223 +1,297 @@
...
-      prev = line
+# evaluate by TREC QA patterns
+def eval_by_pattern(qid, candidates, pattern, out):
+    seen = set([])
 
-  return questions, answers, labels
+    for i, cand in enumerate(candidates):
+        this_candidate = " ".join(cand[0])
 
+        hash_value = md5(this_candidate.encode()).hexdigest()
```
snapbug: You need to repeat the comment saying _why_ you're taking the md5 value.

src/main/python/run_trecqa.py
```diff
@@ -1,223 +1,297 @@
...
-      total_count += 1
+def create_qrel_pattern(all_sentences, pattern, qrels, qid):
+    seen = set([])
+    for i, cand in enumerate(all_sentences):
+        this_candidate = cand.lower().split("\t")[0]
+        result = re.findall(r"{}".format(pattern.lower()), this_candidate)
 
-      if not actual[qid]:
-        continue
+        # use the hash of the candidates as the docid
```
snapbug: _Why_ are we using the hash of the candidate? Is it perhaps because the sentences don't _have_ docids?

src/main/python/run_trecqa.py
```diff
@@ -1,223 +1,297 @@
...
-          doc_id = value[0][0]
-          if doc_id not in seen_docs:
-            out.write('{} Q0 {} {} {} TRECQA\n'.format(qid, doc_id, i, value[2]))
-            labels_predicted[qid].append(doc_id)
-            seen_docs.append(doc_id)
+    parser = argparse.ArgumentParser(description="Evaluate the QA system")
+    parser.add_argument("-input", help="path of a TrecQA file", required=True)
+    parser.add_argument("-output", help="path of output directory", default=".")
+    parser.add_argument("-qrel", help="path of qrel file")
+    parser.add_argument("-h0", help="number of hits", default=1000)
```
snapbug: "number of hits" is not an adequate description of what `h0` is.

src/main/python/run_trecqa.py
```diff
@@ -1,223 +1,297 @@
...
+    parser.add_argument("-model", help="[idf|sm]", default="idf")
+    parser.add_argument("-index", help="path of the index", required=True)
+    parser.add_argument("-w2v-cache", help="word embeddings cache file")
+    parser.add_argument("-qa-model-file", help="the path to the model file")
+    parser.add_argument("-eval", help="[pattern|jaccard]", default="pattern")
+    parser.add_argument("-pattern", help="path to the pattern file")
+
+    args = parser.parse_args()
+
+    if (args.model == "sm" and not args.w2v_cache and not args.qa_model_file):
```
snapbug: Remove parens

src/main/python/run_trecqa.py
```diff
@@ -1,223 +1,297 @@
...
+def evaluate_at_k(fname, eval_depth):
+    for k in eval_depth:
+        with open(fname) as f, open("{}.k{}.txt".format(fname, k), "w") as output_file:
+            for line in f:
+                det = line.strip().split()
+                rank = int(det[3])
 
-  for qid in predicted.keys():
-    run_labels = predicted.copy()
+                if rank <= int(k):
```
snapbug: `k` is already an `int`, because we mapped it prior to the function call.

src/main/python/e2e_experimental_settings.py
```diff
@@ -31,34 +31,27 @@ def __init__(self, qa_data_xml, qrel, word_embeddings_file, index_path, model_fi
...
-
-        self.combinations = []
-        for c in itertools.product(*all_settings):
-             self.combinations.append(c)
+        all_settings = [setting.get_options() for setting in self.settings.values()]
+        self.combinations = itertools.product(*all_settings)
 
     def _run_cmd(self, cmd):
         pargs = shlex.split(cmd)
         p = subprocess.Popen(pargs, stdout=subprocess.PIPE, stderr=subprocess.PIPE, \
```
snapbug: Other calls that are spread over multiple lines don't have the `\` to indicate a split line. It's not needed 😄

docs/end2end-TrecQA-eval.md
rosequ: The process takes long to complete. Hence, a background process. Should I explain this?

snapbug: I would do the opposite, say that it's a long process with a very rough time estimate (something like "~30m on a MacBook", or whatever) and remove the backgrounding. People can always open up another terminal/ssh session if they want to use the machine for other things.

src/main/python/run_trecqa.py
rosequ: Yes, that's correct. I have updated the comment.

src/main/python/run_trecqa.py
rosequ: Yes, I have made sure it works.

src/main/python/e2e_experimental_settings.py
rosequ: I'm removing `e2e_experimental_settings.py` since its purpose was to automate different e2e QA experiments. The code, however, is outdated and should rather be replaced by a simple shell script.

rosequ:
@snapbug can you have another look? I have addressed your comments.

Issue #190, opened 2017-05-23
Build Collection API for indexing tweets


lintool:
We have a nice collection API for indexing TREC doc, ClueWeb WARCs, etc., but indexing tweets is on a completely different code path - the APIs are in `io.anserini.document.twitter`. We should integrate tweets indexing into our existing framework.

Concretely, this means implementing `io.anserini.collection.TweetsCollection` that implements `Collection` and all the supporting code around it.

PR #204, opened 2017-07-18
Index microblog collection (#190)

rosequ:
- Integrated tweet indexing of into our existing framework
- Updated docs

@lintool @snapbug could you please have a look?

### DIFF
--- a/docs/experiments-microblog.md
+++ b/docs/experiments-microblog.md
@@ -1,0 +1,44 @@
+# Anserini Experiments on TREC Microblog Collection
+
+
+### Tweets2011 collection
+
+Note that the Tweets2011 collection is distributed as a list of tweet ids that you have to download yourself, so the 
+effectiveness results you'll get should be similar, but will likely not be identical to the scores reported here.
+
+Indexing the Tweets2011 collection:
+
+```
+nohup sh target/appassembler/bin/IndexCollection -collection TwitterCollection  -input \
+/path/to/Tweets2011/ -generator JsoupGenerator  -index lucene-index.Tweets2011.pos+docvectors -threads 32 \
+-storePositions -storeDocvectors -optimize > log.Tweets2011.txt &
+```
+
+Running topics from TREC 2011 (also look in `src/main/resources/topics-and-qrels/` for topics from TREC 2012):
+
+```
+sh target/appassembler/bin/SearchTweets -collection Twitter -index lucene-index.Tweets2011.pos+docvectors -bm25 \
+-hits 1000 -topics src/main/resources/topics-and-qrels/topics.microblog2011.txt -output run.mb11.bm25.txt
+```
+
+Options for the retrieval model are similar to the web runs: specify `-bm25` to use BM25, `-ql` to use query likelihood,
+ and add `-rm3` to invoke the RM3 relevance feedback model (requires docvectors index).
+
+For evaluation:
+
+```
+eval/trec_eval.9.0/trec_eval src/main/resources/topics-and-qrels/qrels.microblog2011.txt run.mb11.bm25.txt
+```
+
+You should be able to get effectiveness scores along the following lines:
+
+MAP                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
+---------------------------------------------------------------------------|--------|--------|--------|--------
+[TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.3446 | 0.3743 | 0.3631 | 0.4134
+[TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.2063 | 0.2227 | 0.2203 | 0.2540
+
+P30                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
+---------------------------------------------------------------------------|--------|--------|--------|--------
+[TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.3741 | 0.4027 | 0.3905 | 0.4340
+[TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.3237 | 0.3294 | 0.3249 | 0.3480
+
--- a/docs/experiments-microblog.md
+++ b/docs/experiments-microblog.md
@@ -18,7 +18,7 @@ __NB:__ The process is backgrounded
 Running topics from TREC 2011 (also look in `src/main/resources/topics-and-qrels/` for topics from TREC 2012):
 
 ```
-sh target/appassembler/bin/SearchTweets -collection Twitter -index lucene-index.Tweets2011.pos+docvectors -bm25 \
+sh target/appassembler/bin/SearchTweets -topicreader Twitter -index lucene-index.Tweets2011.pos+docvectors -bm25 \
 -hits 1000 -topics src/main/resources/topics-and-qrels/topics.microblog2011.txt -output run.mb11.bm25.txt
 ```
 
@@ -43,6 +43,19 @@ P@30                                                                       | BM2
 [TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.3741 | 0.4034 | 0.3905 | 0.4306
 [TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.3237 | 0.3305 | 0.3249 | 0.3475
 
+
+Effectiveness scores after retaining the retweets while indexing with `-keepRetweets` option:
+
+MAP                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
+---------------------------------------------------------------------------|--------|--------|--------|--------
+[TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.2947 | 0.3157 | 0.3054 | 0.3269
+[TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.1786 | 0.1917 | 0.1896 | 0.2133
+
+P@30                                                                       | BM25   |BM25+RM3| QL     | QL+RM3
+---------------------------------------------------------------------------|--------|--------|--------|--------
+[TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.3252 | 0.3537 | 0.3327 | 0.3476
+[TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.2938 | 0.3000 | 0.2921 | 0.3062
+
 ### Tweets2013 collection
 The Tweets 2013 collection can be indexed and searched using the commands same commands as above.
 
@@ -61,3 +74,15 @@ P@30                                                                       | BM2
 [TREC 2013 Microblog Track](http://trec.nist.gov/data/microblog2013.html)  | 0.4450 | 0.4328 |0.4533  | 0.4689
 [TREC 2014 Microblog Track](http://trec.nist.gov/data/microblog2014.html)  | 0.6085 | 0.5867 |0.6327  | 0.6533
 
+
+Effectiveness scores after retaining the retweets while indexing with `-keepRetweets` option:
+
+MAP                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
+---------------------------------------------------------------------------|--------|--------|--------|--------
+[TREC 2013 Microblog Track](http://trec.nist.gov/data/microblog2013.html)  | 0.1925 | 0.2027 |0.2027  | 0.2117
+[TREC 2014 Microblog Track](http://trec.nist.gov/data/microblog2014.html)  | 0.2961 | 0.3208 |0.3027  | 0.3205
+
+P@30                                                                       | BM25   |BM25+RM3| QL     | QL+RM3
+---------------------------------------------------------------------------|--------|--------|--------|--------
+[TREC 2013 Microblog Track](http://trec.nist.gov/data/microblog2013.html)  | 0.3722 | 0.3639 |0.3722  | 0.3722
+[TREC 2014 Microblog Track](http://trec.nist.gov/data/microblog2014.html)  | 0.5085 | 0.5097 |0.5055  | 0.4752
--- a/docs/experiments-microblog.md
+++ b/docs/experiments-microblog.md
@@ -38,26 +38,15 @@ MAP                                                                        | BM2
 [TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.3448 | 0.3719 | 0.3632 | 0.4112
 [TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.2064 | 0.2230 | 0.2203 | 0.2541
 
-P@30                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
+P@30                                                                       | BM25   |BM25+RM3| QL     | QL+RM3
 ---------------------------------------------------------------------------|--------|--------|--------|--------
 [TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.3741 | 0.4034 | 0.3905 | 0.4306
 [TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.3237 | 0.3305 | 0.3249 | 0.3475
 
 ### Tweets2013 collection
-Indexing the Tweets2013 collection:
+The Tweets 2013 collection can be indexed and searched using the commands same commands as above.
 
-```
-nohup sh target/appassembler/bin/IndexCollection -collection TwitterCollection  -input \
-/path/to/Tweets2013/ -generator JsoupGenerator  -index lucene-index.Tweets2013.pos+docvectors -threads 32 \
--storePositions -storeDocvectors -optimize > log.Tweets2013.txt &
-```
-
-Running topics from TREC 2013 (also look in `src/main/resources/topics-and-qrels/` for topics from TREC 2014):
-
-```
-sh target/appassembler/bin/SearchTweets -collection Twitter -index lucene-index.Tweets2013.pos+docvectors -bm25 \
--hits 1000 -topics src/main/resources/topics-and-qrels/topics.microblog2011.txt -output run.mb13.bm25.txt
-```
+Running topics from TREC 2013 and from TREC 2014 (look in `src/main/resources/topics-and-qrels/` ):
 
 Retrieval model options are the same as above, as is running `trec_eval`. You should be able to get effectiveness scores 
 along the following lines:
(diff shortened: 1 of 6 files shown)


docs/experiments-microblog.md
```diff
@@ -0,0 +1,44 @@
+# Anserini Experiments on TREC Microblog Collection
+
+
+### Tweets2011 collection
+
+Note that the Tweets2011 collection is distributed as a list of tweet ids that you have to download yourself, so the 
+effectiveness results you'll get should be similar, but will likely not be identical to the scores reported here.
```
snapbug: Because the tweets might have been deleted?

To be pedantic, you mightn't get the same scores regardless. It depends on docid assignment, and how you break ties in ranking. I'm not entirely sure how lucene does either of these, so feel free to ignore/comment back.

docs/experiments-microblog.md
```diff
@@ -0,0 +1,44 @@
...
+
+For evaluation:
+
+```
+eval/trec_eval.9.0/trec_eval src/main/resources/topics-and-qrels/qrels.microblog2011.txt run.mb11.bm25.txt
+```
+
+You should be able to get effectiveness scores along the following lines:
+
+MAP                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
```
snapbug: You know my thoughts on AP/MAP 😀

docs/experiments-microblog.md
```diff
@@ -0,0 +1,44 @@
...
+---------------------------------------------------------------------------|--------|--------|--------|--------
+[TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.3448 | 0.3719 | 0.3632 | 0.4112
+[TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.2064 | 0.2230 | 0.2203 | 0.2541
+
+P30                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
```
snapbug: And to add to the pedantry, it's P@30, rather than P30.

src/main/java/io/anserini/collection/TwitterCollection.java
```diff
@@ -0,0 +1,96 @@
...
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Class representing an instance of a TREC Microblog collection.
```
snapbug: TREC Microblog or Twitter? I know they're essentially the same, but when the class is called `TwitterCollection` it's probably best that the comment matches that.

src/main/java/io/anserini/collection/TwitterCollection.java
```diff
@@ -0,0 +1,96 @@
...
+        InputStream stream = new GZIPInputStream(
+                Files.newInputStream(path, StandardOpenOption.READ), BUFFER_SIZE);
+        bufferedReader = new BufferedReader(new InputStreamReader(stream, StandardCharsets.UTF_8));
+      } else {
+        // in case user had already uncompressed the folder
+        bufferedReader = new BufferedReader(new FileReader(fileName));
+      }
+    }
+
+
```
snapbug: Remove this blank line

src/main/java/io/anserini/collection/TwitterCollection.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    public boolean hasNext() {
+      return !atEOF;
+    }
+
+    @Override
+    public TwitterDocument next() {
+
```
snapbug: Remove this blank line

src/main/java/io/anserini/collection/TwitterCollection.java
```diff
@@ -0,0 +1,96 @@
...
+      TwitterDocument doc = new TwitterDocument();
+      String raw = null;
+
+      try {
+        raw = bufferedReader.readLine();
+      } catch (IOException e) {
+        e.printStackTrace();
+      }
+
+      // check to see if we've reached end of file.
```
snapbug: Is this really a helpful comment? Comments should be about why things are being done, and how in complicated cases. I'd say these few lines of code are relatively obvious. 🤷‍♂️

src/main/java/io/anserini/collection/TwitterCollection.java
```diff
@@ -0,0 +1,96 @@
...
+      try {
+        doc = (TwitterDocument) doc.readNextRecord(raw);
+
+        if (doc == null) {
+          return null;
+        }
+      } catch (IOException e) {
+        e.printStackTrace();
+      }
+      return  doc;
```
snapbug: I could be wrong as it's been some time since I've written java, but if the prior try/catch block throws an exception, then what gets returned here is `new TwitterDocument()` rather than `null`, is this intended?

src/main/java/io/anserini/collection/TwitterCollection.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  @Override
+  public List<Path> getFileSegmentPaths() {
+    Set<String> allowedFileSuffix = new HashSet<>(Arrays.asList(".gz"));
+
+    return discover(path, EMPTY_SET, EMPTY_SET, EMPTY_SET,
+            allowedFileSuffix, EMPTY_SET);
```
snapbug: I think this line has been wrapped prematurely for an 80 character width. Also check the style being used for alignment of wrapped function invocations like this. (I don't know which is being followed so can't make more direct comments)

src/main/java/io/anserini/collection/TwitterCollection.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @Override
+  public FileSegment createFileSegment(Path p) throws IOException {
+    return new FileSegment(p);
+  }
+
```
snapbug: Remove this blank line?

src/main/java/io/anserini/document/TwitterDocument.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * A Twitter document (status).
+ */
+public class TwitterDocument implements SourceDocument {
+  private static final Logger LOG = LogManager.getLogger(io.anserini.document.twitter.Status.class);
+
+  private String id;
+  private String contents;
+
+  private static final JsonParser JSON_PARSER = new JsonParser();
```
snapbug: I would keep the `private static final` declarations together rather than split

src/main/java/io/anserini/document/TwitterDocument.java
```diff
@@ -0,0 +1,63 @@
...
+    if (obj.get("text") == null) {
+      return null;
+    }
+
+    // Do not index retweets
+    try {
+      long retweetStatusID = obj.getAsJsonObject("retweeted_status").get("id").getAsLong();
+      return null;
+    } catch (Exception e) {
+      // Do nothing if the retweet doesn't exist
```
snapbug: This comment is misleading, it threw me. Update it to match what's _actually_ going on. Do nothing if there is no retweeted_status key because it's then not a retweet, it's not that the retweet doesn't exist.

src/main/java/io/anserini/search/SearchTweets.java
```diff
@@ -96,16 +97,17 @@ public static void main(String[] args) throws Exception {
     }
 
     RerankerCascade cascade = new RerankerCascade();
+    EnglishAnalyzer englishAnalyzer = new EnglishAnalyzer();
     if (searchArgs.rm3) {
-      cascade.add(new Rm3Reranker(IndexTweets.ANALYZER, StatusField.TEXT.name, "src/main/resources/io/anserini/rerank/rm3/rm3-stoplist.twitter.txt"));
+      cascade.add(new Rm3Reranker(englishAnalyzer, FIELD_BODY, "src/main/resources/io/anserini/rerank/rm3/rm3-stoplist.twitter.txt"));
```
snapbug: Check the wrapping on this line, and others in this file.

src/main/java/io/anserini/search/SearchTweets.java
```diff
@@ -130,24 +132,24 @@ public static void main(String[] args) throws Exception {
     for ( MicroblogTopic topic : topics ) {
       long curQueryTime = System.nanoTime();
 
-      Query filter = LongPoint.newRangeQuery(StatusField.ID.name, 0L, topic.getQueryTweetTime());
-      Query query = AnalyzerUtils.buildBagOfWordsQuery(StatusField.TEXT.name, IndexTweets.ANALYZER, topic.getQuery());
+      Query filter = TermRangeQuery.newStringRange(FIELD_ID, "0", String.valueOf(topic.getQueryTweetTime()), true, true);
```
snapbug: I'd add a comment about why we're doing a range query, it's non-obvious (to me at least)

src/main/java/io/anserini/search/SearchTweets.java
```diff
@@ -130,24 +132,24 @@ public static void main(String[] args) throws Exception {
     for ( MicroblogTopic topic : topics ) {
       long curQueryTime = System.nanoTime();
```
snapbug: I know you didn't change this line, but given that you're converting to ms later on, would `system.currentTimeMillis()` not be the better call to make? Also, `curQueryTime` isn't a super amazing name for this. It's really `startQueryTime` or even maybe just `start`?

src/main/java/io/anserini/search/SearchTweets.java
```diff
@@ -130,24 +132,24 @@ public static void main(String[] args) throws Exception {
...
+         queryTokens, FIELD_BODY, filter);
       ScoredDocuments docs = cascade.run(ScoredDocuments.fromTopDocs(rs, searcher), context);
 
       for (int i=0; i<docs.documents.length; i++) {
         String qid = topic.getId().replaceFirst("^MB0*", "");
         out.println(String.format("%s Q0 %s %d %f %s", qid,
-            docs.documents[i].getField(StatusField.ID.name).numericValue(), (i+1), docs.scores[i], searchArgs.runtag));
+            docs.documents[i].getField(FIELD_ID).stringValue(), (i+1), docs.scores[i], searchArgs.runtag));
       }
       long qtime = (System.nanoTime()-curQueryTime)/1000000;
```
snapbug: Again, I know you didn't change this line in this commit, but putting the end timer here counts the time to output the results. Typically this isn't counted when measuring query time.

docs/experiments-microblog.md
rosequ: @snapbug since the code base is the same, the only parameter here is the input collection. What I mean to say is that numbers might very different depending on the collection, deletes etc. The numbers reported here are on the original collection and not after the tweets are being deleted.

docs/experiments-microblog.md
```diff
@@ -0,0 +1,44 @@
...
+Note that the Tweets2011 collection is distributed as a list of tweet ids that you have to download yourself, so the 
+effectiveness results you'll get should be similar, but will likely not be identical to the scores reported here.
+
+Indexing the Tweets2011 collection:
+
+```
+nohup sh target/appassembler/bin/IndexCollection -collection TwitterCollection  -input \
+/path/to/Tweets2011/ -generator JsoupGenerator  -index lucene-index.Tweets2011.pos+docvectors -threads 32 \
+-storePositions -storeDocvectors -optimize > log.Tweets2011.txt &
+```
```
rosequ: I remember a similar opinion from the previous CR too :P The documentation throughout Anserini follows a consistent with this notation, lest an unsuspecting user just copy-paste the command on his server. I can change it, if you insist!

docs/experiments-microblog.md
rosequ: Also, about M-nDCG, M-P@30 etc :)

src/main/java/io/anserini/collection/TwitterCollection.java
rosequ: Good catch!

src/main/java/io/anserini/collection/TwitterCollection.java
rosequ: We can start a war over this one: https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file

docs/experiments-microblog.md
snapbug: I'm in two minds, yes protecting against copy+paste is a good thing, but then that could also lead to surprises. For example if I copy+paste this, then it gets backgrounded and I get a new prompt, there's no output because it's all redirected (`nohup` redirects regardless of whether you add the redirection yourself). I _could_ then think that nothing happened and that it's broken.

I'm also against protecting against copy+paste because it's bad practice as a developer to blindly do it without looking at what's being performed.

The simplest way to address this is to just say that it's backgrounded and that you can come back to it. Of course if it's consistent with all the other Anserini docs, then perhaps best leave it as a personal annoyance more than an issue that needs addressing, after all consistency is perhaps the most important thing.

docs/experiments-microblog.md
snapbug: Except they would be MnDCG and MP@30, to match AP/MAP 😛

src/main/java/io/anserini/collection/TwitterCollection.java
snapbug: I did't see that it was the end of file, in that case keeping it is fine.

docs/experiments-microblog.md
```diff
@@ -37,8 +38,37 @@ MAP                                                                        | BM2
...
+nohup sh target/appassembler/bin/IndexCollection -collection TwitterCollection  -input \
+/path/to/Tweets2013/ -generator JsoupGenerator  -index lucene-index.Tweets2013.pos+docvectors -threads 32 \
+-storePositions -storeDocvectors -optimize > log.Tweets2013.txt &
+```
+
+Running topics from TREC 2013 (also look in `src/main/resources/topics-and-qrels/` for topics from TREC 2014):
+
+```
+sh target/appassembler/bin/SearchTweets -collection Twitter -index lucene-index.Tweets2013.pos+docvectors -bm25 \
+-hits 1000 -topics src/main/resources/topics-and-qrels/topics.microblog2011.txt -output run.mb13.bm25.txt
```
snapbug: The 2011 topics on the 2013 collection? Also I'd suggest that the indexing command differences between 2011/2013 are so minor that repeating the commands isn't necessary (much like you did with retrieval and `trec_eval` commands).

src/main/java/io/anserini/search/SearchTweets.java
```diff
@@ -145,13 +146,14 @@ public static void main(String[] args) throws Exception {
       RerankerContext context = new RerankerContext(searcher, query, topic.getId(), topic.getQuery(),
          queryTokens, FIELD_BODY, filter);
       ScoredDocuments docs = cascade.run(ScoredDocuments.fromTopDocs(rs, searcher), context);
+      long qtime = (System.currentTimeMillis() - startQueryTime);
```
snapbug: `qtime` isn't great, `elapsed_time` or `query_time` would be better, even if it is relatively obvious what's being calculated. Also, check `startQueryTime` matches the start time you mean 😄  (my last comment might have been a bit misleading because I missed that there were two timers going on)

And because I can't comment above -- `long curQueryTime = System.nanoTime()` should be switched to `currentTimeMillis` to match.

src/main/java/io/anserini/document/TwitterDocument.java
snapbug: This is just a personal preference, so don't feel obligated to change, but I'd put the `static` set first.

docs/experiments-microblog.md
```diff
@@ -37,8 +38,37 @@ MAP                                                                        | BM2
 [TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.3448 | 0.3719 | 0.3632 | 0.4112
 [TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.2064 | 0.2230 | 0.2203 | 0.2541
 
-P30                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
+P@30                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
```
snapbug: Super pedantic, adding the `@` has shifted the alignment of the later columns. I know that the alignment only affects the source representation and not the output, but you've gone to the effort 😄

docs/experiments-microblog.md
```diff
@@ -37,8 +38,37 @@ MAP                                                                        | BM2
...
+
+Retrieval model options are the same as above, as is running `trec_eval`. You should be able to get effectiveness scores 
+along the following lines:
+
+MAP                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
+---------------------------------------------------------------------------|--------|--------|--------|--------
+[TREC 2013 Microblog Track](http://trec.nist.gov/data/microblog2013.html)  | 0.2438 | 0.2563 |0.2632  | 0.2890
+[TREC 2014 Microblog Track](http://trec.nist.gov/data/microblog2014.html)  | 0.3868 | 0.3990 |0.4122  | 0.4706
+
+P@30                                                                        | BM25   |BM25+RM3| QL     | QL+RM3
```
snapbug: As above w.r.t. `|` alignment.

docs/experiments-microblog.md
```diff
@@ -13,6 +13,7 @@ nohup sh target/appassembler/bin/IndexCollection -collection TwitterCollection
 /path/to/Tweets2011/ -generator JsoupGenerator  -index lucene-index.Tweets2011.pos+docvectors -threads 32 \
 -storePositions -storeDocvectors -optimize > log.Tweets2011.txt &
 ```
+__NB:__ The process is backgrounded
```
snapbug: I know this seems like I'm hammering on a (rather non-crucial) point, but if it's backgrounded how do I tell it's finished? Will `log.Tweets2011.txt` have a line at the end saying it's done? Or do I have to check with ps/top/htop?

src/main/java/io/anserini/search/SearchTweets.java
```diff
@@ -58,7 +58,7 @@
   private SearchTweets() {}
 
   public static void main(String[] args) throws Exception {
-    long curTime = System.nanoTime();
+    long startQueryTime = System.currentTimeMillis();
```
snapbug: In light of a later comment, I missed that there were two timers. This timer isn't really measuring query time, so is it far to call it `startQueryTime`?

src/main/java/io/anserini/search/SearchTweets.java
rosequ: Sorry, missed this one!

rosequ:
@lintool can we merge this?

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -78,6 +79,9 @@
     @Option(name = "-keepStopwords", usage = "boolean switch to keep stopwords")
     public boolean keepStopwords = false;
 
+    @Option(name = "-keepRetweets", usage = "Boolean switch to keep retweets while indexing")
```
snapbug: And here we have an interesting debate, consistency vs. correctness. Consistency would say to lowercase `Boolean` in the help message, correctness would say to change all the others. In this case because it's likely lowercased in other files across the project, consistency probably wins.

src/main/java/io/anserini/document/TwitterDocument.java
```diff
@@ -32,11 +32,13 @@ public SourceDocument readNextRecord(String json) throws IOException {
     }
 
     // Do not index retweets
```
snapbug: And now this comment is a bit redundant from the variable name 😄

rosequ:
@lintool can we merge this?

PR #207, opened 2017-11-25
Updated Answer Retriever

MeowFei:


src/main/java/io/anserini/qa/RetrieveSentences.java
```diff
@@ -131,7 +132,7 @@ public RetrieveSentences(RetrieveSentences.Args args) throws Exception {
   }
 
   public void getRankedPassages(Args args) throws Exception {
-    Map<String, Float> scoredDocs  = retrieveDocuments(args);
+    Map<String, Float> scoredDocs  = retrieveDocuments(args.query, 100);
```
rosequ: I think it looks good now. Just change 100 to `args.hits`.

PR #212, opened 2018-03-30
Refactor Collection class a bit; Add Document tests

Peilin-Yang:
This PR mainly does two things:

1. Refactor `Collection` a bit so that adding new Collection class requires less effort. The primary change is to make a protected variable `dType` which of the type `D` so that we can keep the logic of `next()` in the abstract `Collection` class. The exceptions are CW09, CW12 and WikiAriticle which do not take `BufferReader` as the source of reading the next record. I manually test all types of collections and it works as expected.
```
public abstract class Collection<D extends SourceDocument> {
public abstract class FileSegment implements Iterator<D>, Closeable {
    protected Path path;
    protected BufferedReader bufferedReader;
    protected boolean atEOF = false;
    protected final int BUFFER_SIZE = 1 << 16; // 64K
    protected D dType;

    @Override
    public boolean hasNext() {
      return !atEOF;
    }

    @Override
    public D next() {
      SourceDocument d;
      try {
        d = dType.readNextRecord(bufferedReader);
        if (d == null) {
          atEOF = true;
          d = null;
        }
      } catch (IOException e) {
        d = null;
      }
      return (D)d;
    }

    @Override
    public void remove() {
      throw new UnsupportedOperationException();
    }

    @Override
    public void close() throws IOException {
      atEOF = false;
      if (bufferedReader != null) {
        bufferedReader.close();
      }
    }
  }
```
2. Add tests for most Document classes.

### DIFF
--- a/src/main/java/io/anserini/collection/TwitterCollection.java
+++ b/src/main/java/io/anserini/collection/TwitterCollection.java
@@ -1,5 +1,6 @@
 package io.anserini.collection;
 
+import io.anserini.document.SourceDocument;
 import io.anserini.document.TwitterDocument;
 
 import java.io.*;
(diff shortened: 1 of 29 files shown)


src/main/java/io/anserini/collection/TwitterCollection.java
```diff
@@ -13,76 +14,39 @@
 import java.util.Set;
 import java.util.zip.GZIPInputStream;
```
rosequ: delete empty line here

PR #215, opened 2018-04-14
Tweets indexing/searching/tests refactor

Peilin-Yang:
This PR mainly aims to refactor the old tweets indexing and searching code and make them compatible with the rest of Anserini, i.e. the indexing framework of using XYZCollection, XYZDocument and XYZGenerator; the searching framework of XYZTopicReader and SearchCollection.
This is a big PR and can be roughly split into different parts:

1. **The indexing part**
I've removed the old files under `src/main/java/io/anserini/document/*` and `src/main/java/io/anserini/index/IndexTweets.java`. Instead, `src/main/java/io/anserini/document/TweetDocument.java` and `src/main/java/io/anserini/index/generator/TweetGenerator.java` are created as part of the more general ingestion adapter framework as the rest of Anserini. 
During the refactor, I found that the parsing of tweets could also throw errors. For example, the `"text"` field is absent from the raw Tweet Json. Thus I changed the return type of `SourceDocument.readNextRecord()` to `SourceDocumentResultWrapper<T extends SourceDocument>` which takes the advantage of `Optional` in Java 8. `SourceDocumentResultWrapper<T>` contains two members: `Optional<T> document` and `FailureReason reason` where `FailureReason` is an enum lists possible failure reasons. By doing in this way, we are able to capture the failure when parsing the raw data file whereby we previously only capture during the generator step. This change affects _all_ existing subclasses of SourceDocument but I've fixed all of them. The other main change is to add a new `TweetGenerator` where we can control how to parse the tweet fields and how to index them. I've removed all `StoredField` but only kept one `StoredField` which, if user specifies `-storeRawDocs` when indexing, stores the raw whole Json string. In `TweetGenerator` I also introduced `TweetText` from official tweet github repo which we can use to extract the URL links. User can choose not to include URLs when indexing by specifying `-tweet.removeUrls` flag. Other added flags include `-tweet.keepRetweets`, `-tweet.stemming`, `-tweet.maxId` and `-tweet.deletedIdsFile`. Please refer to `src/main/java/io/anserini/index/IndexCollection.java` for more details.
2. **The searching part**
Searching has relatively small changes. Firstly, `SearchWebCollection` is renamed to `SearchCollection`. Now all searches are fulfilled in `SearchCollection`. For tweet search specific functionalities I introduce a flag `-searchtweets`.  
For topic readers, I've changed the return type from `SortedMap<Integer, String>` to `SortedMap<Integer, Map<String, String>>` in order to meet the microblog topic formats: for microblog topics there are "query" and "querytweettime" fields. Interestingly, TRECTopicReader also benefits from this change as we can store "title", "description" and "narrative" queries at the same time. I've added a flag `-topicfield` to indicate which topic field should be used as query string, the default value is "title". Based the aforementioned details I've refactored `MicroblogTopicReader`.
Here I need some inputs from others: I followed the old `SearchTweets` logic and build the query as follows:
```
if (searchtweets) {
    analyzer = new TweetAnalyzer();
    Query filter = TermRangeQuery.newStringRange(FIELD_ID, "0", queryTweetTime, true, true);
    BooleanQuery.Builder builder = new BooleanQuery.Builder();
    builder.add(filter, BooleanClause.Occur.FILTER);
    builder.add(query, BooleanClause.Occur.MUST);
    Query q = builder.build();
} else { ... }
```
There are two things I am not sure:
(1) Should we be using `TweetAnalyzer`?
(2) Is the query builder the correct way to do it? Specifically, should we be filtering tweets to only return tweets whose id are within [0, queryTweetTime]? Also, should we be using `BooleanClause.Occur.MUST` instead of `BooleanClause.Occur.SHOULD`?
Once we confirm these I can test it against some datasets and verify the results to finalize this PR.
3. **Adding End-to-End indexing, searching, evaluation tests**
Currently, we do not have and end-to-end indexing, searching and evaluation pipeline for testing. 
So I've added the e2e tests for this regard. The e2e tests directly call the main function (not the actual main function but the essential part of it) of `IndexCollection`, `SearchCollection` and `Eval`. 
The e2e tests look at the stats such as total terms in the index to verify the indexes that are created; it also ranks the docs for the sample topics and directly verify the evaluation metrics like MAP. `Eval` was merged long way ago and I think it is time to fully adopt it for streamlining the whole process.
For now, I only created two e2e tests for Trec and Tweet. For the rest of the collections/topics I plan to add them in another PR. Please refer to `src/test/java/io/anserini/end2end/*` for more info. 
_I think we should enable CI or something similar to automatically test the new PRs not breaking the current e2e tests._
4. **Documentation changes**
Mainly change `SearchWebCollection` to `SearchCollection`. 
Also added the `experiment-tweet.md` 

The changes unavoidably introduced some changes to related code. For example, the TweetFeatureExtractor, RTS and NRTS. 
For now, I do not have accurate description on how they are being affected.


Peilin-Yang:
@lintool can you take a look at question 2 above?
Also, I would like to test the new indexing/searching of tweets on MB2011-MB2014 but currently I do not have these datasets at hand. Can you help?
What about RTS201X?

Peilin-Yang:
[UPDATE]
The PR was tested against MB11 and MB13. 
For indexing, `tweet.stemming` and `tweet.keepUrls` were enabled. 
The ranking function used was BM25 with default parameters and the evaluation metric is MAP. The results:

| Collection | Before | After |
|---|---|---|
| MB11 | 0.3448 | 0.3352 |
| MB13 | 0.2438 | 0.2271 |

It seems that the refactor degraded the performance a bit.
I did a quick check but did not find anything suspicious that may cause the difference.
However, I am not sure if the indexing options or the ranking options are exactly the same for the two runs. cc @lintool

Peilin-Yang:
Some updates:
1. For `Analyzer`, Twitter's official analyzer penguin hasn't updated for 4 years... So I would rather stick with the current (simple) TweetTextAnalyzer
2. For ranking the tweets in Evaluation stage, I've added two boolean flags to it - `longdocids` and `docidsdesc` which users can specify and break the score ties by ordering the tweetId in long type with descending order. But there is no performance change after the changes had been applied.

lintool:
From official API deployed for the TREC Microblog Tracks:

MAP       | raw    | baseline
----------|--------|---------
TREC 2011 | 0.3050 | 0.3576
TREC 2012 | 0.1751 | 0.2091
TREC 2013 | 0.2044 | 0.2532
TREC 2014 | 0.3090 | 0.3924

P30       | raw    | baseline
----------|--------|---------
TREC 2011 | 0.3483 | 0.4000
TREC 2012 | 0.2831 | 0.3311
TREC 2013 | 0.3761 | 0.4450
TREC 2014 | 0.5145 | 0.6182

These values are taken from the official [twitter-tools](https://github.com/lintool/twitter-tools) repo.

Both use QL for reranking. They also *explicitly* break scoring ties chronologically (earliest first) by adding epsilon to the score. In addition:

+ Code for [raw](https://github.com/lintool/twitter-tools/blob/master/twitter-tools-core/src/main/java/cc/twittertools/search/api/RunQueriesThrift.java)
+ Code for [baseline](https://github.com/lintool/twitter-tools/blob/master/twitter-tools-core/src/main/java/cc/twittertools/search/api/RunQueriesBaselineThrift.java), which has retweets removed.

Also, IIRC, BM25 isn't particularly effective for ranking tweets, primarily because BM25's length normalization is ineffective.

@Peilin-Yang Can you redo before/after comparisons specifically wrt to above figures, using QL (instead of BM25)?

lintool:
@Peilin-Yang Some discussion about `SourceDocumentResultWrapper` -

Based on my reading of the code, the most substantive change is from:

> `public abstract class FileSegment implements Iterator<D>, Closeable {`

to

> `public abstract class FileSegment implements Iterator<SourceDocumentResultWrapper<T>>, Closeable {`

Where you can more more explicitly track the reason:

```
public enum FailureReason {
    EOF,
    EMPTY,
    ParsingError,
    IOError
};
```

It seems to me that `EOF` is not necessary, since the `Iterator` will naturally stop at the end of the iteration. I don't think `EMPTY` is necessary either, since we can just return an actual empty `SourceDocument`.

For `ParsingError` and `IOError`, why not just throw exceptions? I'm not sure what we gain explain by this extra wrapping layer?

Victor0118:
Hi I agree with @lintool 's comment. Exception is exactly designed to explain the FailureReason and deal with that. We should use it as long as it meets our requirement. What's your opinion @Peilin-Yang ?

Peilin-Yang:
@lintool Revert `SourceDocumentResultWrapper`.

For Tweets indexing test, do you remember which Lucene version did you use?
Otherwise, I will be using the current master to do the comparison.
sg?

lintool:
@Peilin-Yang using current master for comparison - sg.

Peilin-Yang:
I think I have fix the problem of searching/reranking the tweets.
The reproduced results and comparisons are the following:

**Previous**:

MAP                                                                        | BM25   |BM25+RM3| QL     | QL+RM3 |
---------------------------------------------------------------------------|-------:|-------:|-------:|-------:|
[TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.3388 | 0.3514 | 0.3631 | 0.4024
[TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.1923 | 0.2093 | 0.2084 | 0.2377
[TREC 2013 Microblog Track](http://trec.nist.gov/data/microblog2013.html)  | 0.2319 | 0.2409 | 0.2536 | 0.2747
[TREC 2014 Microblog Track](http://trec.nist.gov/data/microblog2014.html)  | 0.3791 | 0.3971 | 0.3976 | 0.4393

**This PR**:

MAP                                                                        | BM25   |BM25+RM3| QL     | QL+RM3 |
---------------------------------------------------------------------------|-------:|-------:|-------:|-------:|
[TREC 2011 Microblog Track](http://trec.nist.gov/data/microblog2011.html)  | 0.3352 | 0.3436 | 0.3614 | 0.4013
[TREC 2012 Microblog Track](http://trec.nist.gov/data/microblog2012.html)  | 0.1910 | 0.2056 | 0.2098 | 0.2405
[TREC 2013 Microblog Track](http://trec.nist.gov/data/microblog2013.html)  | 0.2306 | 0.2363 | 0.2601 | 0.2836
[TREC 2014 Microblog Track](http://trec.nist.gov/data/microblog2014.html)  | 0.3835 | 0.4008 | 0.4185 | 0.4743

Also, a test of RM3 on Gov2 makes sure that this PR won't break TREC searches.
**_Note_**:  I keep the analyzer as TweetAnalyzer as it was used previously.

PR #221, opened 2018-06-05
Add CAR18 collection

Victor0118:
Data link: http://trec-car.cs.unh.edu/datareleases/
Read the paragraph *.cbor file using the tool provided in https://github.com/TREMA-UNH/trec-car-tools/tree/master

Results with default parameters:

  Evaluation Metric  | test_200 | benchmark_train | 
---------|-------------|-------------|
Rprec | 0.1237  | 0.1244 |
MAP | 0.1611 | 0.1639 | 
MRR | 0.2228 | 0.2255 | 

@lintool @Peilin-Yang Could you review this PR? #222


src/main/java/io/anserini/collection/CAR18Collection.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * Class representing an instance of a Wikipedia collection. Note that Wikipedia dumps often come
```
Peilin-Yang: Forget to change the annotation?

Victor0118: Thanks for the reminding!

Victor0118:
You are right. There is no change for the document format and I have renamed the files.

src/main/java/io/anserini/document/CARDocument.java
```diff
@@ -0,0 +1,57 @@
...
+ * Reference: http://trec-car.cs.unh.edu/datareleases/
+ */
+public class CARDocument implements SourceDocument {
+  private final String paraID;
+  private final String paragraph;
+
+  public CARDocument(String paraID, String paragraph) {
+    this.paraID = paraID;
+    this.paragraph = paragraph;
+
```
Peilin-Yang: remove the empty line

Victor0118:
It is in CBOR format. CBOR is similar to JSON, but it is a binary format that compresses better and avoids text file encoding issues. Reference: https://github.com/TREMA-UNH/trec-car-tools/tree/master

Thanks for your reminding! I will do that.

lintool:
Hi @Victor0118 per @Peilin-Yang classes should be named `CarCollection`, `CarDocument`, etc. as opposed to `CARCollection`, etc.

src/main/java/io/anserini/search/query/CARTitleTopicReader.java
```diff
@@ -0,0 +1,59 @@
+package io.anserini.search.query;
```
Peilin-Yang: What is the difference between `CarTitleTopicReader.java` and `CarTopicReader.java`?
I think we should have only one of them?

Victor0118: CAR provides two files for query: title and topic. They are in different formats so I write different topic reader. Might combine them in one Reader class?

Peilin-Yang: I think so. 
Remember topic is represented in `SortedMap<String, Map<String, String>>`? 
The key of the inner Map is the _tag_ of the value (actual query) of the inner Map.
For example, ```{"1": {"title": "university of waterloo", "description": "where is university of waterloo", ...}, ...}```

docs/experiments-car18.md
```diff
@@ -0,0 +1,37 @@
+# Anserini Experiments on TREC Core
```
Peilin-Yang: Change this

Peilin-Yang: Also, do not forget to update the names in the doc such as `CarCollection` after you finish everything.

Victor0118:
Cannot provide hard-coded text for testing since the CarCollection loads the `.cbor` binary file directly. Other issues have been fixed. 
Please review when you are available. @lintool @Peilin-Yang

docs/experiments-car18.md
```diff
@@ -0,0 +1,37 @@
+# Anserini Experiments on TREC Complext Answer Retrieval
+
+Indexing:
+
+```
+nohup sh target/appassembler/bin/IndexCollection -collection CarCollection -input /path/to/car_corpus/ \
+  -generator LuceneDocumentGenerator -index lucene-index-all.car18 -threads 80 -storePositions \
+  -optimize > log_all.car18
```
Peilin-Yang: use `>& log_all.car18 &` to make the command consistent with other collections?

src/main/java/io/anserini/document/CarDocument.java
```diff
@@ -0,0 +1,62 @@
...
+public class CarDocument implements SourceDocument {
+  private final String paraID;
+  private final String paragraph;
+
+  public CarDocument(String paraID, String paragraph) {
+    this.paraID = paraID;
+    this.paragraph = paragraph;
+  }
+
+  /**
```
Peilin-Yang: The javadoc here does not tell anything.
Instead, you can explain here why this function always returns `null` which other document adaptors usually do not, e.g. because the documents are processed somewhere else.

Victor0118: Good suggestion.  Thanks.

src/main/java/io/anserini/document/CarDocument.java
```diff
@@ -0,0 +1,63 @@
...
+   *
+   * @param bRdr
+   * @return
+   * @throws IOException
```
Peilin-Yang: I actually meant the javadoc of the `readNextRecord` function....There are two ways to do it:

1. To put comments in the front of the class (which you currently do) then I think it is probably better to just remove the javadoc of the `readNextRecord` function
2. Or you can move the comments in the javadoc of the `readNextRecord` function.

Personally, I prefer the second option but it is up to you.

PR #235, opened 2018-06-13
pom cleanup

lintool:
I went through and removed all the dependencies that weren't needed to clean up the repo.


Peilin-Yang:
Is there any plan to change the version from `snapshot` to `release`?
We can leverage github's `release` to mark milestones....
It will be useful for, for example, regression test.

lintool:
Yes, that's my plan - once we get the repo to a reasonable shape, I'd like to cut an official first release.

Issue #239, opened 2018-06-14
Open discussion thread: interface between Castor and Anserini


lintool:
What's the interface between Castor and Anserini? This will dictate the shape of the [pyjnius](https://github.com/kivy/pyjnius) we will build.

As a starting proposal: Anserini does document retrieval and that's it. It ranks documents given a query and hands over the text of the document as a text string. Everything else is done on the Python end. This includes sentence splitting, PTB tokenization, looking up word vectors, etc.

Thoughts?

Victor0118:
I agree with the proposal. I have tried Pyjnius and it is really friendly and easy to use. There might be some issues with the special character encoding but at least currently I can handle all issues I got.

tuzhucheng:
Hi @mengf821, do you have any suggestions?

lintool:
@Victor0118 Have you already gotten Pyjnius to work with Anserini? Can you describe exactly what you did, e.g., code snippets?

Victor0118:
This code snippet searches the index by docID and returns the doc.
If you run this in the TUNA with python2, you can a document that some special characters become Mojibake.
If you run it with python3, you will get an exception and nothing is returned.
```
#import sys
#print(sys.getdefaultencoding())

import jnius_config
jnius_config.set_classpath("/home/w85yang/Anserini/target/anserini-0.0.1-SNAPSHOT.jar")

from jnius import autoclass
JString = autoclass('java.lang.String')

index_test = autoclass('io.anserini.index.IndexUtils')
indexes = index_test(JString('/home/w85yang/Anserini/lucene-index-all.car18'))

print(indexes.getTransformedDocument(JString('7250e1b901bb59853deb38a452f9009999e790ae')))

#print(indexes.getTransformedDocument(JString('77ef93823011e891e572ca9d34a6e251f53c052a')))
```

Peilin-Yang:
I think this is the right way to go. Python libs dominate the DL field and we probably shouldn't use any Java libs for this matter.
FWIW, https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/index/IndexUtils.java has pretty much what you want - dump the raw document, dump the docVector, etc. 

For the interface, it can be done by adding a wrapper on top of https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/index/IndexUtils.java or if this is a very demanding feature request a server/client might be even better.

lintool:
[...]

PR #241, opened 2018-06-14
Removed io.anserini.{embeddings, qa, and py4j}

lintool:
Per #239


lintool:
Primary goal is to remove embeddings package, but qa and py4j depend  on it also. Both are also not needed anymore since we've decided to move pyjnius and perform all reranking/scoring on the Python end.

lintool:
@Peilin-Yang this is code by @rosequ so I explicitly want a sign off from him. Let's not merge yet.

rosequ:
I have no objections, this can be removed.

PR #244, opened 2018-06-15
SimpleSearcher Prototype

lintool:
I wrote a `SimpleSearcher` prototype so something like the following with Pyjnius works:

```
import jnius_config
jnius_config.set_classpath("target/anserini-0.0.1-SNAPSHOT-fatjar.jar")

from jnius import autoclass
JString = autoclass('java.lang.String')
JSearcher = autoclass('io.anserini.search.SimpleSearcher')

searcher = JSearcher(JString('lucene-index.robust04.pos+docvectors+rawdocs'))
hits = searcher.search(JString('hubble space telescope'))

# the docid of the 1st hit
hits[0].docid

# the internal Lucene docid of the 1st hit
hits[0].ldocid

# the score of the 1st hit
hits[0].score

# the full document of the 1st hit
hits[0].content
```

Not ready for merging, but comments would be appreciated.

@Victor0118 @tuzhucheng @rosequ I assume something like this will be useful in connecting to Castor?


src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -0,0 +1,99 @@
...
+
+    TopDocs rs = searcher.search(query, k);
+    ScoreDoc[] hits = rs.scoreDocs;
+
+    Result[] results = new Result[hits.length];
+    for (int i = 0; i < hits.length; i++) {
+      Document doc = searcher.doc(hits[i].doc);
+      String docid = doc.getField(LuceneDocumentGenerator.FIELD_ID).stringValue();
+      IndexableField field = doc.getField(LuceneDocumentGenerator.FIELD_RAW);
+      String content = doc == null ? null : field.stringValue();
```
tuzhucheng: `doc` can never be `null` at this point because if it was `null` two lines earlier `doc.getField` would throw NullPointerException? Perhaps you meant `field`?

lintool: good catch.

lintool:
@tuzhucheng give me a shipit on the review so I can merge?

lintool:
@tuzhucheng Thanks

I think I'll expose this feature in the README before merging.

Victor0118:
LGTM!

Issue #227, opened 2018-06-08
Reimplement Axiomatic Approaches in Web Track


lintool:
Original TREC paper:

Evaluating the Effectiveness of Axiomatic Approaches in Web Track
https://trec.nist.gov/pubs/trec22/papers/udel_fang-web.pdf

Broader context is the TREC CENTRE track:
http://www.centre-eval.org/trec2018/index.html

Peilin-Yang:
Jimmy said:
> so here's my idea - if you have free cycles, let's participate in CENTRE and have you implement your > own paper :slightly_smiling_face:
> then we can write a paper title "How hard is it to implement your own work?"
> And I think we can come up with a neural formulation of semantic matching - 
> it's at its core learning a scoring function, right? shouldn't be too hard...

Peilin-Yang:
Algorithm code has been merged as per #247 .
But as @lintool mentioned that it'd be useful to fix the random seed in the algorithm and directly integrate the effectiveness numbers in the docs. That way, it is easier for regression test.

Peilin-Yang:
Making deterministic results of axiomatic reranking is merged in #275 . 
Will do the regression tests.

PR #247, opened 2018-06-16
Axiomatic reranking

Peilin-Yang:
This PR is for adding axiomatic reranking, see https://trec.nist.gov/pubs/trec22/papers/udel_fang-web.pdf 
One thing to notice is that I do not include performances references. The reason: The algorithm itself intentionally has a random reranking documents pool construction which makes it hard to reproduce in nature. Also, the algorithm typically involves in the parameter tuning of `beta` and this also introduces variables, e.g. how to train/test is totally up to the user of the algorithm.
But I do run it against all collections. The parameter `beta` is set to a hard value 0.4 which is an empirical value which is more likely to perform well on TREC Adhoc Collections. And the results confirm this: For TREC Adhoc Collections the performances are better than RM3 in general while for TREC Web Collections it may or may not perform very well.


lintool:
Can you post some effectiveness results here?

So you were able to replicate your old work here without any issues?

That's not going to make a particularly interesting CENTRE paper... "We replicated it successfully, no major issues!" 😄 

Was there anything along the way worth noting? E.g., a corner case you missed before, some detail you forgot to mention in the paper?

Did you write this code looking at previous code you had, or basically this is from-scratch code?

In terms of CR - this PR has both the Axiomatic reranker as well as minor refactoring to other parts of the code base - to keep things clean, I would rather separate out the other edits into another PR. Is that okay?

Peilin-Yang:
Effectiveness reference:

AP:

Collection | BM25+RM3 | BM25+Axiom
----|------|------|
Disk12 51-100 | 0.2581 | 0.2667
Disk12 51-100 | 0.2537 | 0.2678
Disk12 51-100 | 0.3136 | 0.3389
Robust04 701-750 | 0.2729 | 0.2904
Wt10g 451-550 | 0.2188 | 0.1956
Gov2 701-750 | 0.2952 | 0.2490
Gov2 751-800 | 0.3835 | 0.3123
Gov2 801-850 | 0.3409 | 0.2966
CW09b 51-100 | 0.1065 | 0.0762
CW09b 101-150 | 0.1140 | 0.0755
CW09b 151-200 | 0.1336 | 0.1080
CW12 201-250 | 0.1503 | 0.1058
CW12 251-300 | 0.2431 | 0.1497

> So you were able to replicate your old work here without any issues?

I actually did not try to replicate the TREC paper.... In the paper, we used another external resources as the reranking documents pool. The ranking function was F2EXP (I've merged that to Lucene but not Anserini) instead of BM25. The parameter was tuned too. I am actually not quite interested in replaying whatever I have done before - adding the algorithm in this repo is more meaningful.

> Did you write this code looking at previous code you had, or basically this is from-scratch code?

For the algorithm I looked at my previous code (on github). But I basically changed everything including the syntax to make it working with Lucene.

> In terms of CR - this PR has both the Axiomatic reranker as well as minor refactoring to other parts of the code base - to keep things clean, I would rather separate out the other edits into another PR. Is that okay?

Not a problem. I'll see if the refactor is necessary.

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+  private final Analyzer analyzer;
+  private final String field;
+
+  private int R = 30; // factor that used in extracting random documents, we will extract R*M randomly select documents
+  private int M = 20; // number of top documents in initial results
+  private int L = 1000; // top similar terms
+  private int K = 20; // number of expansion terms
+  private float beta; // scaling parameter
+
+  public AxiomReranker(Analyzer analyzer, String field, String stoplist, Boolean fromResource, float beta) {
```
lintool: You don't do anything with `stoplist`?

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+        builder.append(term + "^" + score + " ");
+      }
+      String queryText = builder.toString().trim();
+
+      if (queryText.isEmpty()) {
+        LOG.info("[Empty Expanded Query]: " + context.getQueryTokens());
+        queryText = context.getQueryText();
+      }
+
+      StandardQueryParser p = new StandardQueryParser();
```
lintool: Can you rewrite to build the query expression directly using the API?
We've had issues with the query parser before, where special characters choke it up... and then you have to escape, etc.

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+  private Map<String, Set<Integer>> extractTerms(Set<Integer> docIds, RerankerContext context,
+                                                 Pattern filterPattern) throws Exception, IOException {
+    IndexSearcher searcher = context.getIndexSearcher();
+    IndexReader reader = searcher.getIndexReader();
+    Map<String, Set<Integer>> termDocidSets = new HashMap<>();
+    for (int docid : docIds) {
+      Terms terms = reader.getTermVector(docid, LuceneDocumentGenerator.FIELD_BODY);
+      if (terms == null) {
+        LOG.info("Document vector not stored for docid: " + docid);
+        continue;
```
lintool: Do we actually want to throw an exception here? If doc vector not stored, we can't really do query expansion...

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+      }
+      TermsEnum te = terms.iterator();
+      if (te == null) {
+        LOG.info("Document vector not stored for docid: " + docid);
+        continue;
+      }
+      while ((te.next()) != null) {
+        String term = te.term().utf8ToString();
+        if (term.length() < 2) continue;
+        if (!term.matches("[a-z]+")) continue;
```
lintool: This is some type of heuristic I assume? Document it a bit... I assume this cuts down on noisy terms?

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+        //System.out.println(te.term().utf8ToString() + " " + te.totalTermFreq());
```
lintool: LOG or remove before commit

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+   *
+   *  @param docIds          The reranking pool, see {@link #selectDocs} for explanations
+   *  @param context         An instance of RerankerContext
+   *  @param filterPattern   A Regex pattern that terms are collected only they matches the pattern, could be null
+   *
+   *  @return A Map of <term -> Set<docId>> kind of a small inverted list where the Set of docIds is where the term occurs
+   */
+  @VisibleForTesting
+  private Map<String, Set<Integer>> extractTerms(Set<Integer> docIds, RerankerContext context,
+                                                 Pattern filterPattern) throws Exception, IOException {
```
lintool: I think we want four space indent here (which is what you do below, actually)

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+      queryTermsCounts.put(qt, queryTermsCounts.getOrDefault(qt, 0) + 1);
+    }
+
+    Set<Integer> allDocIds = new HashSet<>();
+    for (Set<Integer> s : termInvertedList.values()) {
+      allDocIds.addAll(s);
+    }
+    int docIdsCount = allDocIds.size();
+
+    List<PriorityQueue<Pair<String, Double>>> allTermScoresPQ = new ArrayList<>();
```
lintool: Can you write in a comment what this deeply nested data structure is?

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+    for (Map.Entry<String, Integer> q : queryTermsCounts.entrySet()) {
+      String queryTerm = q.getKey();
+      int qtf = q.getValue();
+      if (termInvertedList.containsKey(queryTerm)) {
+        PriorityQueue<Pair<String, Double>> termScorePQ = new PriorityQueue<>(new ScoreComparator());
+        double selfMI = calMutualInformation( termInvertedList.get(queryTerm), termInvertedList.get(queryTerm), docIdsCount );
+        for (Map.Entry<String, Set<Integer>> termEntry : termInvertedList.entrySet()) {
+          double score = 1.0 * qtf;
+          if (!termEntry.getKey().equals(queryTerm)) { // do not include the query term itself ...
```
lintool: I think it's cleaner if you rewrite with `if (termEntry.getKey().equals(queryTerm)) continue` - cuts down on the indentation level.

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+
+    for (Map.Entry<String, Double> termScore : resultTermScores.entrySet()) {
+      termScore.setValue(termScore.getValue()/sumScore); // weighted
+    }
+
+    return resultTermScores;
+  }
+
+  @VisibleForTesting
+  private double calMutualInformation(Set<Integer> docIdsX, Set<Integer> docIdsY, int totalDocCount) {
```
lintool: cal -> compute for clarity?

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+    int x1 = docIdsX.size(), y1 = docIdsY.size(); //document that x occurres
```
lintool: docIds -> docids to stay consistent with rest of code base?

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+    int x0 = totalDocCount-x1, y0 = totalDocCount-y1; //document num that x doesn't occurres
+
+    if( x1 == 0 || x0 == 0 || y1 == 0 || y0 == 0 ) {
```
lintool: space after `if`

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,302 @@
...
+      return 0;
+    }
+
+    float pX0 = 1.0f * x0 / totalDocCount;
+    float pX1 = 1.0f * x1 / totalDocCount;
+    float pY0 = 1.0f * y0 / totalDocCount;
+    float pY1 = 1.0f * y1 / totalDocCount;
+
+    //get the intersection of docIds
+    Set<Integer> tmp = new HashSet<>(docIdsX);
```
lintool: hrm... can we have a better variable name than `tmp`?

src/main/java/io/anserini/rerank/AxiomReranker.java
Peilin-Yang: No, the algorithm is agnostic to stopwords.

src/main/java/io/anserini/rerank/AxiomReranker.java
Peilin-Yang: sure

src/main/java/io/anserini/rerank/AxiomReranker.java
Peilin-Yang: No...
When I randomly pick documents from the whole index I used an array [1, maxDocId]. What I found is that some docIds are missing from the consecutive integer array. My guess is that when indexing the document there might be problems, e.g. empty doc, that leave the document not storing the termVector.

src/main/java/io/anserini/rerank/AxiomReranker.java
Peilin-Yang: Yes, will add some comments.

src/main/java/io/anserini/rerank/AxiomReranker.java
Peilin-Yang: sure

src/main/java/io/anserini/rerank/AxiomReranker.java
```diff
@@ -0,0 +1,306 @@
...
+                builder.append(term + "^" + score + " ");
+            }
+            String queryText = builder.toString().trim();
+
+            if (queryText.isEmpty()) {
+                LOG.info("[Empty Expanded Query]: " + context.getQueryTokens());
+                queryText = context.getQueryText();
+            }
+
+            StandardQueryParser p = new StandardQueryParser();
```
Peilin-Yang: The expansion query is explicitly constructed by us - that means we do know there is no characters that can break the query parser.

Peilin-Yang:
I think the new indentation sweep out the old comments....
Basically, I've addressed all your concerns except the `StandardQueryParser`. 
The reason: The expansion query is explicitly constructed by us - that means we do know there is no characters that can break the query parser.

lintool:
One final comment -  we do two-space indent. Please change formatting - otherwise go ahead and merge. Thanks!

Peilin-Yang:
Okay, I actually want to add documentation before merging.

lintool:
Do you want separate docs? Why not integrate directly into existing docs, as part of the current regression suite?

Peilin-Yang:
You mean the effectiveness numbers output by axiomatic reranking? As I mentioned in the doc, this algorithm has its non-deterministic component which cannot be exactly reproduced...

Peilin-Yang:
@lintool Is that possible to merge this?
For regression test, I shall leave the current issue #227 open and create a new PR later on.... 
Now there are several PRs related to the TREC and it is a bit messy to hop between branches....

lintool:
shipit!

PR #248, opened 2018-06-16
Regression script example (for disks12)

lintool:
Per #245 

PR mostly for comments, not for merging.

It's ugly, but it automates what I'm doing now by hand. It's possible to refactor to reduce boilerplate, but regardless there will be lots of repetition.

The important thing is - at the end of the script, we get e2e replication numbers:

```
Topics  51-100: bm25     : map : 0.2249 0.2250  !
Topics 101-150: bm25     : map : 0.2004 0.2004
Topics 151-200: bm25     : map : 0.2572 0.2572
Topics  51-100: bm25+rm3 : map : 0.2581 0.2581
Topics 101-150: bm25+rm3 : map : 0.2537 0.2537
Topics 151-200: bm25+rm3 : map : 0.3136 0.3136
Topics  51-100: ql       : map : 0.2178 0.2178
Topics 101-150: ql       : map : 0.2014 0.2014
Topics 151-200: ql       : map : 0.2532 0.2532
Topics  51-100: ql+rm3   : map : 0.2468 0.2468
Topics 101-150: ql+rm3   : map : 0.2438 0.2437  !
Topics 151-200: ql+rm3   : map : 0.2941 0.2941
Topics  51-100: bm25     : p30 : 0.4493 0.4493
Topics 101-150: bm25     : p30 : 0.4220 0.4220
Topics 151-200: bm25     : p30 : 0.4747 0.4747
Topics  51-100: bm25+rm3 : p30 : 0.4800 0.4800
Topics 101-150: bm25+rm3 : p30 : 0.4560 0.4560
Topics 151-200: bm25+rm3 : p30 : 0.5067 0.5067
Topics  51-100: ql       : p30 : 0.4453 0.4453
Topics 101-150: ql       : p30 : 0.4160 0.4160
Topics 151-200: ql       : p30 : 0.4653 0.4653
Topics  51-100: ql+rm3   : p30 : 0.4633 0.4633
Topics 101-150: ql+rm3   : p30 : 0.4393 0.4393
Topics 151-200: ql+rm3   : p30 : 0.4987 0.4993  !
```

The (!) marks values that are different. A human can look at it and see if there's anything to worry about...

Thoughts?

PR #251, opened 2018-06-19
Regression scripts for various collections

lintool:
Similar to scripts as part of #248 - allows us to easily run end-to-end regression experiments on tuna.


lintool:
Added high-level documentation as requested. Anything else I'm missing?

Peilin-Yang:
Is this PR just for consolidating the e2e scripts or should we also update the numbers in the docs?

lintool:
Just the scripts. I'll update all the numbers once you CR #253 
Sg?

Issue #252, opened 2018-06-19
Address end-to-end effectiveness instability issues


lintool:
In running end-to-end regression experiments, we observe a small amount of effectiveness instability (MAP and P3) - differences are small, mostly in the fourth decimal place.

This is due to tie-breaking behavior of documents with the same score, which is dependent on the order the documents were added to the index (which is non-deterministic due to multi-threaded indexing).

We should address this issue.

Peilin-Yang:
> This is due to tie-breaking behavior of documents with the same score, which is dependent on the order the documents were added to the index (which is non-deterministic due to multi-threaded indexing).

`trec_eval` ranks the documents by scores and breaks the ties using the external document ids string in descending order. See the test: https://github.com/castorini/Anserini/blob/master/src/test/java/io/anserini/eval/RankingResultsTest.java#L30

So the tie-breaking behavior of `trec_eval` is deterministic. 
In order to support the feature such as treating the docId as Integer or breaking the tie by docId in ascending order, I have added these features and tests. See: https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/eval/EvalArgs.java#L33-L42

lintool:
Issue fixed for newswire, gov2, and wt10g collections, complete with update regression experiments - commit 014ebac27539f80f0d3fe980c780f6c3aa0fa367

Merged as part of #253

PR #253, opened 2018-06-19
Reranker deterministically breaks scoring ties

lintool:
See discussion in #252


src/main/java/io/anserini/rerank/lib/TiebreakerReranker.java
```diff
@@ -0,0 +1,92 @@
...
+      int i = 0;
+      int dup = 0;
+      float prevScore = 0.0f;
+      for (Result result : sortedResults) {
+        float curScore = result.score;
+
+        // If we encounter ties, we want to perturb the final score a bit.
+        // Note that we can't use equality comparison directly, because in the case of multiple
+        // ties, we would have perturbed the scores, leading the scores to being not equal.
+        if ( prevScore == 0.0f || prevScore - curScore > 1e-4f ) {
```
Peilin-Yang: I am not sure I understand this piece of code:
1. Why it is necessary to have the if-else block in the first place? The comparison is already handled by `Result` class, no? The following questions assume it do need further tweak on the scores....
2. Lucene's ranking score has precision of 1e-6 so why using 1e-4 here?
3. For `curScore = curScore - 1e-6f * dup;` If there are more than 2 docs with the same score 1.000001 and the document right below them has a slightly lower score 1.000000, then by lowing the `curScore` a bit at a time `1e-6f * dup` would make some docs scores lower than 1.000000. And this essentially breaks the whole ranking list, no?

lintool: Added explanation in code.

lintool:
Added explanation of my approach, finalized regression results on more collections.

lintool:
Updated this PR with score tie-breaking across _k_. Technique is to retrieve 2 * _k_, resorts, and then truncates to top _k_.

Why 2 * _k_? There are examples of hundreds of documents with the same score... so 2 * _k_ "just to be safe".

src/main/java/io/anserini/rerank/lib/TiebreakerReranker.java
```diff
@@ -0,0 +1,131 @@
...
+        //   1 docA 23.4393   (dup=0)
+        //   2 docS 22.0874   (dup=0)
+        //   3 docT 22.0874 - (dup=1)*1e-6
+        //   4 docZ 21.6025   (dup=0)
+        //
+        // Note that we can't use equality comparison directly to detect duplicates, because in the case of multiple
+        // ties, we would have perturbed the scores, leading the scores to not be equal (hence we check for score
+        // difference greater than 1e-4).
+        //
+        // Why 1e-4 and 1e-6? If we make the former larger, than we lose score resolution in the original score. If we
```
Peilin-Yang: What if there are more than 100 documents with the same score 22.0874, and the document below has score 22.0873?

lintool: Correct... this trick can't handle more than 100 score ties... unless we do 1e-3 and 1e-6, in which case we can support up to 1000 scoring ties..

Peilin-Yang: Honestly, I do not think the manipulation here is necessary....and it is hard to maintain and debug.
`Result` class can handle most cases and the only problem with it is that it cannot handle score ties across the retrieval hits.
Now you doubled the first round retrieval results and `Result` should be able to solve the problem mentioned above.
The only problem for now is how safe is the number of hits in the first round - I don't know and nobody knows... But 1000 should be a pretty safe bet as for MB collections I've seen more than 100 score ties but I've never seen more than several hundreds score ties.

lintool:
@Peilin-Yang okay, this is ready for review again.

We are now able to have perfectly repeatable runs, both feedback and non-feedback.
Note that I left rerankers that are unused in the code. Might be useful later (we can always remove).

src/main/java/io/anserini/rerank/lib/ScoreTiesAdjusterReranker.java
```diff
@@ -0,0 +1,65 @@
...
+import io.anserini.rerank.ScoredDocuments;
+
+public class ScoreTiesAdjusterReranker implements Reranker {
+  @Override
+  public ScoredDocuments rerank(ScoredDocuments docs, RerankerContext context) {
+    int dup = 0;
+    for (int i=0; i<docs.documents.length; i++) {
+      docs.scores[i] = Math.round(docs.scores[i] * 1e4f) / 1e4f;
+
+      // If we encounter ties, we want to perturb the final score a tiny bit.
```
Peilin-Yang: nit: can we make these comments shorter per line? so that I can view them on my phone without swipe left and right?

lintool: I'm going by IntelliJ default, which is 120 characters... how much narrower do you want to make it?

Peilin-Yang: Mine is 100?

lintool:
@Peilin-Yang I cleaned up code - still running final regression experiments, but otherwise good to go.

PR #256, opened 2018-06-20
POM cleanup

lintool:
+ binds `appassembler:assemble` to package phase explicitly so it'll run by default.
+ unbinds shader from package phase so it _won't_ run by default - speeds up build (fat jar is only needed if we want to build the Python bridge)


README.md
```diff
@@ -7,7 +7,7 @@ Anserini
 Build using Maven:
 
 ```
-mvn clean package appassembler:assemble
+mvn clean package
 ```
 
 The `eval/` directory contains evaluation tools and scripts, including `trec_eval`. Before using `trec_eval`, unpack and compile it.
```
Peilin-Yang: Can we add more specific commands here?
`tar xvfz trec_eval.9.0.tar.gz && cd trec_eval.9.0 && make`

pom.xml
```diff
@@ -45,40 +45,37 @@
...
+            <configuration>
+              <extraJvmArguments>-Xms512M -Xmx24576M</extraJvmArguments>
+              <programs>
+                <program>
+                  <mainClass>io.anserini.index.IndexCollection</mainClass>
+                  <id>IndexCollection</id>
+                </program>
+                <program>
+                  <mainClass>io.anserini.search.SearchCollection</mainClass>
+                  <id>SearchCollection</id>
```
Peilin-Yang: I think you had accidentally removed `Eval` and `IndexUtils` in one recent PR?
Can you add them back?

lintool:
Addressed CR.

pom.xml
lintool: done

Issue #259, opened 2018-06-22
Refactoring of SearchCollections and Reranker module


lintool:
SearchCollections has gotten a bit too spaghetti in terms of code quality, with complex interleaved code paths for tweets and non-tweets - this needs a bit of cleanup.

Also, the various reranker modules could use a bit of reorganization/cleanup also.

lintool:
Fixed as part of #262 commit b330b47c55c280e75947552edbcd9211efcb74fc

PR #262, opened 2018-06-22
Refactoring of SearchCollection and Reranker module per issue #259

lintool:
+ Refactored SearchCollection: pushed logic out of main and into constructor; added separate code paths for tweets and non-tweets search - will make it easier to integrate tweet regressions.
+ Moved all rerankers into a lib sub-package.


lintool:
@Peilin-Yang Might make sense for you to hold off merging until this one, so you can hook into the correct code path?

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -85,85 +77,102 @@
...
-      List<String> queryTokens = AnalyzerUtils.tokenize(analyzer, queryString);
-      if (searchArgs.searchtweets) { // This is ugly, but we have to reform the tweet query here for reranking
-        query = AnalyzerUtils.buildBagOfWordsQuery(FIELD_BODY, analyzer, queryString);
+    for (Map.Entry<K, Map<String, String>> entry : topics.entrySet()) {
+      K qid = entry.getKey();
+      String queryString = entry.getValue().get(args.topicfield);
+
+      ScoredDocuments docs;
+      if (args.searchtweets) {
+        docs = search(searcher, qid, queryString, Long.parseLong(entry.getValue().get("time")));
```
Peilin-Yang: It seems that you would like to separate searchTweets from search for other collection, right?
How about naming them differently?

lintool: Here's what I'm thinking - I believe searchTweets is actually just search with a temporal filter. So  to generalize we should have two search methods - search with and without Filter query. This can just be overloaded methods.

Thoughts?

Issue #258, opened 2018-06-21
Index of Microblog13 collection has duplicated documents


Peilin-Yang:
At least there are two documents that are consistent across multiple indexing tries: 301142378562068480, 313796170101174272.
Since the internal docids in Anserini are non-deterministic these duplicated doucments are hard to find. What I did is to log the internal docIds when doing the retrieval and look back-n-forth between logs and the output ranking list. 
Then we can use `target/appassembler/bin/IndexUtils -index lucene-index.Tweets2013.pos+docvectors.keepUrls.stemming -convertLuceneDocidToDocid xxxxx` and `target/appassembler/bin/IndexUtils -index lucene-index.Tweets2013.pos+docvectors.keepUrls.stemming -convertLuceneDocidToDocid yyyyy` to verify that there external docId are both 301142378562068480 or 313796170101174272.

For now I've no idea why this could happen but it seems to be an indexing issue.

lintool:
This is because a tweet might be delivered multiple times in the streaming API, e.g., due to a network hiccup.

Peilin-Yang:
I thought docid is the unique identifier of the indexed document and thus be unique.... seems to be not the case

lintool:
No, Lucene treats it just as any other field - it has special semantics to us, but not for Lucene. Only the internal docid is unique.

lintool:
Does this cause a specific issue? To really address this, we'd need to keep a global hash set of all ids encountered - however, since indexing is multi-threaded, this will need to be sync'ed and thus slow indexing by a lot...

Peilin-Yang:
> Does this cause a specific issue? 

Yes, when I did the reranking the duplicated documents are in the ranking list. And this fails `trec_eval`.

I think I already know how to solve it - `IndexWriter.updateDocument(new Term("id", id), doc)`.  And testing on MB13 seems working. But this should be an optional parameter for `IndexCollection` as it slows down the indexing (the updateDocument basically delete the existing doc with the same id from the index and add the new one).
I will push the PR soon.

lintool:
Yes, this should be a separate code path and we should measure the performance effects...

Peilin-Yang:
fixed per #263

PR #263, opened 2018-06-22
Add uniqueDocId option to IndexCollection as per #258

Peilin-Yang:
Testing result on Microblog 2013: 

**Timing** 
master: `01:00:07`
this PR: `02:47:12`

**Index Stats**
master: 
```
documents:             203162948
documents (non-empty): 203160281
unique terms:          122988939
total terms:           2099261924
```

this PR:
```
documents:             203145941
documents (non-empty): 203143274
unique terms:          122988939
total terms:           2099082542
```

### DIFF
--- a/src/main/java/io/anserini/index/IndexCollection.java
+++ b/src/main/java/io/anserini/index/IndexCollection.java
@@ -29,6 +29,7 @@ import org.apache.lucene.document.Document;
 import org.apache.lucene.index.ConcurrentMergeScheduler;
 import org.apache.lucene.index.IndexWriter;
 import org.apache.lucene.index.IndexWriterConfig;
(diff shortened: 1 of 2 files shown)


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -73,6 +74,11 @@
 
     // optional arguments
 
+    @Option(name = "-uniqueDocId", usage = "remove duplicated documents with the same doc id when indexing. " +
+      "please note that this option may slow the indexing a lot and if you are sure there is no " +
```
lintool: I think we've been using `docid` not `docId`, so the option should be `-uniqueDocid`?

Issue #243, opened 2018-06-14
Clean up current QA demo Python and JavaScript code from repo


lintool:
Per #239 the QA demo code (Python and JavaScript) should be moved out of Anserini, e.g.,

https://github.com/castorini/Anserini/tree/master/src/main/js
https://github.com/castorini/Anserini/tree/master/src/main/python

tuzhucheng:
Which repo should they be moved to, or a new repo?

lintool:
Your call - I suppose the options are Castor, or another "app" repo. Does it make sense to keep Castor pure DL models? Thoughts?

tuzhucheng:
Separate repo makes sense. Castor and Anserini are core IR/DL repos and the demos use these repos to build an application.

lintool:
Okay, so our general layout will be like this?

```
castorini/
  |--- Castor/
  |--- Castor-data/
  |--- Anserini/
  |--- app1/
  |--- app2/
...
```

If we agree on this layout, then we can hard code relative paths all over.
Thoughts?

tuzhucheng:
Yep. Currently we already do this for Castor-data and Castor, but the relative paths should just be defaults in configuration files or command line args.

tuzhucheng:
Likewise should https://github.com/castorini/Castor/tree/master/anserini_dependency be moved into the app repo as well? There is a copy of the demo in here worked on @rosequ and @MeowFei.

tuzhucheng:
Fully resolved by #267

PR #264, opened 2018-06-23
Remove E2E QA Demo from Anserini

tuzhucheng:
Remove end-to-end QA demo from Anserini (moved to [piqa](https://github.com/castorini/piqa)).

Related to #243 - All demo code (JS/Python) are completely gone, but other Python files exist (`run_regression*.py` and `run_trecqa.py`)


lintool:
Can you remove `requirements.txt` completely.
Also remove `run_trecqa.py` but ask @rosequ for review - I think that file is no longer necessary.

tuzhucheng:
Hi @rosequ please confirm if `run_trecqa.py` is still needed - if not I'll remove it.

PR #265, opened 2018-06-24
Add whitelistDocids option to indexing

Peilin-Yang:
This is useful if we'd like to only add specific documents to the index, e.g. when reproducing my TREC2013 web track. 
Tested on CW12 and disk12 (with and without)

### DIFF
--- a/src/main/java/io/anserini/index/IndexCollection.java
+++ b/src/main/java/io/anserini/index/IndexCollection.java
@@ -103,9 +103,8 @@ public final class IndexCollection {
     @Option(name = "-memorybuffer", usage = "memory buffer size")
     public int memorybufferSize = 2048;
 
-    @Option(name = "-whitelistFile", usage = "when specified only the docids contained in the " +
-      "file will be included in the index, one docid per line.")
-    public String whitelistFile = null;
+    @Option(name = "-whitelist", usage = "file containing docids, one per line; only specified docids will be indexed.")
+    public String whitelist = null;
 
     @Option(name = "-tweet.keepRetweets", usage = "boolean switch to keep retweets while indexing")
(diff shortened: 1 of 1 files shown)


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -100,6 +92,21 @@
...
+    @Option(name = "-uniqueDocid", usage = "remove duplicated documents with the same doc id when indexing. " +
+      "please note that this option may slow the indexing a lot and if you are sure there is no " +
+      "duplicated document ids in the corpus you shouldn't use this option.")
+    public boolean uniqueDocid = false;
+
+    @Option(name = "-memorybuffer", usage = "memory buffer size")
+    public int memorybufferSize = 2048;
+
+    @Option(name = "-whitelistFile", usage = "when specified only the docids contained in the " +
+      "file will be included in the index, one docid per line.")
```
lintool: How about just `-whitelist`? "File" is kinda redundant.

Also, start the description with a noun, e.g., "file containing docids, one per line; only specified docids will be indexed."

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -198,6 +206,7 @@ public IndexCollection(IndexCollection.Args args) throws Exception {
     LOG.info("Store transformed docs? " + args.storeTransformedDocs);
     LOG.info("Store raw docs? " + args.storeRawDocs);
     LOG.info("Optimize (merge segments)? " + args.optimize);
+    LOG.info("Whitelist File: " + args.whitelistFile);
```
lintool: remove "File"

Issue #254, opened 2018-06-20
SourceDocument refactoring to get rid of readNextRecord


lintool:
I don't remember who originally designed the interface, and it could have very well been me - but it doesn't make sense to me that `readNextRecord` is a method in `SourceDocument`.

`SourceDocument` is an object within a container class (`FileSegment`). It shouldn't need to know about "what's next" - that should be the responsibility of the container?

borislin:
I'll start working on this after #332 is merged

lintool:
@borislin sg

As part of this issue, we should clean up the semantics of `FileSegment`. It's supposed to be an iterator over documents, right? Except we haven't implemented "true" iterators - `hasNext()` doesn't actually do much... we basically keep on calling `next()` until we run out of file... and use `null` to signal no more record. This is janky.

In a "true" iterator impl, `hasNext()` should be advancing the record, and `next()` actually returns it...

Let's make it actually work correctly. And properly tested.

borislin:
@lintool

So we should have a buffer variable `bufferRecord` to store the record obtained in `hasNext()`.

In `hasNext()`, if `bufferRecord` is null and there are more records in the file, we should store the next record into `bufferRecord`. Otherwise, don't advance the record and just return (to tackle the case when user calls `hasNext()` multiple times without calling any `next()`). Also, return `false` when `bufferRecord` is null and the file has reached `EOF`.

In `next()`, we first check if `bufferRecord` is null. If yes, call `hasNext()` first. If no, return `bufferRecord` and set `bufferRecord` to null.

Good?

lintool:
sg, and throw `NoSuchElementException` when you run off the end...

borislin:
Refactored in #352

PR #266, opened 2018-06-24
Json adaptor

Peilin-Yang:
A general Json ingestion adaptor which can deal with either JsonArray (a bunch of documents in one file) and JsonObject (one document in one file).


src/test/java/io/anserini/document/JsonDocumentArrayTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.junit.After;
+import org.junit.Before;
+
+
+public class JsonDocumentArrayTest extends DocumentTest {
+  private String sampleFile = "sampleJsonArray.json";
+
+  @Before
+  public void setUP() throws Exception {
```
lintool: `setUP` -> `setup`

same below.

pom.xml
```diff
@@ -168,7 +168,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.4</version>
+      <version>2.8.5</version>
     </dependency>
```
lintool: I started using gson before knowing much about JSON parsers.
IIRC, Jackson has become the industry standard? e.g., doesn't Twitter use Jackson internally?
If so, we should probably switch over.
Probably not in this PR, but file an issue?

src/main/java/io/anserini/document/JsonDocument.java
```diff
@@ -0,0 +1,98 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+/**
+ *
+ * Json adaptor
+ * Inside each file is either a JSON Object (one document) or a JSON Array (multiple documents)
+ * Example of JSON Object:
+ * {
```
lintool: Wrap this in <code> or <pre> or something like that, otherwise Javadoc is going to look ugly when generated... Also add <p> tags.

Maven automatically generates javadocs - take a look at see if it looks reasonable?

src/main/java/io/anserini/document/JsonDocument.java
```diff
@@ -0,0 +1,98 @@
...
+ *   "id": "doc1",
+ *   "contents": "this is the contents."
+ * }
+ * Example of JSON Array:
+ * [
```
lintool: Will we accept the convention of one JSON per line, but not wrapped inside an array?
This is a pretty common case also - e.g., what comes out of the Twitter streaming API.

pom.xml
Peilin-Yang: Yes, twitter uses Jackson. 
Agreed.

src/test/java/io/anserini/document/JsonDocumentArrayTest.java
Peilin-Yang: Yes.
But I think it should be setUp()... it is a override method from `LuceneTestCase`

src/main/java/io/anserini/collection/JsonCollection.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ *
+ * Json adaptor
+ * Inside each file is either a JSON Object (one document) or a Json Array (multiple documents)
+ *
```
lintool: ... or a JSON document on each line?
Get rid of extra spaces.

src/main/java/io/anserini/document/JsonDocument.java
```diff
@@ -0,0 +1,113 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+/**
+ *
+ * Json adaptor
+ * Inside each file is either a JSON Object (one document) or a JSON Array (multiple documents)
+ * Example of JSON Object:
+ * <pre>
```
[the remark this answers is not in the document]
Peilin-Yang: In this case it is a little different: `JsonDocument` loads the whole file at once and return a single document when `readNextRecord` is called....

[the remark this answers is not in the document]
lintool: Sure, but this is probably going to change when we refactor per #254?
I assume you agree with that?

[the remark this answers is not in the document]
Peilin-Yang: I actually haven't think about #254 too much....
But I can move the javadoc to `JsonCollection`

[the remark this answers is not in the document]
lintool: I have no strong preference either way.

Issue #260, opened 2018-06-22
Build regression tests for tweet collections


lintool:
This is dependent on #259 - after cleanup, we need to build the regression test cases for tweets retrieval also.

lintool:
Merged in #269

PR #269, opened 2018-06-25
Patch for issue #260: Build regression tests for tweet collections

lintool:
See extensive discussion in #260 

+ As a result of this, we're able to remove `RemoveRetweetsTemporalTiebreakReranker`
+ Also, light refactoring of `RerankerContext`


src/main/java/io/anserini/index/generator/TweetGenerator.java
```diff
@@ -153,7 +152,11 @@ public Document createDocument(TweetDocument tweetDoc) {
     Document doc = new Document();
     doc.add(new StringField(FIELD_ID, id, Field.Store.YES));
 
+    // We need this to break scoring ties.
     doc.add(new LongPoint(StatusField.ID_LONG.name, tweetDoc.getIdLong()));
+    doc.add(new NumericDocValuesField(StatusField.ID_LONG.name, tweetDoc.getIdLong()));
+    //doc.add(new StoredField(StatusField.ID_LONG.name, tweetDoc.getIdLong()));
```
Peilin-Yang: remove

lintool:
@Peilin-Yang  I had some conflicts with your axiomatic PR. I think I resolved the issues, but as a sanity check and you rerun your axiomatic runs to make sure?

Peilin-Yang:
@lintool I've pushed some changes to this branch, mainly for axiomatic reranking. 
The main change is to Remove the query filter when ranking the external index while keeping the tie breaking settings.

lintool:
@Peilin-Yang lgtm

I'm running regression one more time - are we good to merge if that passes? Need a shipit from you.

PR #274, opened 2018-06-26
Add rerank cutoff for initial round of the search

Peilin-Yang:
This is esp. useful when applying query expansion on large dataset such as CluwWeb.


src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -50,6 +50,11 @@
   @Option(name = "-hits", metaVar = "[number]", required = false, usage = "max number of hits to return")
   public int hits = 1000;
 
+  @Option(name = "-rerankcutoff", metaVar = "[number]", required = false, usage = "max number of hits " +
```
lintool: `-rerankCutoff`?

lintool:
My more general comment here is that, shouldn't these be reranker-specific?
We have:

```
  private int fbTerms = 20;
  private int fbDocs = 50;
  private float originalQueryWeight = 0.6f;
```
So that means we should actually have `-rm3.fbTerms`, `-rm3.fbDocs`, `-rm3.originalQueryWeight`, right?

Peilin-Yang:
Yes, I think it makes more sense to specify RM3's options in SearchArgs.
But `rerankCutoff` is more general for any reranking model. 
So we probably can have `-rerankCutoff`, `-rm3.fbTerms`, `-rm3.originalQueryWeight`?

lintool:
@Peilin-Yang I think you still want to change `-rerankcutoff` to `-rerankCutoff`?
A quick sanity check on a few quick regressions?

Otherwise, ready to be merged.

PR #275, opened 2018-06-27
Add search options to support deterministic result for axiomatic reranking

Peilin-Yang:
This is critical for reproducibility


Peilin-Yang:
Waiting for #269 to be merged first.

src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -84,6 +84,15 @@
   @Option(name = "-axiom", usage = "use Axiomatic query expansion model for the reranking")
   public boolean axiom = false;
 
+  @Option(name = "-axiom.decisive", usage = "decisive the expansion terms axiomatic reranking")
```
lintool: `-axiom.deterministic` sounds more accurate?

lintool: Also, you don't want the seed to be settable in the cmd line?

Peilin-Yang: Well, I just feel nobody is going to use it unless for deterministic results.... 
But I can add it anyway.

lintool:
@Peilin-Yang Good for merging.

Are you planning on updating the documentation?
What about regression tests (bm25+ax, ql+ax)?

PR #277, opened 2018-06-27
Index logging and counter cleanup

lintool:
+ Tweaked logging output during indexing to make less noisy.
+ Documentation and clarification of what each counter means.
+ Better logging of final counter states.

### DIFF
--- a/src/main/java/io/anserini/index/IndexCollection.java
+++ b/src/main/java/io/anserini/index/IndexCollection.java
@@ -124,32 +124,32 @@ public final class IndexCollection {
     /**
      * Counter for successfully indexed documents.
      */
-    public AtomicLong indexedDocuments = new AtomicLong();
+    public AtomicLong indexed = new AtomicLong();
 
     /**
      * Counter for empty documents that are not indexed. Empty documents are not necessary errors;
      * it could be the case, for example, that a document is comprised solely of stopwords.
      */
-    public AtomicLong emptyDocuments = new AtomicLong();
+    public AtomicLong empty = new AtomicLong();
 
     /**
      * Counter for unindexed documents. These are cases where the {@link SourceDocument} returned
      * by {@link Collection.FileSegment} is {@code null} or the {@link LuceneDocumentGenerator}
-     * returned {@code null}.
+     * returned {@code null}. These are not necessarily errors.
      */
-    public AtomicLong unindexedDocuments = new AtomicLong();
+    public AtomicLong unindexed = new AtomicLong();
 
     /**
      * Counter for unindexable documents. These are cases where {@link SourceDocument#indexable()}
      * returns false.
      */
-    public AtomicLong unindexableDocuments = new AtomicLong();
+    public AtomicLong unindexable = new AtomicLong();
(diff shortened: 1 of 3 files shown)


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -152,18 +175,16 @@ public void run() {
...
             counters.unindexableDocuments.incrementAndGet();
             continue;
           }
 
           @SuppressWarnings("unchecked") // Yes, we know what we're doing here.
           Document doc = transformer.createDocument(d);
-
           if (doc != null && (whitelistDocids == null || whitelistDocids.contains(d.id()))) {
             if (args.uniqueDocid) {
               writer.updateDocument(new Term("id", d.id()), doc);
```
Peilin-Yang: What about the `else` case? Should we increment `counters.unindexedDocuments` here?

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -152,31 +182,36 @@ public void run() {
         while (iter.hasNext()) {
           SourceDocument d = iter.next();
           if (d == null) {
-            LOG.info("Null document encountered");
+            counters.unindexed.incrementAndGet();
```
Peilin-Yang: So I think this is a problem: 
Currently, if we encounter null document we do not know if it is EOF or the real null document (b/c of errors or sth). This is mainly due to how `Collection` read the documents. 
I think we should remove this for now and revisit it in #254

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -300,13 +335,21 @@ public void run() throws IOException, InterruptedException {
       }
     }
 
-    LOG.info("Indexed documents: " + counters.indexedDocuments.get());
-    LOG.info("Empty documents: " + counters.emptyDocuments.get());
-    LOG.info("Errors: " + counters.errors.get());
+    if (numIndexed != counters.indexed.get()) {
```
Peilin-Yang: Is this tested for Microblog13 corpus with index option `uniqueDocid`?

lintool:
@Peilin-Yang Addressed CR. Running from scratch:
```
nohup python src/main/python/run_regression_tweets2013.py >& log.tweets2013 &
```

lintool:
@Peilin-Yang Actually, I just realized `run_regression_tweets2013.py` is likely to fail... let me rethink how to handle.

lintool:
@Peilin-Yang Actually, you were correct. For Tweets2013 the indexer did throw an exception because of the duplicate docs issue. However, since by that time the index had already been built, the exception doesn't really do anything and the regression completes as usual.

I've changed to logging a warning instead of throwing an exception. Running one final time and then I'll merge.

Peilin-Yang:
I am a bit confused. Is this related to #282 ?

lintool:
@Peilin-Yang This is related to your `-uniqueDocid` comment.
Regardless, tweets2011 and tweets2013 passed perfectly.

PR #281, opened 2018-06-28
Add --index option to regression scripts to explicitly build index

lintool:
By default, this doesn't rebuild index:
```
nohup python src/main/python/run_regression_disks12.py >& log.disks12 &
```
Instead, it uses indexes from `/tuna1/scratch/`

To explicitly rebuild index, do:
```
nohup python src/main/python/run_regression_disks12.py --index >& log.disks12 &
```

### DIFF
--- a/src/main/python/run_regression_clueweb09b.py
+++ b/src/main/python/run_regression_clueweb09b.py
@@ -1,5 +1,7 @@
 import sys
 import os
+import argparse
+
 from subprocess import call
 
 index_cmd = """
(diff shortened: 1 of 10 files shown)


src/main/python/run_regression_clueweb09b.py
```diff
@@ -37,9 +39,23 @@ def ndcg_eval(qrels, run):
     return round(float(os.popen("eval/gdeval.pl {} {} | grep 'amean' | sed 's/.*amean.//'".format(qrels, run)).read().split(',')[0]), 4)
 
 if __name__ == "__main__":
-    call(index_cmd, shell=True)
+    parser = argparse.ArgumentParser(description='Run regression tests on ClueWeb09b.')
+    parser.add_argument('--index', dest='index', action='store_true', help='rebuild index from scratch')
+    parser.set_defaults(index=False)
```
Peilin-Yang: `store_true` means if we provide `--index` to the script it will be `true`, right?
Then `parser.set_defaults(index=False)` is redundant.

lintool: Well, if the `--index` flag isn't set, we need to default to False, right?

Peilin-Yang: https://docs.python.org/3/library/argparse.html#action

```
import argparse
  

if __name__ == "__main__":
  parser = argparse.ArgumentParser(description='store_true test')
  parser.add_argument('--index', dest='index', action='store_true')

  args = parser.parse_args()
  print(args)
```
`$ python3 test.py` 
`Namespace(index=False)`

`$ python3 test.py --index`
`Namespace(index=True)`

lintool: I stand corrected!

lintool:
Hrm... clueweb12-b13 regression broke at some point in time:

```
trec_eval.form_res_qrels: duplicate docs clueweb12-0000wb-07-13559trec_eval: Can't calculate measure 'map'
True
Topics 701-750: bm25     : map  : 0.0457 0.0457
Topics 751-800: bm25     : map  : 0.0219 0.0219
Traceback (most recent call last):
  File "src/main/python/run_regression_clueweb12-b13.py", line 61, in <module>
    actual_t1_map = trec_eval_metric("map", t1_qrels, "run.web.201-250.bm25+rm3.txt")
  File "src/main/python/run_regression_clueweb12-b13.py", line 30, in trec_eval_metric
    return float(os.popen("eval/trec_eval.9.0/trec_eval -m {} {} {}".format(metric, qrels, run)).read().split("\t")[2].strip())
IndexError: list index out of range
```

Looking into it.

lintool:
False alarm re: clueweb12-b13 regression - see #287

Issue #234, opened 2018-06-12
Strange non-determinism and errors when indexing NYT collection


lintool:
This is the indexing command I ran on `tuna`:

```
nohup sh target/appassembler/bin/IndexCollection -collection TrecCoreCollection \
 -input /tuna1/collections/newswire/NYTcorpus.uncompressed/data/ -generator JsoupGenerator \
 -index lucene-index.nyt.pos+docvectors+rawdocs -threads 16 \
 -storePositions -storeDocvectors -storeRawDocs -optimize \
 >& log.nyt.pos+docvectors+rawdocs &
```

This is the detail of the log:

```
$ tail log.nyt.pos+docvectors+rawdocs
2018-06-12 16:53:38,081 INFO  [main] index.IndexCollection (IndexCollection.java:277) - Indexed documents: 1807636
2018-06-12 16:53:38,082 INFO  [main] index.IndexCollection (IndexCollection.java:278) - Empty documents: 0
2018-06-12 16:53:38,082 INFO  [main] index.IndexCollection (IndexCollection.java:279) - Errors: 48022
2018-06-12 16:53:38,088 INFO  [main] index.IndexCollection (IndexCollection.java:282) - Total 1807636 documents indexed in 00:05:58
```

You can see a number of errors above. These correspond to "0 docs added" below:

```
$ grep "0 docs added" log.nyt.pos+docvectors+rawdocs | head
2018-06-12 16:47:49,499 INFO  [pool-3-thread-2] index.IndexCollection$IndexerThread (IndexCollection.java:160) - 07/0054831.xml: 0 docs added.
2018-06-12 16:47:49,499 INFO  [pool-3-thread-5] index.IndexCollection$IndexerThread (IndexCollection.java:160) - 07/0054915.xml: 0 docs added.
2018-06-12 16:47:49,954 INFO  [pool-3-thread-16] index.IndexCollection$IndexerThread (IndexCollection.java:160) - 07/0054852.xml: 0 docs added.
2018-06-12 16:47:49,967 INFO  [pool-3-thread-6] index.IndexCollection$IndexerThread (IndexCollection.java:160) - 07/0054845.xml: 0 docs added.
2018-06-12 16:47:49,975 INFO  [pool-3-thread-10] index.IndexCollection$IndexerThread (IndexCollection.java:160) - 07/0054905.xml: 0 docs added.
[...]

PR #283, opened 2018-06-28
 fixed non-determinism and errors when indexing NYT collection (#234)

Kytabyte:
As per #234, the bug arises from two parts, one is when handling empty-body documents, and the other is due to a static variable shared across instances of `NYTCorpusDocumentParser` which should not be. Just two simple tweaks to fix the issue.

### DIFF
--- a/src/main/java/io/anserini/document/nyt/NYTCorpusDocumentParser.java
+++ b/src/main/java/io/anserini/document/nyt/NYTCorpusDocumentParser.java
@@ -65,7 +65,7 @@ public class NYTCorpusDocumentParser {
 	private static final String SERIES_NAME_TAG = "series.name";
 
 	/** NITF Constant */
-	private static final DateFormat format = new SimpleDateFormat(
+	private DateFormat format = new SimpleDateFormat(
 			"yyyyMMdd'T'HHmmss");
 
 	/** NITF Constant */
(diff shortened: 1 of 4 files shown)


src/main/java/io/anserini/document/nyt/NYTCorpusDocumentParser.java
```diff
@@ -65,7 +65,7 @@
 	private static final String SERIES_NAME_TAG = "series.name";
 
 	/** NITF Constant */
-	private static final DateFormat format = new SimpleDateFormat(
+	private DateFormat format = new SimpleDateFormat(
 			"yyyyMMdd'T'HHmmss");
```
lintool: Don't we still want `final`?

Kytabyte: Yes.

lintool:
+ Please update `docs/experiments-core17.md`
+ Add regression test script, see `src/main/python/run_regression_robust04.py`

Kytabyte:
Done!

Kytabyte:
And the script works with the result

```
All Topics: bm25     : map : 0.1996 0.1996
All Topics: bm25+rm3 : map : 0.2534 0.2534
All Topics: ql       : map : 0.1928 0.1928
All Topics: ql+rm3   : map : 0.2363 0.2363
All Topics: bm25     : p30 : 0.4207 0.4207
All Topics: bm25+rm3 : p30 : 0.4767 0.4767
All Topics: ql       : p30 : 0.4327 0.4327
All Topics: ql+rm3   : p30 : 0.4640 0.4640
```

Issue #250, opened 2018-06-18
Change WashPost collection source to two-space indent


lintool:
@Kytabyte 

I forgot to note this upon check-in:
https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/document/WashPostDocument.java

Can you change to two-space indent per rest of repo? I'd like you to do it to preserve history info for `git blame`.

PR #284, opened 2018-06-28
reformat code to two-space indent (#250)

Kytabyte:


Kytabyte:
Sorry I just used intelliJ to reformat and didn't notice the difference in headers... now it should be fine...

Issue #268, opened 2018-06-24
Migrate from Gson to Jackson


Peilin-Yang:
Now we depend on Gson as JSON parser for Tweets/JSON parsing.
As Jackson is being more and more as the industrial standard, it is time to evaluate if it is possible to migrate to Jackson.

borislin:
Finished migration in #294

Issue #282, opened 2018-06-28
Exceptions in Tweets2013 indexing


lintool:
Tweets2013 indexing gives a bunch of exceptions:

```
2018-06-28 11:37:47,773 INFO  [main] index.IndexCollection (IndexCollection.java:234) - Collection path: /tuna1/collections/twitter/Tweets2013-corpus/data/
2018-06-28 11:37:47,774 INFO  [main] index.IndexCollection (IndexCollection.java:235) - Index path: lucene-index.tweets2013.pos+docvectors+rawdocs
2018-06-28 11:37:47,775 INFO  [main] index.IndexCollection (IndexCollection.java:236) - CollectionClass: TweetCollection
2018-06-28 11:37:47,775 INFO  [main] index.IndexCollection (IndexCollection.java:237) - Generator: TweetGenerator
2018-06-28 11:37:47,776 INFO  [main] index.IndexCollection (IndexCollection.java:238) - Threads: 32
2018-06-28 11:37:47,776 INFO  [main] index.IndexCollection (IndexCollection.java:239) - Keep stopwords? false
2018-06-28 11:37:47,776 INFO  [main] index.IndexCollection (IndexCollection.java:240) - Store positions? true
2018-06-28 11:37:47,776 INFO  [main] index.IndexCollection (IndexCollection.java:241) - Store docvectors? true
2018-06-28 11:37:47,776 INFO  [main] index.IndexCollection (IndexCollection.java:242) - Store transformed docs? false
2018-06-28 11:37:47,777 INFO  [main] index.IndexCollection (IndexCollection.java:243) - Store raw docs? true
2018-06-28 11:37:47,777 INFO  [main] index.IndexCollection (IndexCollection.java:244) - Optimize (merge segments)? false
2018-06-28 11:37:47,777 INFO  [main] index.IndexCollection (IndexCollection.java:245) - Whitelist: null
2018-06-28 11:37:47,796 INFO  [main] index.IndexCollection (IndexCollection.java:276) - Starting indexer...
2018-06-28 11:37:48,469 INFO  [main] index.IndexCollection (IndexCollection.java:298) - 1416 files found in /tuna1/collections/twitter/Tweets2013-corpus/data
2018-06-28 11:37:55,257 WARN  [pool-2-thread-16] collection.Collection$FileSegment (Collection.java:101) - Exception when parsing document:null
[...]

PR #294, opened 2018-07-04
Patch for issue #268: migrate from Gson to Jackson

borislin:
I've successfully run both disks12 and tweets2011 regression tests.


src/main/java/io/anserini/document/tweet/Delete.java
```diff
@@ -0,0 +1,9 @@
+package io.anserini.document.tweet;
```
lintool: Would it make sense to make these inner classes of `Tweet`?

src/main/java/io/anserini/document/TweetDocument.java
```diff
@@ -16,18 +16,16 @@
```
lintool: How about just rename to `Tweet`? See below.

src/main/java/io/anserini/document/tweet/Tweet.java
```diff
@@ -0,0 +1,81 @@
+package io.anserini.document.tweet;
```
lintool: Rename to `Status` to make consistent with `RetweetedStatus` above?

src/main/java/io/anserini/document/tweet/Tweet.java
```diff
@@ -0,0 +1,81 @@
...
+        // retweet_count might say "100+"
+        // TODO: This is ugly, come back and fix later.
+        retweet_count = Long.parseLong(retweet_count_internal.asText().replace("+", ""));
+      } else if (retweet_count_internal.isNumber()) {
+        retweet_count = retweet_count_internal.asLong();
+      }
+    }
+  }
+
+  public OptionalLong getIn_reply_to_status_id() {
```
lintool: this is pretty ugly... how about we break Java convention and just have method names be the actual tweet fields? e.g., `in_reply_to_status_id`

Getters and setter can just be overloaded?

src/main/java/io/anserini/document/tweet/User.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public String getScreen_name() {
+    return screen_name;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public String getProfile_image_url() {
```
lintool: so, `profile_image_url` per above?

src/main/java/io/anserini/document/JsonDocument.java
```diff
@@ -34,45 +36,38 @@
...
-        raw = parser.parse(new BufferedReader(new FileReader(path))).getAsJsonObject();
-      } catch (IllegalStateException e2) {
-        raw = parser.parse(new BufferedReader(new FileReader(path))).getAsJsonNull();
-      }
+      JsonParser jsonParser = new JsonFactory().createParser(new BufferedReader(new FileReader(path)));
+      ObjectMapper objectMapper = new ObjectMapper();
+      objectMapper.configure(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY, true);
+      node = objectMapper.readTree(jsonParser);
+    } catch (IOException e) {
+      node = null;
```
Peilin-Yang: My understanding is that this block is actually a logic if-else block instead of try-catch block?
If so, we should add comments to not confuse everyone.

src/main/java/io/anserini/document/JsonDocument.java
```diff
@@ -34,45 +36,38 @@
...
-      id = o.get("id").getAsString();
-      contents = o.get("contents").getAsString();
-      i++;
-      return this;
-    } else if (raw.isJsonObject()) {
-      id = raw.getAsJsonObject().get("id").getAsString();
-      contents = raw.getAsJsonObject().get("contents").getAsString();
-      return this;
-    } else {
+    if (node == null || i >= node.size()) {
```
Peilin-Yang: Under that condition can `i >= node.size()`?

Peilin-Yang:
Is #282 solved in this PR?

src/main/java/io/anserini/document/TweetDocument.java
borislin: I think we should keep this unchanged since its name is consistent with other document classes e.g. `TrecDocument`, `TrecwebDocument`.

src/main/java/io/anserini/document/JsonDocument.java
borislin: Yes, will add comments now.

src/main/java/io/anserini/document/JsonDocument.java
borislin: when we have already read and parsed everything in `node` (i.e. `raw` in old code)

Peilin-Yang: So either an array of objects or a single object will be stored in `ArrayNode node`, right?
And that will be handled in the below `else` block. 
My understanding of `i >= node.size()` is that it will only be executed when there is a mixed array and line objects file, no?

borislin: Yes, either an array of JSON objects or a single JSON object will be stored in `ArrayNode node`.

When we have already read and parsed all JSON objects in `node`, we will read and parse from `BufferedReader bRdr`. This is the same logic as `i < raw.getAsJsonArray().size()` in the old code, right?

Peilin-Yang: In the old code, we first categorize the different types:  `raw.isJsonArray()` or `raw.isJsonObject()` or the rest (line json object). There can not be mixed types.
Here, `node==null` means the file contains only line json objects and should be treated as such. So we will never hit `i >= node.size()`.

borislin: When `JsonDocument` is created, will the JSON file contain a mixed content of both array of JSON objects and individual JSON objects?

borislin: I see. So `i >= node.size()` is unnecessary. I'll remove it.

Peilin-Yang: Yea, let's assume that the file is either:

1. An array of json objects (valid JSON format)
2. A single json object (valid JSON format)
3. line json objects (invalid JSON format, only be used for tweet-like files)

Peilin-Yang: Agreed, but the `i >= node.size()` should be in yet another `else if` branch, right?

borislin:
@Peilin-Yang I'm running both tweet2011 and tweet2013 regressions now (slow). So far I haven't seen such exceptions.

src/main/java/io/anserini/document/TweetDocument.java
lintool: okay, sg

src/main/java/io/anserini/document/Status.java
```diff
@@ -0,0 +1,213 @@
...
+
+  @JsonGetter("delete")
+  public Optional<Delete> delete() {
+    return delete;
+  }
+
+  @JsonGetter("coordinates")
+  public Optional<Coordinates> coordinates() { return coordinates; }
+
+  //TODO: to be deleted before merging
```
Peilin-Yang: What to do with this?

src/main/java/io/anserini/document/TweetDocument.java
```diff
@@ -72,79 +67,115 @@ public TweetDocument() {
...
+    Status tweetObj = null;
     try {
-      obj = (JsonObject) JSON_PARSER.parse(json);
-    } catch (JsonSyntaxException e) {
+      tweetObj = mapper
+        .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES) // Ignore unrecognized properties
+        .registerModule(new Jdk8Module()) // Deserialize Java 8 Optional: http://www.baeldung.com/jackson-optional
+        .readValue(json, Status.class);
+    } catch (IOException e) {
+//      LOG.error(e.getMessage());
```
Peilin-Yang: enable the log?

src/main/java/io/anserini/document/TweetDocument.java
```diff
@@ -72,79 +67,115 @@ public TweetDocument() {
...
       return false;
     }
-    if (obj.has("delete")) {
+//    catch (Exception e2) {
```
Peilin-Yang: remove

src/main/java/io/anserini/document/TweetDocument.java
```diff
@@ -72,79 +67,115 @@ public TweetDocument() {
...
     try {
       timestamp_ms = (new SimpleDateFormat(DATE_FORMAT, Locale.ENGLISH)).parse(createdAt).getTime();
       epoch = timestamp_ms / 1000;
     } catch (ParseException e) {
-      //LOG.debug(e);
       timestamp_ms = -1L;
       epoch = -1L;
+      return false;
     }
+//    catch (NullPointerException e2) {
```
Peilin-Yang: remove

src/main/java/io/anserini/document/Status.java
```diff
@@ -0,0 +1,213 @@
...
+ */
+public class Status {
+
+  // Required fields
+  protected String created_at;
+  protected String id_str;
+  protected String text;
+  protected User user;
+
+  protected long retweet_count;
```
Peilin-Yang: Is `retweet_count` a required field?

src/main/java/io/anserini/document/TweetDocument.java
borislin: There are too many such errors due to missing required JSON fields

src/main/java/io/anserini/document/Status.java
borislin: Nope

src/main/java/io/anserini/document/Status.java
borislin: An overloaded getter method for Jackson deserialization

borislin:
@Peilin-Yang @lintool Regressions disks12, tweets2011 and tweets 2013 have all passed without any warnings and errors. #282 has also been solved in this PR. Do you still have more comments?

src/main/java/io/anserini/document/Status.java
Peilin-Yang: Then I think we should make it as OptionalLong, no?

borislin: It doesn't have to be `OptionalLong`. When the value is missing during deserialization, we will get `null` if we try to access it.

If we make this `OptionalLong`, then we should make all optional fields to have `Optional` type. Do you want this?

borislin:
@Peilin-Yang For MB11, there is no difference. For MB13, there are 25 lesser indexed docs after refactoring. Below is the current log (you can also get it from my personal folder):

```2018-07-07 00:05:29,873 INFO  [main] index.IndexCollection (IndexCollection.java:343) - # Final Counter Values
2018-07-07 00:05:29,875 INFO  [main] index.IndexCollection (IndexCollection.java:344) - indexed:      203,162,923
2018-07-07 00:05:29,875 INFO  [main] index.IndexCollection (IndexCollection.java:345) - empty:            143,890
2018-07-07 00:05:29,875 INFO  [main] index.IndexCollection (IndexCollection.java:346) - unindexed:     55,894,079
2018-07-07 00:05:29,875 INFO  [main] index.IndexCollection (IndexCollection.java:347) - unindexable:            0
2018-07-07 00:05:29,875 INFO  [main] index.IndexCollection (IndexCollection.java:348) - skipped:       55,747,435
2018-07-07 00:05:29,876 INFO  [main] index.IndexCollection (IndexCollection.java:349) - errors:             2,754
2018-07-07 00:05:29,945 INFO  [main] index.IndexCollection (IndexCollection.java:352) - Total 203,146,197 documents indexed in 04:45:18
```

FYI, this PR has also fixed a bug. We previously printed out # of skipped using # of errors data.

src/main/java/io/anserini/document/Status.java
Peilin-Yang: I just saw you made other fields below as Optional fields. 
Shouldn't we be consistent here?

Peilin-Yang:
Now I have problem logging to the server but I roughly remember the indexing time of MB13 with option `-uniqueDocid` on was under 3hrs .... 
Almost 5hrs seems to be high. When and where did you test it?

src/main/java/io/anserini/document/Status.java
borislin: Right, I'll make this change.

borislin:
@Peilin-Yang I didn't use that option and I run MB13 regression in the midnight of last Friday.

PR #298, opened 2018-07-05
Updated badges, wrote HISTORY

lintool:


HISTORY.md
```diff
@@ -0,0 +1,21 @@
+Anserini
```
Peilin-Yang: `docs/release-notes/release-notes-0.1.0.md` ?

lintool: You prefer each release have it's own separate release notes, as opposed to everything all gathered together in a file?

lintool: I suppose if we wanted to go this route, we can create a new "Release History" section in the README.

Peilin-Yang: We could have a release history link in README.md 
And I am fine with release history in a single file. But let's put it under `docs/`?

HISTORY.md
```diff
@@ -0,0 +1,21 @@
...
++ Robust05 experiments on the AQUAINT collection
++ CORE17 experiments on the New York Times collection
++ TREC Microblog 2011-2014 experiments
++ Experiments on Wt10g collection
++ Experiments on Gov2 collection
++ Experiments on ClueWeb09b collection
++ Experiments on ClueWeb12-B13 collection
++ Experiments on ClueWeb12 collection
+
+Working ranking models include QL, BM25, and RM3.
```
Peilin-Yang: I think it is also possible to list all the contributors (github usernames) for THIS release.
We can keep the list onward for the contributors just for a release? We can get the creators of PRs from the project page, right?

lintool:
@Peilin-Yang I thought about it and like your original suggestion of separate release notes better.

PR #303, opened 2018-07-06
add python interface integration test to travis

Peilin-Yang:
Travis will fail if the python interface is corrupted.

### DIFF
--- a/src/test/python/python_interface.py
+++ b/src/test/python/python_interface.py
@@ -1,7 +1,7 @@
 import sys
 import jnius_config
 
-jnius_config.set_classpath(sys.argv[1])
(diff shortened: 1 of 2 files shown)


src/test/python/python_interface.py
```diff
@@ -0,0 +1,7 @@
+import jnius_config
+jnius_config.set_classpath("target/anserini-0.1.1-SNAPSHOT-fatjar.jar")
```
tuzhucheng: Is there a way to do this without hardcoding the version number? We may need to manually bump this if we hardcode it.

tuzhucheng: One way I can think of is to use `os` to find a jar in `target` matching some regex and get the path of the jar.

PR #304, opened 2018-07-06
Refactoring NYT collection

lintool:
Cleaning up implementation of how NYT documents are parsed, along the lines of #254 

Passes regression test.

### DIFF
--- a/src/main/java/io/anserini/collection/NewYorkTimesCollection.java
+++ b/src/main/java/io/anserini/collection/NewYorkTimesCollection.java
@@ -23,11 +23,18 @@ import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/java/io/anserini/collection/NewYorkTimesCollection.java
```diff
@@ -31,35 +31,40 @@
...
     public boolean hasNext() {
-      return !atEOF;
+      return !docRead;
     }
 
     @Override
     public NewYorkTimesDocument next() {
-      NewYorkTimesDocument doc = new NewYorkTimesDocument(new File(fileName));
-      atEOF = true;
+      if (docRead) return null;
```
lintool: @Peilin-Yang would you prefer we return null here or throw a `NoSuchElementException`?

Peilin-Yang: `NoSuchElementException` is better.

lintool:
@Peilin-Yang Agreed

PR #305, opened 2018-07-06
axiom algorithm fix

Peilin-Yang:


src/main/java/io/anserini/rerank/lib/AxiomReranker.java
```diff
@@ -345,13 +335,27 @@ private ScoredDocuments processExternalContext(ScoredDocuments docs, RerankerCon
    */
   @VisibleForTesting
   private Map<String, Double> calTermScore(
```
lintool: rename method `compute` instead of `cal`?

lintool:
In your commit notes, please explain exactly what this PR fixes?

PR #308, opened 2018-07-07
Make outputQuery of reranking as an option

Peilin-Yang:
Make outputQuery of reranking as an option in order to suppress massive logging messages


src/main/java/io/anserini/rerank/lib/AxiomReranker.java
```diff
@@ -293,12 +299,12 @@ private ScoredDocuments processExternalContext(ScoredDocuments docs, RerankerCon
     for (int docid : docIds) {
       Terms terms = reader.getTermVector(docid, LuceneDocumentGenerator.FIELD_BODY);
       if (terms == null) {
-        LOG.info("Document vector not stored for docid: " + docid);
+        LOG.warn("Document vector not stored for docid: " + docid);
```
lintool: should we throw an exception here, since obviously we can't run this without doc vectors...? same with RM3?

Peilin-Yang: We've discussed this before, right?
Enabling the indexing option `-storeDocvectors` cannot guarantee that all the documents vectors are available... I do not know the exact reason though....

lintool: hrm. don't remember, but I trust you...

PR #310, opened 2018-07-07
Add -dumpAllDocids and -dumpRawDocs options to IndexUtils

Peilin-Yang:
For the context, I found axiomatic relevance feedback model does not scale well on CW12Full collection (I got OOM errors). 
It seems we can solve it by first dumping all docids in sorted order to disk (`-dumpAllDocids`) and then read them from file.
For `-dumpRawDocs`, it is useful if we would like to build the sub-index which includes a subset of the documents from an index. For example, this is essential to reproduce what I've done for TREC 2013 Web Track where I firstly built a small index by using the top 10,000 documents (per query) returned by LM model and then did all the following experiments using the small index. 

The added options are also generic and may be useful for other use cases


docs/index-utils.md
```diff
@@ -36,6 +36,13 @@ Converts a collection lookupDocid (in the raw corpus) to a Lucene internal looku
 Converts to a Lucene internal lookupDocid to a collection lookupDocid  (default: 0)
 
 
+```
+-dumpAllDocids outputCompressionFormat
+```
+Dumps all Docids in the index. For non-tweets collection the order is in ascending of String docid;
+For tweets collection the order is in descending of Long tweet id. Please provide the compression format for the output
+
```
lintool: non-tweets -> non-tweet
what are the options for the compression formats?

docs/index-utils.md
```diff
@@ -36,6 +36,13 @@ Converts a collection lookupDocid (in the raw corpus) to a Lucene internal looku
 Converts to a Lucene internal lookupDocid to a collection lookupDocid  (default: 0)
 
 
+```
+-dumpAllDocids outputCompressionFormat
+```
+Dumps all Docids in the index. For non-tweets collection the order is in ascending of String docid;
```
lintool: you have Docid here and docid below - I think we've used mostly docid (lowercase)?

docs/index-utils.md
```diff
@@ -44,6 +51,18 @@ Dumps raw document.
 _NOTICE:_ available only if the raw documents are stored with indexing option `-storeRawDocs`
 
 
+```
+-dumpRawDocs docidsInputPath
```
lintool: what about `-dumpRawDocsWithDocid` and `-dumpRawDocsWithoutDocid`?

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -54,9 +72,22 @@
     @Option(name = "-printDocvector", metaVar = "docid", usage = "prints the document vector of a document")
     String docvectorDocid;
 
+    @Option(name = "-dumpAllDocids", usage = "dumps all Docids in sorted order. For non-tweets collection the order is " +
```
lintool: So... I'm not sure we need to have the file above?
It's just a repeat of what's here, no?

Peilin-Yang: I think the `index-utils.md` is useful in the sense that we could have an entry in the `README.md` so that people know about this utility?

docs/index-utils.md
Peilin-Yang: sg

docs/index-utils.md
Peilin-Yang: The options are defined as enum `Compression`

docs/index-utils.md
Peilin-Yang: okay

docs/index-utils.md
Peilin-Yang: Or `-dumpRawDocsWithDocid` and `-dumpRawDocs`?
Anserini by default does not include docid in the raw doc

src/main/java/io/anserini/index/IndexUtils.java
lintool: Sure, but no need to repeat the options? This forces us to keep the docs in sync with the code... over time, it'll drift apart...

lintool:
One philosophical point worth discussion - do you want one mega index util that does everything? or smaller utils that each do an individual job?

I don't feel too strongly one way or another, but to be, dumping out raw docs seems like a different functionality than "point queries" into the index...

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -138,6 +183,51 @@ public void printDocumentVector(String docid) throws IOException, NotStoredExcep
...
+    } catch (IllegalStateException e) { // because this is tweets collection
+      scoreDocs = searcher.search(new FieldValueQuery(TweetGenerator.StatusField.ID_LONG.name), reader.maxDoc(),
+              BREAK_SCORE_TIES_BY_TWEETID).scoreDocs;
+    }
+
+    String basePath = directory.getDirectory().getFileName().toString() + ".allDocids";
+    OutputStream outStream = null;
+    String outputPath = "";
+    switch (compression) {
+      case RAW:
```
lintool: how about "NONE"? RAW might be confused for some compression scheme.

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -138,6 +183,51 @@ public void printDocumentVector(String docid) throws IOException, NotStoredExcep
     }
   }
 
+  public void getAllDocids(Compression compression) throws IOException {
+    Query q = new FieldValueQuery(LuceneDocumentGenerator.FIELD_ID);
+    IndexSearcher searcher = new IndexSearcher(reader);
+    ScoreDoc[] scoreDocs;
+    try {
+      scoreDocs = searcher.search(new FieldValueQuery(LuceneDocumentGenerator.FIELD_ID), reader.maxDoc(),
+              BREAK_SCORE_TIES_BY_DOCID).scoreDocs;
```
lintool: I think we want 4-space indent here?

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -138,6 +183,51 @@ public void printDocumentVector(String docid) throws IOException, NotStoredExcep
...
+    OutputStream outStream = null;
+    String outputPath = "";
+    switch (compression) {
+      case RAW:
+        outputPath = basePath+".txt";
+        outStream = Files.newOutputStream(Paths.get(outputPath));
+        break;
+      case GZ:
+        outputPath = basePath+".gz";
+        outStream = new GzipCompressorOutputStream(new BufferedOutputStream(
```
lintool: four-space indent, same below?

Peilin-Yang:
Sure. 

But I would support a unified place for all index utils:
Although the dumpRawDocs and point queries are different use cases they are all with index so to me they are not too drift away from each other...
Now we mainly have three interfaces: `IndexCollection`, `SearchCollection` and `IndexUtils`. I believe we will add more interfaces in the future. The current interfaces are already complicated (none of us has tried each single option). IMO, Reducing the number of interfaces would make users on-boarding easier.

lintool:
> For example, this is essential to reproduce what I've done for TREC 2013 Web Track where I firstly 
> built a small index by using the top 10,000 documents (per query) returned by LM model and then 
> did all the following experiments using the small index.

How exactly do you do this? You fetch top 10k hits using `SearchCollection`, then take those docids and run through this util?

src/main/java/io/anserini/index/IndexUtils.java
Peilin-Yang: You mean the multi-lines statement?
I just use my IntelliJ's default indent....and it seems that it doesn't allow me to have self-defined indent here.... (backspace will go to the line above, enter will do the exactly same here)

Peilin-Yang:
Yes. Actually there are two different ways to achieve this:

1. What you've mentioned. This requires that we already have the index with raw docs in it.
2. Or we can use `IndexCollection` with `-whitelist` option to build a fresh index from scratch.

src/main/java/io/anserini/index/IndexUtils.java
lintool: I believe it's called a "continuation indent" in IntelliJ... set to four spaces?

lintool:
Alternatively, we could add another option to `SearchCollection` to dump out the retrieved docs? But this would make SearchCollection even more complicated, I suppose...

Peilin-Yang:
yea, `SearchCollection` is already insanely complicated...

lintool:
Okay, let's not bloat `SearchCollection` even more then. 🚢

PR #311, opened 2018-07-08
WikipediaParagraph Collection

tuzhucheng:
Add a new Collection class to be used for indexing Wikipedia articles by paragraphs.


src/main/java/io/anserini/collection/WikipediaParagraphsCollection.java
```diff
@@ -0,0 +1,133 @@
...
+          .withLanguage(WikiLanguage.EN).withTitle(false)
+          .withFooter(false).build();
+
+      CONTAINS_PERIOD = Pattern.compile("\\.");
+      SINGLE_WORD_WITH_PERIOD = Pattern.compile("\\w+\\.");
+    }
+
+    @Override
+    public WikipediaParagraph next() {
+      if (bufferedParagraphs != null && currentParagraphIdx < bufferedParagraphs.size()) {
```
Peilin-Yang: I think this block can be put after the `bufferedParagraphs` is populated below?

src/main/java/io/anserini/collection/WikipediaParagraphsCollection.java
```diff
@@ -0,0 +1,133 @@
...
+
+          // Skip redirects
+          if (s.startsWith("#REDIRECT")) {
+            continue;
+          }
+
+          String[] parts = s.split("\\n+");
+          bufferedParagraphs = new ArrayList<>();
+          for (String p : parts) {
+            // Heuristic is that parts without periods ('.') or has one word followed by a period is a section header and should not be indexed.
```
Peilin-Yang: I think it is better to explicitly add comments about how a paragraph is identified?
My understanding is that, first a page is split into lines; for each line, it has to contain at least a dot `.` in order to be included in `bufferedParagraphs`. That means a line w/o dot won't be included in any paragraphs? This seems to be counter-intuitive...

src/test/java/io/anserini/document/WikipediaParagraphTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.apache.tools.ant.filters.StringInputStream;
+import org.junit.Before;
+import org.junit.Test;
+
+import io.anserini.collection.Collection;
+import io.anserini.collection.WikipediaParagraphsCollection;
+
+
+public class WikipediaParagraphTest extends DocumentTest<WikipediaParagraph> {
+    protected static Path tmpPath;
```
lintool: two-space indent.

lintool:
I agree with @Peilin-Yang - paragraph-based indexing should be a "primitive" indexing option. I'll start a separate issue to capture discussion.

lintool:
Let's have the discuss in #312

lintool:
There hasn't been any action on this issue for a while... closing for now.

PR #314, opened 2018-07-08
Add f2log ranking model

Peilin-Yang:
The F2Log model at current adopted version Lucene 6.3 is slow and problematic and I plan to submit a PR to Lucene 7.x 
For now, let us just create a temporary solution mainly for TREC 2018. 
After we migrate Lucene 7.x we will be good to use native Lucene's implementation.
The model implemented in this PR is essentially from BM25. BM25 has some final methods which we need to override that's why I cannot directly subclass BM25. 
Please note that I do not include test cases since Lucene's `BaseSimilarityTestCase` is not available outside Lucene repo and I largely believe the implementation is correct.

### DIFF
--- a/src/main/java/io/anserini/search/SearchArgs.java
+++ b/src/main/java/io/anserini/search/SearchArgs.java
@@ -83,6 +83,11 @@ public class SearchArgs {
   @Option(name = "-b", metaVar = "[value]", required = false, usage = "BM25 b parameter")
   public float b = 0.4f;
 
+  @Option(name = "-f2log", usage = "use F2Log scoring model")
+  public boolean f2log = false;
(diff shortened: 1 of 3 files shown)


lintool:
@Peilin-Yang breaks tests?

src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -83,6 +83,11 @@
   @Option(name = "-b", metaVar = "[value]", required = false, usage = "BM25 b parameter")
   public float b = 0.4f;
 
+  @Option(name = "-f2log", usage = "use F2Log scoring model")
+  public boolean f2log = false;
+  @Option(name = "-f2log.s", metaVar = "[value]", required = false, usage = "F2Log s parameter")
```
lintool: add newline

PR #315, opened 2018-07-08
Add search option `-axiom.docids` to axiomatic reranking

Peilin-Yang:
Add search option `-axiom.docids` to axiomatic reranking so that the algorithm can opt to read external docids from an input file. This will eliminate the need of searching and sorting all docids if we need deterministic reranking results and thus greatly improve the efficiency.

### DIFF
--- a/src/main/java/io/anserini/rerank/lib/AxiomReranker.java
+++ b/src/main/java/io/anserini/rerank/lib/AxiomReranker.java
@@ -98,15 +98,15 @@ public class AxiomReranker<T> implements Reranker<T> {
     this.R = args.axiom_r;
     this.N = args.axiom_n;
     this.beta = args.axiom_beta;
-    this.externalIndexPath = args.axiom_external_index;
+    this.externalIndexPath = args.axiom_index;
     this.outputQuery = args.axiom_outputQuery;
 
     if (this.deterministic && this.N > 1) {
       if (args.axiom_docids != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/anserini/rerank/lib/AxiomReranker.java
```diff
@@ -176,21 +191,46 @@ private ScoredDocuments searchTopDocs(Query query, RerankerContext<T> context) t
...
+    return IOUtils.readLines(bRdr);
+  }
+
   /**
    * If the result is deterministic we can cache all the docids. All queries can share this
    * cache.
    */
-  private ScoreDoc[] makeDocidsCache(SearchArgs args) throws IOException {
-    String index = args.axiom_external_index.trim().isEmpty() ? args.index : args.axiom_external_index.trim();
+  private ScoreDoc[] makeInternalDocidsCache(SearchArgs args) throws IOException {
```
lintool: I don't like `make` as a verb in a method name... maybe `build`?

src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -120,7 +124,7 @@
   public float axiom_beta = 0.4f;
 
   @Option(name = "-axiom.index", usage = "path to the external index for generating the reranking doucments pool")
-  public String axiom_external_index = "";
+  public String axiom_external_index = null;
```
lintool: make arg and variable name consistent? `index` vs. `external_index`

Issue #313, opened 2018-07-08
Reorganization of collections: use inner classes to associate related objects


lintool:
Currently, we have `FooCollection` and `FooDocument`, like `ClueWeb09Collection` and `ClueWeb09Document`. Since they are a tightly-bound semantic unit, one could reorganize as inner classes. So:

+ `ClueWeb09` would be the name of the collection (implicitly)
+ `ClueWeb09.Document` would be the name of the document from that class.

Same with all the other collections. Then we can reorganize all these bundles in `io.anserini.collection.lib`; the interface definitions go in `io.anserini.collection`.

This would also make Javadocs a lot cleaner.

The only major downside I can see is that our source code files will be much longer... but given that one navigates with an IDE these days, it's not that big a deal...?

Thoughts @Peilin-Yang?

Peilin-Yang:
I think it is reasonable to merge `Collection` and `Document`.
What do you think about another component `Iterator`? Basically, a `Collection` consists of `Iterator` and `Document`. The `Iterator` discovers documents (now we only deal with files but we can support streaming data too) and the `Document` parse the document into a structure or throw an error. 

Another thing is that we should unify or at least keep error handling consistent. Now when indexing a collection the error could be from `Document`; it could also be from `Collection`; it can yet be from `Generator`.... and I think we definitely need to fix this.

lintool:
@borislin Can you organize the twitter and WashingtonPost collections in the same way? I.e., nest as inner classes.

If I understand correctly, `Status` and `WashingtonPost` in `io.anserini.document` are helpers for Jackson parsing? If so, can we given them a name that makes these clear?

Put Washington Post and Twitter on separate PRs please.

borislin:
@lintool Jackson uses class definitions of `Status` and `WashingtonPost` to parse (deserialize & serialize) a tweet JSON and a WashingtonPost JSON data respectively. How about naming them `TweetObject` and `WashingtonPostObject`?

lintool:
@borislin Three more collections to go and we're done: Wikipedia, Car, and Json. Separate PR for each?

@tuzhucheng should review Wikipedia
@Peilin-Yang should review Json
@Victor0118 should review Car

lintool:
Yay, we're all done. Thanks @borislin !

PR #316, opened 2018-07-10
Organizing ClueWeb collection classes into inner classes

lintool:
Issue #313: Organized ClueWeb{09,12} collections - documents are now inner classes of the collection classes.


src/main/java/io/anserini/collection/ClueWeb09.java
```diff
@@ -0,0 +1,745 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Class representing an instance of the
+ * <a href="https://www.lemurproject.org/clueweb09.php/">ClueWeb09 collection</a>.
+ * This can be used to read the complete ClueWeb09 collection or the smaller ClueWeb09b subset.
+ */
+public class ClueWeb09 extends Collection {
```
Peilin-Yang: I assume we will be changing other collection names too?

lintool: Yes, if you like this, I will slowly change other collection names as we refactor. `ClueWeb09Collection` sounds redundant, but I'd keep `TrecCollection`?
Or would you rather everything be named `FooCollection`?

Peilin-Yang: Yea, I am thinking this change will break everyone's local scripts... which is bad...
I prefer FooCollection since it is pretty recognizable and keep it as is makes everyone happy

Issue #325, opened 2018-07-11
Error when dump raw docs on Robust04: requested size exceeds TarArchiveEntry set size


Kytabyte:
When I try to dump raw docs on Robust04 collection, I ran

```
sh target/appassembler/bin/IndexUtils -index lucene-index.robust04.pos+docvectors+rawdocs/ \
-dumpRawDocs lucene-index.robust04.pos+docvectors+rawdocs.allDocids.txt
```

I got
```
Exception in thread "main" java.io.IOException: request to write '24645' bytes exceeds size in header of '24627' bytes for entry 'FBIS3-20646'
	at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.write(TarArchiveOutputStream.java:382)
	at java.io.OutputStream.write(OutputStream.java:75)
	at io.anserini.index.IndexUtils.dumpRawDocuments(IndexUtils.java:361)
	at io.anserini.index.IndexUtils.main(IndexUtils.java:456
```

I'm having the same problem while I'm using the same idea to write code for dumping out document vectors. I think this is because it uses `String.length()` to set the initial size of a `TarArchiveEntry`, which may not match the size of some UTF-8 char when writing.

lintool:
@Kytabyte seems to be solved, so I'm closing...

Issue #318, opened 2018-07-10
Write utility to dump out document vectors


lintool:
To replicate Gord's CORE17 runs, we need to dump out document vectors for the qrels (for simplicity, he weights by tf.idf). This should go into `IndexUtils`.

Kytabyte:
Anserini changes soooooo fast...

My proposal is to use the dump raw docs idea, replicating `-dumpRawDoc`, `-dumpRawDocs`, `-dumpRawDocsWithDocid` args with `-dumpTfidf`, `-dumpTfidfs`, and `dumpTfidfsWithDocid` with the same functionality for each. Sounds good?

lintool:
How about `-dumpDocVectors` and `-dumpDocVectors.weights` (`tf.idf` is the only option supported right now, but we could add `bm25` later)?

I don't think we need `WithDocid` because it wouldn't make sense to dump the doc vectors without docids, right?

Kytabyte:
> How about -dumpDocVectors and -dumpDocVectors.weights (tf.idf is the only option supported right now, but we could add bm25 later)?

That's good.

> I don't think we need WithDocid because it wouldn't make sense to dump the doc vectors without docids, right?

Yes, but actually I'm not quite sure the use case to add a `WithDocid` option in dumping raw docs, since in `dumpRawDocuments` each doc will be written into a single tar file named as `docid`. Users can get the docid by name in anyways...

So basically I agree that we can make it only one option, just with tag or without it

Peilin-Yang:
> Yes, but actually I'm not quite sure the use case to add a WithDocid option in dumping raw docs, since in dumpRawDocuments each doc will be written into a single tar file named as docid. Users can get the docid by name anyway...

@Kytabyte Anserini by default does store `<DOCNO>docid</DOCNO>` in the front of the raw documents (let's assume it is a TREC document). By adding `<DOCNO>docid</DOCNO>` we are able to directly index the dumped documents using Anserini or Indri. When the dumped archive is huge this is very handy.

Kytabyte:
@Peilin-Yang Thank you for the explanation!

What about the `-dumpRawDoc` (dump one doc given docid) and `-dumpRawDocs` (dump all docs given a file of docid list) options? Do we need both options for dump doc vectors as well?

lintool:
Sure, have both options - dump one doc vector and dump entire file of doc vectors

Kytabyte:
[...]

PR #326, opened 2018-07-11
Dump document vectors in IndexUtil (#318)

Kytabyte:
1. I changed `printDocvector` argument into `dumpDocVector` to keep the consistency.
2. This structure is the same as `DumpRawDocs`, where the input is a file with all required docids and output is a tar.gz file containing the doc vector for one document per file.
3. the usage is `sh target/appassembler/bin/IndexUtils -index /path/to/index -dumpDocVectors /path/to/docid -docVectorWeight tf.idf`, where the `-docVectorWeight` is optional and can be easily expanded to other weight method in the future.


src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -105,6 +113,12 @@ public NotStoredException(String message) {
     }
   }
 
+  public class InvalidDocVectorWeightException extends Exception {
```
lintool: no need for this - see below.

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -183,6 +197,104 @@ public void printDocumentVector(String docid) throws IOException, NotStoredExcep
...
+        outLength += (entry.getKey().getBytes(StandardCharsets.UTF_8).length
+                + entry.getValue().getBytes(StandardCharsets.UTF_8).length
+                + " \n".length());
+      }
+
+      TarArchiveEntry tarEntry = new TarArchiveEntry(new File(docid));
+      tarEntry.setSize(outLength + String.format("<DOCNO>%s</DOCNO>\n", docid).length());
+      tOut.putArchiveEntry(tarEntry);
+      tOut.write(String.format("<DOCNO>%s</DOCNO>\n", docid).getBytes());
+      for (Map.Entry<String, String> entry: docVectors.entrySet()) {
```
lintool: Do you know about this?
```
Joiner.MapJoiner mapJoiner = Joiner.on(",").withKeyValueSeparator("=");
System.out.println(mapJoiner.join(map));
```
see: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Joiner.MapJoiner.html

lintool:
High-level question - do we really need to compress the doc vectors?

@Peilin-Yang is that your feature request?

Seems a bit like overkill?

Kytabyte:
NYT collection contains 1,800,000+ documents, and the dumped uncompressed TF-IDF file is around 6 GB. I think for even larger collection, like ClueWeb12, the compressed file would be useful?

lintool:
@Kytabyte I see. Is there a use case where we would dump the entire collection? I thought we'd only do this selectively, e.g., qrels...

src/main/java/io/anserini/index/IndexUtils.java
Peilin-Yang: How about 
```
String s = ranking.entrySet().stream().map(e -> e.getKey()+" "+e.getValue()).collect(joining("\n"));
```
?
Guava is less and less popular after Java 8 and Scala....

lintool: Oh really? Should we kill it as a dependency then?

It was very much in use when I was at Twitter. Every method started with Guava precondition checks...

Peilin-Yang:
For raw documents we can benefit from tar-balling the outputs.
I am not sure about the exact use case here but enabling compression makes it more scalable.

src/main/java/io/anserini/index/IndexUtils.java
Peilin-Yang: I think we could file an issue for it.
In Java 8 and Scala/Rust, the introduction of `Option` forces the programmers to deal with `Null` which eliminates most of the precondition checks

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -183,6 +194,98 @@ public void printDocumentVector(String docid) throws IOException, NotStoredExcep
...
+
+    InputStream in = getReadFileStream(reqDocidsPath);
+    BufferedReader bRdr = new BufferedReader(new InputStreamReader(in));
+    FileOutputStream fOut = new FileOutputStream(new File(outFileName));
+    BufferedOutputStream bOut = new BufferedOutputStream(fOut);
+    GzipCompressorOutputStream gzOut = new GzipCompressorOutputStream(bOut);
+    TarArchiveOutputStream tOut = new TarArchiveOutputStream(gzOut);
+
+    Map<Term, Integer> docFreqMap = new HashMap<>();
+    Map<String, String> docVectors;
```
Peilin-Yang: The definition can be moved to inside the loop?
We will introduce GC cycles anyway...

Kytabyte:
> Could you please also fix #325 in this PR?
And also make the loggings consistent for dumpDocVectors and dumpRawDocs?

will do!

src/main/java/io/anserini/index/IndexUtils.java
Kytabyte: The `docFreqMap` is a global variable for the entire loops. It stores the document frequency of every `Term` when it occurs and we query it from indexing, thus we can read it from memory next time we see this term. It'll be much slower if we query DF every time...

Peilin-Yang: Yes, what I meant is `docVectors`.

Kytabyte: Oh, sure, will do.

Peilin-Yang:
@lintool do you have any further comments?

PR #329, opened 2018-07-12
Remove all classes related to Freebase

lintool:
This is legacy code, originally for our NAACL 2018 short paper on KBQA. The plan was to do entity linking using Lucene, but Salman eventually implemented the entity linking algorithm directly in Python. This code is not being used and thus not maintained.


pom.xml
```diff
@@ -201,7 +201,7 @@
       <artifactId>lucene-benchmark</artifactId>
       <version>${LUCENE_VERSION}</version>
     </dependency>
-    <dependency> <!-- needed only for org.apache.lucene.search.CheckHits: candidate for cleanup -->
+    <dependency>
```
lintool: Actually, as it turns out, we have test cases that also depend on this artifact.

lintool: done!

PR #330, opened 2018-07-12
Refactor WashingtonPostCollection class to have inner classes for Document

borislin:
Issue #313: Refactored `WashingtonPostCollection` - `Document` is now an inner class of the collection class.


src/main/java/io/anserini/collection/WashingtonPostCollection.java
```diff
@@ -53,4 +58,137 @@ public FileSegment(Path path) throws IOException {
...
+        //  published date. So we just simply log a warning and return null
+        //  here in case future data may bring up this issue
+        LOG.warn("No unique ID or published date for this record, ignored...");
+        return null;
+      }
+
+      id = recordObj.id();
+      published_date = recordObj.published_date();
+
+      if (recordObj.contents != null && recordObj.contents.isPresent()) {
```
Peilin-Yang: This block seems to be to verbose, can we make it simpler?

borislin: When a field such as `contents` is missing in the JSON string, `recordObj.contents` will be equal to `null` after deserialization. When the JSON string contains `{"contents" : null}`, `recordObj.contents` will be equal to `Optional.empty()` after deserialization.

That's why I've used `if (recordObj.contents != null && recordObj.contents.isPresent())`.

borislin: That makes sense. I'll do it now.

borislin:
@Peilin-Yang are you happy with this PR now?

PR #333, opened 2018-07-12
Remove all dependencies on Guava

Peilin-Yang:


src/main/java/io/anserini/ltr/TwitterFeatureExtractor.java
```diff
@@ -102,10 +106,12 @@ protected Analyzer getAnalyzer() {
 
   @Override
   protected Set<String> getFieldsToLoad() {
-    return Sets.newHashSet(getIdField(), getTermVectorField(),
-            StatusField.FOLLOWERS_COUNT.name,
-            StatusField.FRIENDS_COUNT.name,
-            StatusField.IN_REPLY_TO_STATUS_ID.name);
+    return Stream.of(
```
lintool: Is this more clear than something like?

> `Set<String> h = new HashSet<>(Arrays.asList("a", "b"));`

I actually like above better...

src/main/java/io/anserini/ltr/feature/FeatureExtractor.java
```diff
@@ -16,28 +17,29 @@
...
-          .put("UniqueQueryTerms", UniqueTermCount.class)
-          .put("UnorderedSequentialPairs", UnorderedSequentialPairsFeatureExtractor.class)
-          .put("OrderedSequentialPairs", OrderedSequentialPairsFeatureExtractor.class)
-          .put("TwitterHashtagCount", HashtagCount.class)
-          .put("IsTweetReply", IsTweetReply.class)
-          .put("TwitterLinkCount", LinkCount.class)
-          .put("TwitterFollowerCount", TwitterFollowerCount.class)
-          .put("TwitterFriendCount", TwitterFriendCount.class).build();
+  final Map<String, Class<?>> EXTRACTOR_MAP = new HashMap<String, Class<?>>() {{
+    put("AvgICTF", AvgICTFFeatureExtractor.class);
```
lintool: Can use HashMap<>()?

src/main/java/io/anserini/rerank/lib/AxiomReranker.java
```diff
@@ -117,7 +114,7 @@ public AxiomReranker(String field, SearchArgs args) throws IOException {
 
   @Override
   public ScoredDocuments rerank(ScoredDocuments docs, RerankerContext<T> context) {
-    Preconditions.checkState(docs.documents.length == docs.scores.length);
+    assert (docs.documents.length == docs.scores.length);
```
lintool: no space, right? below also.

src/main/java/io/anserini/ltr/feature/FeatureExtractor.java
Peilin-Yang: Cannot use diamond constructor with anonymous inner class....

PR #334, opened 2018-07-12
Refactor TweetCollection class to have inner classes for Document

borislin:
Issue #313: Refactored TweetCollection - Document is now an inner class of the collection class.


src/main/java/io/anserini/document/Status.java
```diff
@@ -1,201 +1,201 @@
...
-
-  @JsonGetter("delete")
-  public Optional<Delete> delete() {
-    return delete;
-  }
-
-  @JsonGetter("coordinates")
-  public Optional<Coordinates> coordinates() { return coordinates; }
-}
+//package io.anserini.document;
```
lintool: I think you actually want to remove the file...

src/main/java/io/anserini/collection/TweetCollection.java
```diff
@@ -67,4 +76,444 @@ public FileSegment createFileSegment(Path p) throws IOException {
...
+    public String content() {
+      return text;
+    }
+
+    @Override
+    public boolean indexable() {
+      return true;
+    }
+
+    public long getIdLong() { return idLong; }
```
lintool: make formatting consistent.

src/main/java/io/anserini/collection/TweetCollection.java
```diff
@@ -67,4 +76,444 @@ public FileSegment createFileSegment(Path p) throws IOException {
...
+    public String getScreenname() {
+      return screenname;
+    }
+    public String getCreatedAt() {
+      return createdAt;
+    }
+    public String getText() {
```
lintool: add newline between methods

src/main/java/io/anserini/document/TweetDocument.java
```diff
@@ -1,273 +1,273 @@
...
-    return retweetStatusId;
-  }
-  public OptionalLong getRetweetedUserId() {
-    return retweetUserId;
-  }
-  public OptionalLong getRetweetCount() {
-    return retweetCount;
-  }
-}
+///**
```
lintool: same, actually delete file.

src/test/java/io/anserini/document/TweetDocumentTest.java
```diff
@@ -18,16 +18,17 @@
...
+import io.anserini.collection.TweetCollection;
 import org.junit.Before;
 import org.junit.Test;
 
 
-public class TweetDocumentTest extends DocumentTest<TweetDocument> {
+public class TweetDocumentTest extends DocumentTest<TweetCollection.Document> {
 
   @Before
   public void setUP() throws Exception {
```
lintool: `setUp`

src/main/java/io/anserini/collection/TweetCollection.java
borislin: Should we use snake case for our variables across all files? Only 'TweetCollection.Document.TweetObject` uses snake case because the field name has to match exactly in the JSON string when parsing using Jackson.

The other place that uses snake case is `WashingtonPostCollection.Document.WashingtonPostObject`. I believe we've used camel case throughout all other places. That's what I'm currently doing.

lintool: Oh what I mean above was the braces on a separate line:
```
public long getIdLong() {
  return idLong;
}
```
But I think everything should be `CamelCase` except for the Jackson stuff - leave in snake?

borislin: Nvm, I've figured out a way to use camel case in all cases including in `TweetCollection.Document.TweetObject` and `WashingtonPostCollection.Document.WashingtonPostObject`.

PR #337, opened 2018-07-13
Add ndeval and related qrels for web track 2013-2014

Peilin-Yang:
`ndeval` is another evaluation metric for web track 13-14 and we should include it.


lintool:
I don't think we want to check in the binaries? `eval/ndeval/ndeval`

Can we add to the README the original source of the program - also for `trec_eval`.

PR #332, opened 2018-07-12
Patch for Issue #331: Collection class needs more direct connection with document type

lintool:
See Issue #331 

A collection is now:

```
public class ClueWeb09Collection extends DocumentCollection
    implements FileSegmentProvider<ClueWeb09Collection.Document>
```
`DocumentCollection` now only takes care of discovering files. The `FileSegmentProvider` explicitly connects the collection to the type of `SourceDocument` that it provides.

We don't need `Gov2Collection` and `WtCollection` anymore... `TrecwebCollection` suffices.

PR #338, opened 2018-07-13
Refactor JsonCollection class to have inner classes for Document

borislin:
Issue #313: Refactored JsonCollection - Document is now an inner class of the collection class.


borislin:
@Peilin-Yang I'll fix those coding warnings in a separate PR...let's unblock #332 first

Issue #321, opened 2018-07-11
Refactor NYT collection to directly index tgz files


lintool:
Currently, we can only index the NYT collection after decompressing it first. We should be able to use TarArchiveInputStream in Apache Commons Compress to index the tgz files directly.

For example, something like: https://stackoverflow.com/questions/25749550/read-tar-gz-in-java-with-commons-compression

Kytabyte:
Do we make it only read from the tgz files, or we should keep both options?

lintool:
Fixed with #340

PR #340, opened 2018-07-13
Index NYT Collection for tgz files (#321)

Kytabyte:
I rewrite some NYT API to make it read files from `BufferedReader` because `TarArchiveInputStream` is already a stream.

### DIFF
--- a/src/main/java/io/anserini/collection/NewYorkTimesCollection.java
+++ b/src/main/java/io/anserini/collection/NewYorkTimesCollection.java
@@ -17,6 +17,9 @@
 package io.anserini.collection;
 
 import io.anserini.document.SourceDocument;
+import org.apache.commons.compress.archivers.ArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
+import org.apache.commons.compress.compressors.gzip.GzipCompressorInputStream;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.w3c.dom.NamedNodeMap;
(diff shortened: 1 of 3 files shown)


lintool:
Did you run the regression script to verify it works? Change the path to the compressed version on tuna?

Kytabyte:
I run the command in docs and the results matched. I'll change the script, verify it, and push the changed script in this PR.

src/main/java/io/anserini/collection/NewYorkTimesCollection.java
```diff
@@ -47,25 +50,22 @@
...
+ * {@code FileSegment} can take either a {@code tgz} file or a {@code xml} file.
+ * If the {@code FileSegment} contains is an XML file, it contains only a single document.
  */
 public class NewYorkTimesCollection extends Collection<NewYorkTimesCollection.Document> {
   private static final Logger LOG = LogManager.getLogger(NewYorkTimesCollection.class);
 
   @Override
   public List<Path> getFileSegmentPaths() {
-    Set<String> allowedFileSuffix = new HashSet<>(Arrays.asList(".xml"));
+    Set<String> allowedFileSuffix = new HashSet<>(Arrays.asList(".xml", ".tgz"));
```
lintool: I think we want to remove `".xml"` here, right?

lintool:
One final nit. Please fix and I'll merge.

src/main/java/io/anserini/collection/NewYorkTimesCollection.java
Kytabyte: My current design actually supports both `tgz` and `xml` files to be indexed, just in case.

PR #342, opened 2018-07-15
delete no longer used docs

Peilin-Yang:


lintool:
@Peilin-Yang can we have the renaming of `qrels.web.201-250.ndeval.txt` etc. in a separate PR? I'd like to have each commit be semantically coherent.

PR #347, opened 2018-07-16
Examples test case to print out BM25 weights for each document vector

lintool:
Kinda roundabout: but the way to compute the BM25 score is to issue a query with the exact docid and the term in question, and look at the retrieval score.

### DIFF
--- a/src/test/java/io/anserini/integration/IndexerTest.java
+++ b/src/test/java/io/anserini/integration/IndexerTest.java
@@ -16,13 +16,14 @@
 
 package io.anserini.integration;
 
+import io.anserini.util.AnalyzerUtils;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.en.EnglishAnalyzer;
 import org.apache.lucene.codecs.FieldsProducer;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
 import org.apache.lucene.document.FieldType;
-import org.apache.lucene.document.TextField;
+import org.apache.lucene.document.StringField;
 import org.apache.lucene.index.CodecReader;
 import org.apache.lucene.index.DirectoryReader;
 import org.apache.lucene.index.FilterCodecReader;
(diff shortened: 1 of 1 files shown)


src/test/java/io/anserini/integration/IndexerTest.java
```diff
@@ -149,6 +159,70 @@ public void testReadingPostings() throws Exception {
...
+      String docid = reader.document(i).getField("docid").stringValue();
+      System.out.println(reader.document(i));
+      System.out.println(i+ ": " + docid);
+      Terms terms = reader.getTermVector(i, "text");
+      TermsEnum te = terms.iterator();
+
+      // For this document, iterate through the terms.
+      while (te.next() != null) {
+        String term = new Term("text", te.term()).bytes().utf8ToString();
+        String stemmed = AnalyzerUtils.tokenize(new EnglishAnalyzer(), term).get(0);
```
Kytabyte: This is redundant. The term stored in the Lucene index should already be tokenized and stemmed.

src/test/java/io/anserini/integration/IndexerTest.java
```diff
@@ -149,6 +159,70 @@ public void testReadingPostings() throws Exception {
...
+        Query filterQuery = new TermQuery(new Term("docid", docid)); // the docid
+        Query termQuery =  new TermQuery(new Term("text", term));    // the term
+        BooleanQuery.Builder builder = new BooleanQuery.Builder();   // must have both
+        builder.add(filterQuery, BooleanClause.Occur.MUST);
+        builder.add(termQuery, BooleanClause.Occur.MUST);
+        Query finalQuery = builder.build();
+        TopDocs rs = searcher.search(finalQuery, 1);                 // issue the query
+
+        // The BM25 weight is the maxScore
+        System.out.println(stemmed + " " + tf + " " + rs.getMaxScore());
```
Kytabyte: print `term` instead

lintool:
@Kytabyte addressed CR!

PR #351, opened 2018-07-19
Fix Trec Topic Reader to read TREC Core Track 2018 topics

Peilin-Yang:


Peilin-Yang:
Originally I also included the topics file. But later on removed it cause probably we are not allowed to put it publicly

Kytabyte:
@Peilin-Yang just saw you had a core18 topic file in this PR but then removed it... Do we already have the topics?

PR #359, opened 2018-07-22
Implement Sequential Dependency Model query constructor

Peilin-Yang:
It turns out that we can combine different query semantics from Lucene to construct SDM query.
For example, for ordered window and unordered window query, we can use [SpanQuery](https://lucene.apache.org/core/6_3_0/core/org/apache/lucene/search/spans/SpanQuery.html).
We then use `BooleanQuery` together with `BoostQuery` to construct the complete query. 

Please see the unit test for examples.

Test on Gov2 701-750: MAP - 0.2922 (0.2832 was reported in the original paper)


lintool:
Would it make sense to have a generic `QueryBuilder` API instead of these static methods?

Peilin-Yang:
You mean subclass `Query` and create an inner `Builder` class?

Peilin-Yang:
After thinking about it and looking at the `Query` class code, I think we do not necessary need SDM as a query class. 
The `Query` class and its subclasses mainly implements primitive query functions and SDM query can be constructed by leveraging the existing query semantics. 
(From the discussion in Slack channel: Jimmy mentioned to also make query expansion as subclasses of `Query`. I don't think this is suitable too. `Query` mainly deal with raw text queries w/o interacting with other components, e.g. index but query expansion involves reading indexing which is complicated to reason about)

I think we could make a separated `Query` package under `Search` and wrap this up?

lintool:
sg

src/main/java/io/anserini/search/query/QueryBase.java
```diff
@@ -1,23 +1,35 @@
...
 import org.apache.lucene.search.Query;
-import org.apache.lucene.search.TermQuery;
 
 import java.io.IOException;
 import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.List;
 
-public class AnalyzerUtils {
+public abstract class QueryBase {
```
lintool: I'd still like to keep `AnalyzerUtil` for the `tokenize` method - which is used beyond just query formulation.

lintool:
How about `QueryGenerator` as an abstract class? I don't quite like how `QueryBase` sounds.
Then we can have `{X,Y,Z}QueryGenerator`.
How about folding the field and the analyzer in the constructor? I think that'd make the API cleaner?

src/main/java/io/anserini/search/query/BagOfTermsQueryGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.lucene.search.BooleanQuery;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.TermQuery;
+
+import java.util.List;
+
+/*
+ * Bag of Terms query builder
+ */
+public class BagOfTermsQueryGenerator extends QueryGenerator {
```
lintool: Can we rename it `BagOfWords` since that's commonly known?

src/main/java/io/anserini/search/query/TermDependencyQueryGenerator.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.lucene.search.spans.SpanNearQuery;
+import org.apache.lucene.search.spans.SpanQuery;
+import org.apache.lucene.search.spans.SpanTermQuery;
+
+import java.util.List;
+
+/* Build the Term Dependency query. See:
+ * D. Metzler and W. B. Croft. A markov random field model for term dependencies. In SIGIR ’05.
+ */
+public class TermDependencyQueryGenerator extends QueryGenerator {
```
lintool: "TermDependency" is vague. How about SdmQueryGenerator to make consistent with test case below?

lintool:
@Peilin-Yang I know we were going to defer this until v0.3.0, but should we just move this up to v0.2.0 and get it merged in?

We'll need regressions in a separate PR?

Peilin-Yang:
Actually, we do not need to move this to 2 in order to get it merged....
We just merge it and it will be marked as done in the project page.

lintool:
Sure, either way is fine with me.
Fix conflicts and I'll do a CR now?

src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -89,6 +89,18 @@
   @Option(name = "-f2log.s", metaVar = "[value]", required = false, usage = "F2Log s parameter")
   public float f2log_s = 0.5f;
 
+  @Option(name = "-sdm", usage = "boolean switch to use Sequential Dependence Model query")
+  public boolean sdm = false;
+
+  @Option(name = "-sdm.tw", metaVar = "[value]", usage = "term weight in sdm")
+  public float sdm_tw = 0.85f;
```
lintool: "term weight in sdm" -> "SRM term weight"?
And below.

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -84,6 +86,12 @@
   private final boolean isRerank;
   private final RerankerCascade cascade;
 
+  enum QueryConstructor {
+    BagOfTerms,
+    SequentialDependenceModel
+  }
+  private final QueryConstructor qc;
```
lintool: add empty line?

src/test/java/io/anserini/search/SdmQueryTest.java
```diff
@@ -0,0 +1,171 @@
...
+    String sdmQueryStr = "fox information river";
+    Query sdmQuery1 = new SdmQueryGenerator(1.0f, 0.0f, 0.0f).buildQuery(field, analyzer, sdmQueryStr);
+    assertEquals(sdmQuery1.toString(), "(text:fox text:inform text:river)^1.0 " +
+            "(spanNear([text:fox, text:inform], 1, true) spanNear([text:inform, text:river], 1, true))^0.0 " +
+            "(spanNear([text:fox, text:inform], 8, false) spanNear([text:inform, text:river], 8, false))^0.0");
+    TopDocs rs1 = searcher.search(sdmQuery1, 1);
+    Query termQuery = new BagOfWordsQueryGenerator().buildQuery(field, analyzer, sdmQueryStr);
+    TopDocs rsTerm = searcher.search(termQuery, 1);
+    assertEquals(rs1.scoreDocs[0].score, rsTerm.scoreDocs[0].score, 1e-6f);
+
```
lintool: kill extra empty line?

src/test/java/io/anserini/search/SdmQueryTest.java
```diff
@@ -0,0 +1,171 @@
...
+    assertEquals(rs.scoreDocs.length, 0);
+
+    q = new SpanNearQuery(new SpanQuery[]{t2, t1}, 8, false);
+    rs = searcher.search(q, 1);
+    assertEquals(rs.scoreDocs.length, 1);
+
+    q = new SpanNearQuery(new SpanQuery[]{t2, t1}, 16, false);
+    rs = searcher.search(q, 1);
+    assertEquals(rs.scoreDocs.length, 1);
+
```
lintool: kill extra empty line?

lintool:
@Peilin-Yang you're changing the BoW code path, right? So please run all regressions before merging?

src/main/java/io/anserini/search/SearchArgs.java
Peilin-Yang: SDM?

Issue #366, opened 2018-07-27
Separate FileSegment interface and AbstractFileSegment abstract class?


lintool:
Okay, this might be a bit pedantic, but I think `FileSegment` should be an interface with `AbstractFileSegment` as the base class implementing that interface.

From a hygiene point of view, separating interfaces from impl is "a good thing".

From a practical perspective, our docs say something like:
> A static collection of documents, comprised of one or more {@link AbstractFileSegment}s.

Which is technically not true and sounds a bit odd. But saying *subclasses* or *concrete implementations* of `AbstractFileSegment` everywhere sounds too awkward. 

@borislin @Peilin-Yang thoughts?

I can work on it.

borislin:
I'm a bit confused.

Right now, we have our `FileSegment` in each collection, which extends to `AbstractFileSegment` as the base class.

Do you mean we should create an interface and let our `FileSegment` in each collection to implement that interface?

lintool:
No, just that we should have `interface FileSegment` and `abstract class AbstractFileSegment implements FileSegment`.

borislin:
But right now, we already have `abstract class AbstractFileSegment<T extends SourceDocument> implements Iterator<T>, Closeable`. If we create another interface `FileSegment`, what does this interface contain?

lintool:
So, it would be:

> interface FileSegment extends Iterator

and

> abstract class AbstractFileSegment implements FileSegment, Closeable

Something like that?

borislin:
Sure, I'm okay with this.

lintool:
Fixed in #366 - see CR for additional discussion on class names we settled on.

PR #367, opened 2018-07-27
Patch for #366: Separate FileSegment interface from AbstractFileSegment impl

lintool:
Patch for issue #366 
Also cleaned up Javadoc. All docs in collection package no read consistently.

### DIFF
--- a/src/main/java/io/anserini/collection/AbstractFileSegment.java
+++ b/src/main/java/io/anserini/collection/AbstractFileSegment.java
@@ -16,23 +16,18 @@
 
 package io.anserini.collection;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-
 import java.io.BufferedReader;
-import java.io.Closeable;
 import java.io.IOException;
 import java.nio.file.Path;
-import java.util.Iterator;
 import java.util.NoSuchElementException;
 
 /**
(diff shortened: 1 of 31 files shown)


lintool:
@Peilin-Yang yes I agree it's a bit confusing but not that bad...

+ A user is unlikely going to be using more than one collection at once, so I think it's okay that all the collection inner classes have the same name.
+ From the perspective of the Indexer, it only needs to care about the interface.

I don't like `FileSegmentImpl` because then we should have `DocumentImpl` to be consistent, which is kinda ugly.

How's this?

+ `AbstractFileSegment` -> `BaseFileSegment` per your suggestion.
+ `Foo.FileSegment` -> `Foo.Segment`: since they are under the collection class, it's clear that they are file segments. In principle we could have `MemorySegment` or even `StreamSegment` later?

So `Segment` will always refer to concrete impl in the context of various collections, and `FileSegment` will unambiguously refer to the abstraction?

Peilin-Yang:
I feel it should be the other way around - `Segment` as the name of the interface and `FileSegment` inside each collection. `MemorySegment` and `StreamSegment` are other types of `Segment`.

lintool:
Okay, `Segment` interface, `BaseFileSegment`, and `FileSegment` inner classes?

Peilin-Yang:
LGTM

Issue #353, opened 2018-07-20
"Mapper" over document collection


lintool:
As a way to demonstrate #254 (SourceDocument refactoring to get rid of readNextRecord), we could build a generic mapper framework -  an interface that lets us map over an entire collection in parallel. Design would be similar to how the indexer currently works - parallelism at the level of file segments, using a work pool.

The simplest implement would be a counter util to count the number of documents in a collection - this would help with issue #317 (Clean up semantics of index counters).

lintool:
With #352 and #254 done, @borislin can turn attention to this now?

Once this has been implemented, we can sanity check against counts here for ClueWeb12:
https://lemurproject.org/clueweb12/specs.php

And similar ground-truth counts for other collections.

The reason I want this broken out separately is that we can isolate reading records from other issues that crop up during indexing (e.g., parse errors, empty docs, etc.).

borislin:
@lintool I think we still need to parse each document, right? For instance, for each tweet document, we need to update `counter.errors` for tweet documents that contain the `delete` field.

lintool:
Well, deletes aren't errors though. By error I mean something like invalid JSON? Yes, there are going to be invalid JSON... but for ClueWeb{09,12}, we should be able to match exactly the record count in the websites above.

borislin:
#370 added this generic mapper framework

Issue #317, opened 2018-07-10
Clean up semantics of index counters


lintool:
Currently, the semantics of index counters are a mess. We use `null` to signal end of reading from a `FileSegment`, but `null` can also mean error processing a record. We log counters as multiple points in the code path, and so the counter values are partially overlapping...

Let's clean up.

lintool:
As noted by @borislin, this is related to #254.

borislin:
@lintool @Peilin-Yang 

I propose the following `Counters` class (`unindexed` removed):

```
public final class Counters {
    /**
     * Counter for successfully indexed documents.
     */
    public AtomicLong indexed = new AtomicLong();

    /**
     * Counter for empty documents that are not indexed. Empty documents are not necessary errors;
     * it could be the case, for example, that a document is comprised solely of stopwords.
     */
    public AtomicLong empty = new AtomicLong();

    /**
     * Counter for unindexable documents. These are cases where {@link SourceDocument#indexable()}
     * returns false.
     */
    public AtomicLong unindexable = new AtomicLong();

    /**
     * Counter for skipped documents. These are cases documents are skipped as part of normal
     * processing logic, e.g., using a whitelist, not indexing retweets or deleted tweets.
     */
    public AtomicLong skipped = new AtomicLong();

    /**
     * Counter for unexpected errors.
     */
    public AtomicLong errors = new AtomicLong();
  }
```

num of unindexed docs = num of `empty` + num of `unindexable` + num of `errors` + num of `skipped`
So we don't need the `unindexed` counter.

Sg?

Btw, I've noticed that we have two places that have deleted tweets. One is the `-tweet.deletedIdsFile` option passed into `IndexCollection` and the other one is when the `delete` field is detected in the JSON string when parsing each tweet document. I believe we may have counter overlapping issue here, right?

lintool:
@borislin overall, sg.

I'll let @Peilin-Yang clarify the semantics of Twitter deletes. 

> the other one is when the delete field is detected in the JSON string when parsing each tweet document.

Do you mean that the JSON of the tweet marks it as deleted? Can you show an example?

borislin:
@lintool Yes, I mean the JSON of the tweet marks it as deleted.

Examples:

https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/collection/TweetCollection.java#L118

https://github.com/castorini/Anserini/blob/master/src/test/resources/sample_docs/tweets/tweets1#L3

[...]

PR #369, opened 2018-07-28
Fix exception handling and counter update

borislin:
Fixed exception handling when parsing each document and counter update

### DIFF
--- a/src/main/java/io/anserini/index/IndexCollection.java
+++ b/src/main/java/io/anserini/index/IndexCollection.java
@@ -183,16 +183,23 @@ public final class IndexCollection {
 
         int cnt = 0;
         AbstractFileSegment iter = ((FileSegmentProvider) collection).createFileSegment(inputFile);
-        while (iter.hasNext()) {
-          SourceDocument d;
+        while (true) {
+          boolean hasNext = false;
           try {
-            d = iter.next();
+            hasNext = iter.hasNext();
           } catch (NoSuchElementException e1) {
-            continue;
-          } catch (Exception e2) { // TODO: update related counters (#317)
+            break;
+          } catch (Exception e2) {
             LOG.warn("Exception when parsing document: ", e2);
+            counters.errors.incrementAndGet();
             continue;
           }
+
+          if (!hasNext) {
+            break;
+          }
+
+          SourceDocument d = iter.next();
 
           if (!d.indexable()) {
             counters.unindexable.incrementAndGet();
--- a/src/main/java/io/anserini/index/IndexCollection.java
+++ b/src/main/java/io/anserini/index/IndexCollection.java
@@ -194,10 +194,15 @@ public final class IndexCollection {
             hasNext = iter.hasNext();
           } catch (NoSuchElementException e1) {
             break;
-          } catch (Exception e2) {
-            LOG.warn("Exception when parsing document: ", e2);
-            counters.errors.incrementAndGet();
-            continue;
+          } catch (RuntimeException e2) {
+            if (e2.getMessage().contains("IOException")) {
+              LOG.warn("Exception when parsing document: ", e2);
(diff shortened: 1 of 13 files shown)


lintool:
Hi @borislin I'm lacking the broader context - can you explain how the code was buggy before?

lintool:
@borislin Hrm. Interesting that it doesn't have an impact on the regressions...

borislin:
@lintool @Peilin-Yang I just pushed one more commit. This won't affect the effectiveness of our program for now. But it fixed an issue if user accidentally accessed the iterator again after closing it (marked by `atEOF` rather than the `NoSuchElementException`).

Review this change and I'll merge it after that...

lintool:
lgtm, but let's see if @Peilin-Yang has any comments.

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -183,17 +183,24 @@ public void run() {
...
+          boolean hasNext = false;
           try {
-            d = iter.next();
+            hasNext = iter.hasNext();
           } catch (NoSuchElementException e1) {
-            continue;
-          } catch (Exception e2) { // TODO: update related counters (#317)
+            break;
+          } catch (Exception e2) {
             LOG.warn("Exception when parsing document: ", e2);
```
Peilin-Yang: We are just reading the next document so it should be "Exception when reading document: ", right?

borislin:
@Peilin-Yang The reason I catch `IOException` in `hasNext()` is because the overwritten method `hasNext()` in `java.util.Iterator` doesn't throw `IOException`. We don't have this problem if `Segment` doesn't extend to `Iterator`.

Peilin-Yang:
Then we will have to pass the global error counter(s) to the collection class and update the error counter(s) when encountering IOException

Peilin-Yang:
Or we can throw a RuntimeException when encountering the IOException

borislin:
@Peilin-Yang When we encounter an `IOException`, we should throw an exception that's different than the one that's thrown when parsing a document (right now we use `RuntimeException`) so we can differentiate them in `IndexCollection` to determine whether to continue reading/parsing more documents or not.

How about changing current `RuntimeException` to `ParseException` and throw `RuntimeCollection` when we encounter `IOException` in `hasNext()`?

Peilin-Yang:
@borislin sounds good to me

lintool:
@borislin In general, I'm in favor of throwing our own exceptions.

borislin:
I've decided to throw `RuntimeException` when parsing document as well as for `IOException`. So, different than the approach I've discussed earlier. Reason is that `ParseException` is a checked exception and must be handled locally or declared in the method signature, conflicting with the method signature for `hasNext()` as discussed earlier.

In short, we are now using the exception message (containing "IOException" or not) to determine whether the `RuntimeException` is intended to be an `IOException` or not.

Peilin-Yang:
Yea, `ParseException` is not a subclass of `RuntimeException`...

Regarding the counters, I think it is better to have just three of them?

- Indexed
- Error (including parsing error, the bug of our code, etc)
- Empty

I doubt we really need that detailed counters. It is likely that nobody will remember what is exactly the rule of `skipped`....

borislin:
@Peilin-Yang 

Then we will have a lot of errors since there are a lot of deleted tweets and I think deleted tweets are technically not errors.

Can we deal with this specific index counter issue in #317 and get this PR shipped first? Let's fix the buggy code first.

I've also re-run all regression tests excluding cw12 and they have all passed.

Peilin-Yang:
Shouldn't we just fix it in this PR? I thought this is for #317, no?

borislin:
There are more places that we need to fix the counter issue. I believe we need to get #353 done first then #317.

PR #370, opened 2018-07-28
Patch for issue #353: Add MapCollection

borislin:
Now we can run command like: `target/appassembler/bin/MapCollections -collection TrecCollection -generator JsoupGenerator -threads 16 -input /tuna1/collections/newswire/disk12/ -mapper CountDocumentMapper` to get all the counter values.


lintool:
I would separate out a generic "MapCollection" API from an actual util called `CountDocuments` or something like that. `CountDocuments` passes a "mapper" into `MapCollections` that just increments a bunch of counters.

src/main/java/io/anserini/index/generator/DocumentGenerator.java
```diff
@@ -0,0 +1,153 @@
...
+      // If there's a transform, use it.
+      contents = transform != null ? transform.apply(src.content()) : src.content();
+    } catch (Exception e) {
+      LOG.error("Error extracting document text, skipping document: " + id, e);
+      counters.errors.incrementAndGet();
+      return null;
+    }
+
+    if (contents.trim().length() == 0) {
+      // Log and move on, otherwise output can be very noisy.
```
Peilin-Yang: The comment is not compatible with the code

borislin:
@Peilin-Yang 

To answer your questions, the purpose of this PR and #353 is that we want to build a generic mapper framework (let's call it `MapCollections` for now) that processes each file segment in parallel just like what our current indexer is doing.

This allows us to create utility functions such as `CountDocuments` and `IndexDocuments` which update counters and create indexes respectively by passing a mapper (`updateCounter` or `indexDocument`) to `MapCollections`.

So for the next step, we can refactor our current `IndexCollection` into calling both util functions `CounterDocuments` and `IndexDocuments`.

Correct? @lintool

borislin:
I've verified both ClueWeb09 and ClueWeb12b13 datasets. (`indexed` + `empty`) is the same as the ground truth.

ClueWeb09:
```
2018-07-30 18:42:49,830 INFO  [main] util.MapCollections (MapCollections.java:387) - # Final Counter Values
2018-07-30 18:42:49,832 INFO  [main] util.MapCollections (MapCollections.java:388) - indexed:       50,220,189
2018-07-30 18:42:49,832 INFO  [main] util.MapCollections (MapCollections.java:389) - empty:                234
2018-07-30 18:42:49,832 INFO  [main] util.MapCollections (MapCollections.java:390) - unindexed:              0
2018-07-30 18:42:49,832 INFO  [main] util.MapCollections (MapCollections.java:391) - unindexable:        1,492
2018-07-30 18:42:49,832 INFO  [main] util.MapCollections (MapCollections.java:392) - skipped:                0
2018-07-30 18:42:49,832 INFO  [main] util.MapCollections (MapCollections.java:393) - errors:                 0
2018-07-30 18:42:49,840 INFO  [main] util.MapCollections (MapCollections.java:396) - Total 50,220,189 documents indexed in 00:30:50
```

ClueWeb12b13:
```
2018-07-30 19:50:30,664 INFO  [main] util.MapCollections (MapCollections.java:387) - # Final Counter Values
2018-07-30 19:50:30,665 INFO  [main] util.MapCollections (MapCollections.java:388) - indexed:       52,249,039
2018-07-30 19:50:30,665 INFO  [main] util.MapCollections (MapCollections.java:389) - empty:             93,982
2018-07-30 19:50:30,665 INFO  [main] util.MapCollections (MapCollections.java:390) - unindexed:              0
2018-07-30 19:50:30,665 INFO  [main] util.MapCollections (MapCollections.java:391) - unindexable:       33,447
2018-07-30 19:50:30,665 INFO  [main] util.MapCollections (MapCollections.java:392) - skipped:                0
2018-07-30 19:50:30,665 INFO  [main] util.MapCollections (MapCollections.java:393) - errors:                 0
2018-07-30 19:50:30,686 INFO  [main] util.MapCollections (MapCollections.java:396) - Total 52,249,039 documents indexed in 00:40:28
```

I'm running on ClueWeb12 collection now.

lintool:
@borislin correct.

I want `MapCollections` for the ability to process the collection in other ways - for example, say I want to extract all named entities from all documents? I write a wrapper around an NER, and the feed it in as the mapper. Counting documents is simplest mapper.

src/main/java/io/anserini/index/generator/DocumentGenerator.java
```diff
@@ -0,0 +1,149 @@
+/**
```
lintool: hrm... why is this file part of the commit?

lintool:
@borislin did you push your latest code? I started doing CR, but it appears to be old code?

borislin:
@lintool Yes, it's the latest code.

Right now, some parts of the code are still tailored to counting documents. But `MapCollections` right now accepts a `DocumentMapper` (i.e. the wrapper you just mentioned). We can call something like `DocumentMapper.process()` to process each indexed document.

For the `DocumentGenerator.java` file, this is a copy of `LuceneDocumentGenerator.java` with `IndexCollection` replaced with `MapCollections`. I don't want to break current `IndexCollection` code.

In the future, if we refactor `IndexCollection` using `MapCollections`, then we can get rid of `DocumentGenerator.java` and use `LuceneDocumentGenerator.java` instead.

borislin:
Now we have `MapCollections` which accepts a mapper class `*DocumentMapper` that `extends DocumentMapper`. In `MapCollections`, it calls `mapper.process(d)` which passes a `SourceDocument` into the `process()` function, allowing each `SourceDocument` to be processed individually and in parallel by each mapper.

borislin:
@lintool 

In short, the reason I've created a copy of `LuceneDocumentGenerator` and call it `DocumentGenerator` is that `LuceneDocumentGenerator` has a protected variable `args` of type `IndexCollection.Args`. I want to use our custom version of `MapCollections.Args`.

I've just figured out a way to get rid of `DocumentGenerator` and make `LuceneDocumentGenerator` more generalized and able to accept `MapCollections.Arg` as an argument in constructor.

Please review this PR.

Btw, for `TweetCollection`, I need to make some changes in the `TweetGenerator` in order for us to use `MapCollections` on TweetCollection. I think we should do this in another PR?

lintool:
Hi @borislin - I think what I'm suggesting here is that we don't bother creating the actual Lucene document here:
https://github.com/castorini/Anserini/pull/370/files#diff-ab9174405388932274c02f05692f0e24R147

That way, this can be completely untangled from the indexing code.

Since the signature is `public boolean process(SourceDocument d)` anyway, right?

If the developer wants to write a mapper that takes a `SourceDocument` and convert it into a Lucene `Document`, they'd write it in the mapper itself?

I think that would be a better design... thoughts?

Yes, this means that the current implementation can't count parsing errors into Lucene, but I think that's fine for this specific application.

Also, can we have a high-level document on how to use, linked to "tools" section of readme?

src/main/java/io/anserini/util/MapCollections.java
```diff
@@ -0,0 +1,247 @@
...
+    @Option(name = "-generator", required = true, usage = "document generator in io.anserini.index.generator")
+    public String generatorClass;
+
+    @Option(name = "-mapper", required = true, usage = "mapper class in io.anserini.util.mapper")
+    public String mapperClass;
+
+    // optional arguments
+
+    @Option(name = "-output", metaVar = "[Path]", usage = "output path")
+    public String output;
```
lintool: If I'm not mistaken, this isn't used anywhere? (Other than logging the variable?)

borislin:
@lintool 

Right, I modified `LuceneDocumentGenerator` and made it more generalized because I didn't want to lose those signal updates in `createDocument()` in the generator.

If we can ignore them, then getting rid of dependency on `LuceneDocumentGenerator` is a good idea.

Btw, do you actually like the name `MapCollections`? Should we use `MapCollection` instead? Since we already have `IndexCollection` and `SearchCollection`...

src/main/java/io/anserini/util/MapCollections.java
```diff
@@ -0,0 +1,247 @@
...
+    public String input;
+
+    @Option(name = "-threads", metaVar = "[Number]", required = true, usage = "Number of Threads")
+    public int threads;
+
+    @Option(name = "-collection", required = true, usage = "collection class in io.anserini.collection")
+    public String collectionClass;
+
+    @Option(name = "-generator", required = true, usage = "document generator in io.anserini.index.generator")
+    public String generatorClass;
```
lintool: I don't think you need this now either?

src/main/java/io/anserini/util/mapper/DocumentMapper.java
```diff
@@ -0,0 +1,30 @@
+package io.anserini.util.mapper;
+
+
```
lintool: extra space...

src/main/java/io/anserini/util/mapper/DocumentMapper.java
```diff
@@ -0,0 +1,30 @@
+package io.anserini.util.mapper;
```
lintool: Add license header?

src/main/java/io/anserini/util/MapCollections.java
borislin: Since developers can build their own mapper, I think we should allow them to write output to somewhere other than stdout? Providing this option will be useful to them...

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
```diff
@@ -0,0 +1,114 @@
...
+    counters.indexed.incrementAndGet();
+  }
+
+  public void incrementIndexedBy(int numIndexed) {
+    counters.indexed.addAndGet(numIndexed);
+  }
+
+  public boolean process(SourceDocument d) {
+    if (!d.indexable()) {
+      incrementUnindexable();
```
lintool: Is there a reason why you dispatch to a method here as opposed to just incrementing the counter directly?

And furthermore, does it make sense for incrementFoo methods to be public?

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
```diff
@@ -0,0 +1,114 @@
...
+      return false;
+    }
+
+    incrementIndexed();
+    return true;
+  }
+
+  public void printResult(long durationMillis) {
+    LOG.info("# Final Counter Values");
+    LOG.info(String.format("indexed:     %,12d", counters.indexed.get()));
```
lintool: probably should be called `processed`?

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
borislin: 1) Huhhh, actualy, no reason.

2) I made them public because there I have to call them in multiple places inside `MapCollections`. There is no way to put all the counter updates inside `process()` and keep the signature of current `process()` method unchanged. So I used something like:
```
if (mapper.isCountDocumentMapper()) {
  ((CountDocumentMapper)mapper).incrementSkipped();
}
```
inside `MapCollections`.

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
borislin: let's make it more distinctive from `process`? how about `done` or `complete`?

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
lintool: I see. This is a bit janky but I can't think of anything better right now... so this is fine, and we'll let this evolve as we find more use cases for mappers.

PR #372, opened 2018-07-30
Oracle effectiveness

Peilin-Yang:
This PR adds scripts and auto-doc-gen test for oracle effectiveness numbers.
The main reason of this PR is to let the readers of our JDIQ paper easily repeat the effectiveness numbers shown in the paper.
The added code can also be used to generate optimal performance if we have underlying changes, e.g. Lucene upgrade.


lintool:
Can we rename to JDIQ2018 or something that makes the connection explicit and clear? In the future, when we run more oracle experiments, it'll get confusing...

Peilin-Yang:
You mean the filename? dir names?

lintool:
Both, the connection of the experiments with our JDIQ should be absolutely clear?

docs/oracle-effectiveness.md
```diff
@@ -0,0 +1,94 @@
+#### disk12
```
lintool: this file should probably be renamed `jdiq2018-results.md`?

src/main/python/jdiq2018_effectiveness/README.md
```diff
@@ -0,0 +1,22 @@
+### JDIQ2018 Effectiveness Scripts
```
lintool: why put this file here and not `docs/`?

src/main/python/jdiq2018_effectiveness/performance.py
```diff
@@ -0,0 +1,134 @@
+# -*- coding: utf-8 -*-
```
lintool: By performance you actually mean effectiveness? I like to be clear about the distinction between effectiveness (e.g., AP) vs. efficiency (e.g., latency).

src/main/python/jdiq2018_effectiveness/performance.py
```diff
@@ -0,0 +1,134 @@
...
+            frameinfo = getframeinfo(currentframe())
+            self.logger.error(frameinfo.filename, frameinfo.lineno)
+            self.logger.error('[Performances Constructor]:Please provide a valid index path - ' + self.index_path)
+            exit(1)
+
+        self.run_files_root = 'run_files'
+        self.eval_files_root = 'eval_files'
+        self.performances_root = 'performance_files'
+
+    def gen_output_performances_paras(self, output_root):
```
lintool: what's "paras"? params?

src/main/python/jdiq2018_effectiveness/performance.py
```diff
@@ -0,0 +1,134 @@
...
+                all_results[k].append( os.path.join(output_root, self.eval_files_root, fn) )
+        for k in all_results:
+            performace_fn = os.path.join(output_root, self.performances_root, k)
+            tmp = [ self.index_path, performace_fn ]
+            tmp.extend( all_results[k] )
+            all_paras.append(tuple(tmp))
+
+        return all_paras
+
+    def output_performances(self, output_fn, eval_fn_list):
```
lintool: metric instead of performances?

src/main/python/jdiq2018_effectiveness/performance.py
```diff
@@ -0,0 +1,134 @@
...
+                        'model': model,
+                        'topic': topic,
+                        'metric': metric,
+                        'params': all_performance[metric]['all']['max']['para'],
+                        'actual': all_performance[metric]['all']['max']['value']
+                    }
+                    data.append(res)
+        return data
+
+
```
lintool: kill empty lines?

lintool:
Also, we should update the log here: `docs/regression-tuna.md` - maybe fill in everything except for the commit id, and then add that back in as part of another PR?

Issue #374, opened 2018-08-02
`Nan` score in RM3 term weight on CAR corpus


Victor0118:
I got `Nan` weight when applying RM3 to CAR corpus for one query. The reason is that norm[i] == 0 in https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/rerank/lib/Rm3Reranker.java#L162 for one 
special `i`. Then all terms got Nan weight because of https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/rerank/lib/Rm3Reranker.java#L168. Detail reason causing the empty doc `i` need to be investigated.

lintool:
1. Yes, please investigate how an empty doc made its way into the index... the indexer is supposed to discard empty documents.
2. The RM3 code should not be trying to estimate relevance models from empty documents anyway - please add a check there to discard any empty documents.

Victor0118:
It is because there is a Document that contains only one word: `gītā`
`Document<stored,indexed,tokenized,omitNorms,indexOptions=DOCS<id:2e68fcb2ca494be68c5721f87a1a2be6c8121341> stored<raw:gītā>>`

Since it contains characters not in [a-z0-9], it is filtered out at https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/rerank/lib/Rm3Reranker.java#L191 and it is not in the RM3 candidates. So the filtered doc is empty.

PR #376, opened 2018-08-02
Fix zero norm bug

Victor0118:
@lintool @Peilin-Yang Could you take a quick look at this PR?

The bug detail is in #374 .


lintool:
@Victor0118 I don’t think this is the right fix... this will impact the quality of the estimated RM. we should actually remove those documents...

Victor0118:
According to the [source code](https://github.com/apache/lucene-solr/blob/master/lucene/analysis/common/src/java/org/apache/lucene/analysis/en/EnglishAnalyzer.java), Englishanalyzer does not remove the word containing special character.

But in RM3 [implementation](https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/rerank/lib/Rm3Reranker.java#L191) in Anserini, you remove words contain character not in [a-z0-9]. Is there any special reason to do this?

lintool:
The `[a-z0-9]` throws away terms that are likely to be junk... if we remove that line, then all the RM3 results will change, then we need to verify all the effectiveness numbers, etc. I'd rather not do that...

Victor0118:
Another choice is adding a condition like this:
``` java
if (len(doc) < threshold){
   // do not add doc to the RM3 candidate
}
```

This can also avoid the junk document in Wikipedia corpus. Good?

lintool:
Yes, although that might also change regression results...

Why don't we just throw away all feature vectors with norm of 0?

Victor0118:
Okay. That makes sense!

lintool:
Can you run the regression experiments to make sure everything still works?
https://github.com/castorini/Anserini/blob/master/docs/regression-tuna.md

Victor0118:
OKay!

Victor0118:
```
no segments* file found in MMapDirectory@/tuna1/indexes/lucene-index.aquaint.pos+docvectors+rawdocs
no segments* file found in MMapDirectory@/tuna1/indexes/lucene-index.nyt.pos+docvectors+rawdocs
no segments* file found in MMapDirectory@/tuna1/indexes/lucene-index.tweets2011.pos+docvectors+rawdocs
no segments* file found in MMapDirectory@/tuna1/indexes/lucene-index.tweets2013.pos+docvectors+rawdocs
```
@lintool All regression tests passed except four index files not found above.

Victor0118:
@lintool All regression test passed!

lintool:
@Victor0118 great!
One final thing - can you add a comment that explains why we do that check in the code? Then we'll be good for merging.

Victor0118:
@lintool Done.

lintool:
How about a comment that is more descriptive? Like:

Avoids zero-length feedback documents, which causes division by zero when computing term weights.
Zero-length feedback documents occur (e.g., with CAR17) when a document has only terms that accents (which are indexed, but not selected for feedback).

Victor0118:
@lintool Thanks for your kind suggestion. I have updated the comment.

Issue #377, opened 2018-08-02
IOException handling needs improvement


borislin:
Currently in each collection, we use `hasNext()` and `next` to iterate through the next document. When we encounter an `IOException` in `hasNext()`, we stop reading more documents. However, if a subclass of IOException is thrown, we won't be able to catch it as a signal to stop reading since we use `if (e2.getMessage().contains("IOException"))` and error message of subclass of IOException won't contain "IOException".

We need a more customized message without losing the actual error message.

borislin:
#378 has fixed this

PR #378, opened 2018-08-02
Patch for #377: Make IOException error message more customized

borislin:


lintool:
Hrm... let's actually think through what the right solution is:

The issue is that `IOException`s need to be checked, so we simply catch and rethrow as `RuntimeException`s, and now the PR just adds a msg that says it's a file exception.

All of this happens in the `hasNext()` method. So I'm wondering, why not just catch the exception, log error, and return `false`. This would abort record reading without an abnormal control flow....

Thoughts?

borislin:
@lintool That's actually my original design. We've discussed extensively in #369.

If we don't want to throw a `RuntimeException` when we catch `IOException` in `hasNext()`, then we need to pass `counters` to `createFileSegment` in all collections and update the `counters.error` counter inside `hasNext()`.

Do you prefer this way? I'm fine with both.

lintool:
Ah, I see. Sorry I lost the context.

I don't like the idea of passing around the counters too much, so I like this better.

PR #387, opened 2018-08-05
update jdiq effectiveness commit sha

Peilin-Yang:


lintool:
hey @Peilin-Yang this might be unrelated, but why are `docs/experiments-jdiq2018.md` and `docs/jdiq2018-effectiveness.md` two separate files? Why don't we auto-gen the effectiveness numbers as part of `experiments-jdiq2018.md`?

Peilin-Yang:
Sure, I can do that

PR #392, opened 2018-08-09
fix search runtag arg

Peilin-Yang:
Currently the `runtag` search arg does not work properly. 
As we need this for TREC submission this PR fixes this issue.


lintool:
Do we really want to append all the metadata? Let's just use the runtag submitted on the cmdline verbatim?

PR #393, opened 2018-08-09
Remove -optimize option for regression test

Peilin-Yang:


lintool:
Why have tests failed?

PR #394, opened 2018-08-10
Refactor dumpRawDoc in IndexUtils

Peilin-Yang:
Context: I encountered file quota issue when dumping too many docs.
The new design discards the support for dumping as archival.
Instead, we leverage multi-threading to dump docs to a directory with each thread dump one doc.
Users can then feel free to archive and/or compress by themselves.

### DIFF
--- a/src/main/java/io/anserini/index/IndexUtils.java
+++ b/src/main/java/io/anserini/index/IndexUtils.java
@@ -355,17 +355,20 @@ public class IndexUtils {
       LOG.info("Create dump directory failed: "+ outputDir);
       return;
     }
+    
     final class DumpThread extends Thread {
       final private IndexReader reader;
       final private String docid;
       final private String outputDir;
       final private boolean prependDocid;
+      
       public DumpThread(IndexReader reader, String docid, String outputDir, boolean prependDocid) throws IOException {
         this.reader = reader;
         this.docid = docid;
         this.outputDir = outputDir;
         this.prependDocid = prependDocid;
       }
+      
       @Override
       public void run() {
         try {
@@ -383,6 +386,7 @@ public class IndexUtils {
         }
       }
     }
+    
     List<String> docids = new ArrayList<>();
     String docid;
     while ((docid = bRdr.readLine()) != null) {


lintool:
This seems like a good use case for `io.anserini.util.MapCollections`? I've been refactoring that to make more useful... perhaps opportunity to try out?

Processing collections in parallel on multiple threads requires boilerplate... we shouldn't write more than once?

lintool:
Oh actually, I take it back. You're mapping over the _index_, not the _collection_. nm.

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -342,37 +348,69 @@ public void dumpRawDocuments(String reqDocidsPath, boolean prependDocid) throws
...
-      if (doc == null) {
-        throw new NotStoredException("Raw documents not stored!");
+    
+    String outputDir = Paths.get(reqDocidsPath).getFileName()+"_rawdocs.dump";
+    boolean mkdirStatus = new File(outputDir).mkdirs();
+    if (!mkdirStatus) {
+      LOG.info("Create dump directory failed: "+ outputDir);
+      return;
+    }
+    final class DumpThread extends Thread {
```
lintool: newline above.

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -342,37 +348,69 @@ public void dumpRawDocuments(String reqDocidsPath, boolean prependDocid) throws
...
+      final private IndexReader reader;
+      final private String docid;
+      final private String outputDir;
+      final private boolean prependDocid;
+      public DumpThread(IndexReader reader, String docid, String outputDir, boolean prependDocid) throws IOException {
```
lintool: newline

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -342,37 +348,69 @@ public void dumpRawDocuments(String reqDocidsPath, boolean prependDocid) throws
...
+        this.prependDocid = prependDocid;
       }
-      TarArchiveEntry tarEntry = new TarArchiveEntry(new File(docid));
-
-      byte[] bytesOut = doc.stringValue().getBytes(StandardCharsets.UTF_8);
-      tarEntry.setSize(bytesOut.length + (prependDocid ? String.format("<DOCNO>%s</DOCNO>\n", docid).length() : 0));
-      tOut.putArchiveEntry(tarEntry);
-      if (prependDocid) {
-        tOut.write(String.format("<DOCNO>%s</DOCNO>\n", docid).getBytes());
+      @Override
```
lintool: newline

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -342,37 +348,69 @@ public void dumpRawDocuments(String reqDocidsPath, boolean prependDocid) throws
...
+      public void run() {
+        try {
+          Document d = reader.document(convertDocidToLuceneDocid(docid));
+          IndexableField doc = d.getField(LuceneDocumentGenerator.FIELD_RAW);
+          if (doc == null) {
+            LOG.error("Raw documents not stored: " + docid);
```
lintool: This should be an exception that bombs out of the util, no?
Without the docs stored you really can't do anything...

src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -342,37 +348,69 @@ public void dumpRawDocuments(String reqDocidsPath, boolean prependDocid) throws
...
       }
-      tOut.write(bytesOut);
-      tOut.closeArchiveEntry();
-
-      if (counter % 100000 == 0) {
-        LOG.info(counter + " files have been dumped.");
+    }
+    List<String> docids = new ArrayList<>();
+    String docid;
+    while ((docid = bRdr.readLine()) != null) {
```
lintool: Isn't there some higher-level util that reads all lines in from a file in one call?

PR #395, opened 2018-08-10
Change axiomatic ranking model's to 0.5

Peilin-Yang:
I just found `f2exp.s` and `f2log.s` were 0.25 where they should be 0.5


lintool:
Can you explain a bit more? Does this change regressions? etc.?

PR #396, opened 2018-08-10
Refactoring of MapCollections and doc counter

lintool:
Cleaned up and simplified code; adopted more functional style.


src/main/java/io/anserini/util/MapCollections.java
```diff
@@ -98,50 +101,22 @@ private MapThread(DocumentCollection collection, Path inputFile) {
...
-          numMapped++;
-        }
+        // We're calling these records because the documents may not in indexable.
+        AtomicInteger records = new AtomicInteger();
+        try {
+          iter.forEachRemaining(d -> {
+            mapper.process(d, context);
+            records.incrementAndGet();
+          });
+        } catch (NoSuchElementException e) {}
```
borislin: Se we are not catching `RuntimeException` here?

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
```diff
@@ -27,48 +27,16 @@
...
-     */
-    public AtomicLong errors = new AtomicLong();
-  }
-
-  private final Counters counters;
+  private CountDocumentMapperContext counters;
   private final Set whitelistDocids;
 
-  public CountDocumentMapper(MapCollections.Args args) throws Exception {
+  public CountDocumentMapper(MapCollections.Args args)
```
borislin: Put these two lines on one line?

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
```diff
@@ -78,53 +46,32 @@ public CountDocumentMapper(MapCollections.Args args) throws Exception {
...
   }
 
-  public boolean process(SourceDocument d) {
-    if (!d.indexable()) {
-      incrementUnindexable();
-      return false;
+  @Override
+  public void process(SourceDocument doc, DocumentMapperContext context) {
+    if (!doc.indexable()) {
+      ((CountDocumentMapperContext) context).unindexable.incrementAndGet();
```
borislin: Simplify code by using `counters` instead of `((CountDocumentMapperContext) context)`? And I think the meaning is more clear...

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
```diff
@@ -78,53 +46,32 @@ public CountDocumentMapper(MapCollections.Args args) throws Exception {
...
+      return;
     }
 
-    if (whitelistDocids != null && !whitelistDocids.contains(d.id())) {
-      incrementSkipped();
-      return false;
+    if (whitelistDocids != null && !whitelistDocids.contains(doc.id())) {
+      ((CountDocumentMapperContext) context).skipped.incrementAndGet();
```
borislin: Simplify code by using `counters` instead of `((CountDocumentMapperContext) context)`? And I think the meaning is more clear...

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
```diff
@@ -78,53 +46,32 @@ public CountDocumentMapper(MapCollections.Args args) throws Exception {
...
+      return;
     }
 
-    incrementIndexed();
-    return true;
+    ((CountDocumentMapperContext) context).processed.incrementAndGet();
```
borislin: Simplify code by using `counters` instead of `((CountDocumentMapperContext) context)`? And I think the meaning is more clear...

src/main/java/io/anserini/util/MapCollections.java
lintool: okay, will do.

lintool: Actually, I just noticed, there's another catch block below:
```
catch (Exception e) {
        LOG.error(Thread.currentThread().getName() + ": Unexpected Exception:", e);
      }
```
This will catch all exceptions for this file segment, so I think we're good.

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
lintool: +1

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
lintool: For generality, I actually want the access the counter through the context. Rationale:  I want to test the generality of the `process` method signature - all the data structures the mapper needs access to should be accessible without referencing anything external.

src/main/java/io/anserini/util/mapper/CountDocumentMapper.java
lintool: Same response as above + below.

lintool:
@borislin let me know what you think!

src/main/java/io/anserini/util/MapCollections.java
borislin: I think we can remove this `try...catch...` block after #399 is merged.

PR #401, opened 2018-08-12
Cleaned up top-level regression docs

lintool:
Added CAR17, few other minor tweaks.


lintool:
I just copied the timestamps from the Git commit logs. -4 is east coast, -7 is west coast, I think. Git take local settings?

Issue #402, opened 2018-08-12
Wrong description of parameter options (topic v.s. topics) for experiments on trec collections


lukuang:
Hi,

I tried to run experiments using the provided commands on [gov2](https://github.com/castorini/Anserini/blob/master/docs/experiments-gov2.md) and [robust04](https://github.com/castorini/Anserini/blob/master/docs/experiments-robust04.md). However, the parameter options of performing retrieval seem to be wrong. Instead of "-topics", the commands listed on the above links use "-topic" instead.

Peilin-Yang:
merged as #405

PR #405, opened 2018-08-13
Fix experiments docs generation error #402

Peilin-Yang:
`-topics` instead of `-topic`


lintool:
hey @Peilin-Yang I'm a bit confused - the code that generates the doc _isn't_ the same that actually runs the commands? Otherwise how can we generate non-valid run commands?

Peilin-Yang:
@lintool no, they are different.
The docs are generated with testing code while the actual tests are fired from the python script.

lintool:
okay I see - merge once CI turns green...

Issue #403, opened 2018-08-13
Refactor WashingtonPostCollection


Peilin-Yang:
Current implementation is a bit problematic: 
Currently the [WashingtonPostCollection](https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/collection/WashingtonPostCollection.java)'s `next` method returns an object that contains only three fields: `id`, `publishedDate` and `content`. This object largely loses the original structure of the document which makes some use cases difficult to implement. For example, for News Track which uses the same corpus, we need filter out "Opinion", "Letters to the Editor", or "The Post's View" sections, as labeled in the "kicker" field b/c these documents are non-relevant and this is explicitly said in the guideline.  Ideally, these documents could be easily filtered by specifying a boolean query filter but without indexing the "kicker" field it is impossible to do so. 

I think we should index the `WashingtonPostCollection` the similar way with what `TweetCollection` does. That is, for `WashingtonPostCollection` we should return the original document with only necessary preprocessing. Then adding a `WashingtonPostGenerator` to index the required fields.

Peilin-Yang:
Merged as #408

PR #408, opened 2018-08-14
Refactor WashingtonPost Collection Indexing

Peilin-Yang:
Please see #403 for details.

This PR adds the `WapoGenerator` where the actual contents are parsed.
Now the `WashingtonPostCollection` is mainly a Json adaptor and it sends the raw Json object to `WapoGenerator`.

### DIFF
--- a/src/main/java/io/anserini/index/generator/WapoGenerator.java
+++ b/src/main/java/io/anserini/index/generator/WapoGenerator.java
@@ -1,0 +1,135 @@
+/**
+ * Anserini: An information retrieval toolkit built on Lucene
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


src/main/java/io/anserini/index/generator/WapoGenerator.java
```diff
@@ -0,0 +1,143 @@
+/**
```
lintool: I know it's long a verbose, but consider renaming `WashingtonPostDocumentGenerator`?

lintool:
@borislin I'm not familiar with this code, can you please review?

Also, @Peilin-Yang please verify regressions pass? Now that we have lots of regressions, be specific about which ones you ran? Thanks!

PR #411, opened 2018-08-16
Adding code for TREC2018 News Track Background Linking task runs

Peilin-Yang:
This task is a bit different from other adhoc/web tasks where the query of this task is actually a document id instead of one or multiple terms.
I basically changed three places to accommodate the task:
1. Create the `NewsTrackBLTopicReader` where it reads the query document id and applies different strategies to generate the actual query terms. The strategies include: 
    1) extract high tf-idf terms with the weights from the document
    2) extract the terms as their original order from the document and applies SDM
    3) extract the unweighted high tf-idf terms and applies Axiomatic Reranking
    4) Similar to ii, but on paragraph level. The final result is the generated by picking the results from the paragraph queries in a round-robin fashion. 
    5) Similar to iii, but on paragraph level. The final result is the generated by picking the results from the paragraph queries in a round-robin fashion. 
2. Create the function `searchNewsBackground` in `SearchCollection.java` in order to deal with the non-trivial processing of the task.
3. Create `NewsTrackBLReranker` where it can remove the duplicated results as per task requirement.


src/main/java/io/anserini/rerank/lib/NewsTrackBLReranker.java
```diff
@@ -0,0 +1,139 @@
+/**
```
lintool: How about `NewsBackgroundLinkingReranker`?

src/main/java/io/anserini/rerank/lib/NewsTrackBLReranker.java
```diff
@@ -0,0 +1,139 @@
...
+import io.anserini.rerank.Reranker;
+import io.anserini.rerank.RerankerContext;
+import io.anserini.rerank.ScoredDocuments;
+import io.anserini.search.topicreader.NewsTrackBLTopicReader;
+import org.apache.lucene.document.Document;
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.Terms;
+import org.apache.lucene.index.TermsEnum;
+
+import java.util.*;
```
lintool: Let's avoid wildcard imports?

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -238,10 +236,89 @@ public void close() throws IOException {
     }
 
     List<String> queryTokens = AnalyzerUtils.tokenize(analyzer, queryString);
-    RerankerContext context = new RerankerContext<>(searcher, qid, query, queryString, queryTokens, null, args);
+    RerankerContext context = new RerankerContext<>(searcher, qid, query, null, queryString, queryTokens, null, args);
 
     return cascade.run(ScoredDocuments.fromTopDocs(rs, searcher), context);
   }
+  
+  public<K> ScoredDocuments searchNewsBackground(IndexSearcher searcher, K qid, String queryString)
```
lintool: let's make the terminology consistent?
above you called it "BL"... here it's something different.
I propose "BackgroundLinking"

src/main/java/io/anserini/search/topicreader/NewsTrackBLTopicReader.java
```diff
@@ -0,0 +1,275 @@
+/**
```
lintool: Same comment on naming.

lintool:
Just a quick review for now...

@Kytabyte comments?

src/main/java/io/anserini/rerank/lib/NewsBackgroundLinkingReranker.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.List;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.HashMap;
+import java.util.Map;
+
+import static io.anserini.index.generator.LuceneDocumentGenerator.FIELD_BODY;
+import static io.anserini.index.generator.LuceneDocumentGenerator.FIELD_ID;
+
+
```
lintool: extra space

src/main/java/io/anserini/rerank/lib/NewsBackgroundLinkingReranker.java
```diff
@@ -0,0 +1,144 @@
...
+/*
+* TREC News track Background Linking task postprocessing.
+* Near-duplicate documents (similar/same with the query docid) will be removed by comparing
```
lintool: track -> Track?

src/main/java/io/anserini/rerank/lib/NewsBackgroundLinkingReranker.java
```diff
@@ -0,0 +1,144 @@
...
+    final Map<String, Long> queryTermsMap = convertDocVectorToMap(reader, queryDocId);
+    
+    List<Map<String, Long>> docsVectorsMap = new ArrayList<>();
+    for (int i = 0; i < docs.documents.length; i++) {
+      String docid = docs.documents[i].getField(FIELD_ID).stringValue();
+      docsVectorsMap.add(convertDocVectorToMap(reader, docid));
+    }
+    
+    // remove the duplicates: 1. the same doc with the query doc 2. duplicated docs in the results
+    Set<Integer> forbiddenDocIdsIndex = new HashSet<>();
```
lintool: "forbidden" sounds scary... how about just "duplicates"?

src/main/java/io/anserini/rerank/lib/NewsBackgroundLinkingReranker.java
```diff
@@ -0,0 +1,144 @@
...
+      }
+    }
+  
+    return scoredDocs;
+  }
+  
+  private Map<String, Long> convertDocVectorToMap(IndexReader reader, String docid) {
+    Map<String, Long> m = new HashMap<>();
+    try {
+      Terms terms = reader.getTermVector(NewsBackgroundLinkingTopicReader.convertDocidToLuceneDocid(reader, docid), FIELD_BODY);
```
lintool: I think this line is too long?

src/main/java/io/anserini/rerank/lib/NewsBackgroundLinkingReranker.java
```diff
@@ -0,0 +1,144 @@
...
+        long tf = it.totalTermFreq();
+        m.put(term, tf);
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+    return m;
+  }
+  
+  private double dotProduct(Map<String, Long> profile1, Map<String, Long> profile2) {
```
lintool: we have similar methods in `io.anserini.util.FeatureVector` - does it make sense to combine functionality?

src/main/java/io/anserini/rerank/RerankerContext.java
```diff
@@ -27,16 +27,18 @@
...
   private final Query filter;
   private final SearchArgs searchArgs;
 
-  public RerankerContext(IndexSearcher searcher, K queryId, Query query, String queryText,
+  public RerankerContext(IndexSearcher searcher, K queryId, Query query, String queryDocId, String queryText,
       List<String> queryTokens, Query filter, SearchArgs searchArgs) throws IOException {
     this.searcher = searcher;
     this.query = query;
     this.queryId = queryId;
+    this.queryDocId = queryDocId;
```
lintool: what exactly the is the queryDocId here?
We should be careful to add to `RerankerContext`... it's getting pretty bloated...

Peilin-Yang: Yes. 
So for News BL Track, the query is actually a docid and the task is to find the background news for that docid (news). There is also a query id from which the query docid is probably part of the ranking list.
It is probably wise to keep queryId, queryDocId and queryString at the same time.  
For example, the guideline explicitly states that duplicated results will be judged as non-relevant. So for the post-processing (implement as reranking) probably includes the comparison and dedup which requires the raw doc retrieved by queryDocId.

Issue #409, opened 2018-08-14
Regression Failure: MB13


lintool:
I'm getting the following error on MB13 regression (from indexing scratch):

```
2018-08-14 15:45:54,999 INFO  [pool-2-thread-19] index.IndexCollection$IndexerThread (IndexCollection.java:243) - data/statuses.log.2013-02-08-15.gz: 185018 docs added.
2018-08-14 15:56:55,819 WARN  [main] index.IndexCollection (IndexCollection.java:370) - Unexpected difference between number of indexed documents and index maxDoc.
2018-08-14 15:56:55,819 INFO  [main] index.IndexCollection (IndexCollection.java:373) - # Final Counter Values
2018-08-14 15:56:55,820 INFO  [main] index.IndexCollection (IndexCollection.java:374) - indexed:      203,162,923
2018-08-14 15:56:55,821 INFO  [main] index.IndexCollection (IndexCollection.java:375) - empty:            143,890
2018-08-14 15:56:55,821 INFO  [main] index.IndexCollection (IndexCollection.java:376) - unindexed:     55,894,079
2018-08-14 15:56:55,821 INFO  [main] index.IndexCollection (IndexCollection.java:377) - unindexable:            0
2018-08-14 15:56:55,821 INFO  [main] index.IndexCollection (IndexCollection.java:378) - skipped:       88,516,634
2018-08-14 15:56:55,821 INFO  [main] index.IndexCollection (IndexCollection.java:379) - errors:             2,754
2018-08-14 15:56:55,833 INFO  [main] index.IndexCollection (IndexCollection.java:382) - Total 203,145,943 documents indexed in 08:33:45
2018-08-14 15:57:02,736 - regression_test - INFO - ==========Verifying Index==========
2018-08-14 15:57:02,738 - regression_test - INFO - [Index]: lucene-index.mb13.pos+docvectors+rawdocs
2018-08-14 15:57:06,630 - regression_test - INFO - documents:             203145916
Traceback (most recent call last):
  File "src/main/python/run_regression.py", line 203, in <module>
    verify_index(yaml_data, args.index, args.dry_run)
  File "src/main/python/run_regression.py", line 98, in verify_index
    assert value == yaml_data['index_stats'][stat]
AssertionError
```
I'm at this current commit:

```
commit d7764a12c21de01492dd7967cfe347186f068746
Author: Boris Lin <boruilin224@gmail.com>
Date:   Tue Aug 14 10:00:10 2018 +0800

[...]

PR #415, opened 2018-08-17
Fixed broken MB13 regression

lintool:
See issue #409 documenting MB13 regression failure. Source can be traced back to
6b19d373634d510e8ac5f5b26b1a3d5c321926e4

In MB13 there are duplicate documents: optimize flag forces segment merge to clean this up,
yielding different index stats.

Rerunning regression now as final confirmation.

### DIFF
--- a/src/main/resources/regression/mb13.yaml
+++ b/src/main/resources/regression/mb13.yaml
@@ -21,8 +21,6 @@ index_options:
   - -tweet.stemming
 index_stats:
   documents: 203145916
-  documents (non-empty): 203143269
-  total terms: 2099082414
 topic_reader: Microblog
 topics:
   - name: "[TREC 2013 Microblog Track](http://trec.nist.gov/data/microblog2013.html)"
--- a/src/main/resources/regression/mb13.yaml
+++ b/src/main/resources/regression/mb13.yaml
@@ -21,8 +21,8 @@ index_options:
   - -tweet.stemming
 index_stats:
   documents: 203145916
-  documents (non-empty): 203143249
-  total terms: 2099082204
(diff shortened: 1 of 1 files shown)


lintool:
My hypothesis is that "update" is the culprit... not sure how it's implemented behind the scenes, but - say, if update is actually an insertion + deletion, then deletion leaves a tombstone that would be reclaimed on optimize, which is a compaction, right?

Peilin-Yang:
I dunno... Just curious that why the issue wasn't there.. 
Apparently, for collection such as MB13 that has duplicated documents we will not be able to get the actual indexed documents until the index is completely built.

lintool:
Regression failed... It appears that the number of docs are non-deterministic! What??

```
$ target/appassembler/bin/IndexUtils -index lucene-index.mb13.pos+docvectors+rawdocs -stats
Index statistics
----------------
documents:             203145916
documents (non-empty): 203143470
unique terms:          -1
total terms:           2099084597
stored fields:
  contents (indexOption: DOCS_AND_FREQS_AND_POSITIONS, hasVectors: true)
  id (indexOption: DOCS, hasVectors: false)
  lang (indexOption: DOCS, hasVectors: false)
  screen_name (indexOption: DOCS, hasVectors: false)
```

Peilin-Yang:
@lintool Is that possible different indexing threads can not see each other's (duplicated) documents? Thus we will have to turn on `-optimize` to finalize?

Peilin-Yang:
Tested with `-optimize` two times and both have passed.
The run files are exactly the same too. 
So it seems that we will have to keep `-optimize` for MB collections.

Issue #404, opened 2018-08-13
Adding capability to index multiple fields


Peilin-Yang:
Currently, it is impossible to index fields other than `contents` for most of the collections.
However, such functionality is desirable for LTR, Background linking and other use cases. 

With the help of Lucene's native fielded indexing, it is possible to add this functionality at the Generators.

lintool:
Really old issue... but we've made progress: https://github.com/castorini/pyserini/blob/master/docs/experiments-elastic.md

Closing for now.

PR #418, opened 2018-08-23
Adding the capability to adding fields other than `id` and `contents`

Peilin-Yang:
In this PR, I create the framework to add fields other than `id` and `contents`.
The implementation leaves the specific fields to the collections, i.e., what fields are supported should be decided by each collection and the type of the fields are also finalized in the collection.
Then the additional list of fields are passed to the generator to be added into Lucene Documents. 

Also, as the part of this PR I've added 4 fields `url`, `title`, `anchor_text` and `html_body` to ClueWeb collections.

### DIFF
--- a/docs/experiments-trec2018.md
+++ b/docs/experiments-trec2018.md
@@ -104,6 +104,22 @@ The above commands generate _ALL_ results we have put in the notebook paper. To 
 TBD
 ```
 
+### Task 3: Web Track 2014
+
+Steps to deterministically reproduce the results:
+
+#### Get code and data:
+```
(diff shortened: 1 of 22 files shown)


mrnolte:
Hey guys,
are there any updates on this pull request and issue #404?
That feature would be really appreciated.
Cheers,
Robin

lintool:
hi @Peilin-Yang we never merged this (for whatever reason) and it's gone out of date... what do you want to do? should we close it for now and reopen when we have the need?

lintool:
Closing, since there are too many conflicts to merge cleanly... we'll have to reimplement when we need these features...

PR #433, opened 2018-09-14
Add Dockerfile and scripts for integrating with Solr.

### LINKED ISSUE #364: Explore integration of Anserini w/ Solr or Elasticsearch
Solr or Elasticsearch are standard industry solutions for search, but their APIs are higher level than what's needed for research. This is where Anserini comes in. However, can we have our cake and eat it too?

By this, I mean: can we take an Anserini Lucene index and stick it into Solr or Elasticsearch directly, and have it be usable? That way, we can take advantage of standard user interfaces built on top of either search engine.

A quick search suggests that this is fairly straightforward with Solr: https://datafireball.com/2016/03/10/solr-load-lucene-index/

... but very difficult to do with Elasticsearch...

@r-clancy can you look into this and play around?

r-clancy:
See PR #432 for documentation.

Closes #364


lintool:
hrmm... this is a lot of data to check in and it's going to make the repo quite messy... let's discuss what the right thing to do is.

Should the files really be checked into `.docker`? Would be make sense to put under `src/resources`? Alternatively, should it be in a separate repo, e.g. `Anserini-docker`?

@Peilin-Yang any thoughts?

lintool:
@r-clancy I think it make sense to merge #433 and #432 together in a single PR so we can comment on put at once.

Merge one branch into the other to update PR and close the other one?

r-clancy:
I've merged them together into a single PR.

I got rid of one directory that had lots of files (`_default`). The files in `.docker/configsets/{core17,mb11}` are required to create the Solr core (index), so I think this is a good location for them. They could be moved into `src/resources`, though.

We could move it to a new repo if that would be a better place for it.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN apk update && apk add curl
+
+# Work as the solr user.
+USER solr
+
+# Work from the base dir of the Solr install.
+WORKDIR /opt/solr/
+
+# Anserini JAR for TwitterAnalyzer.
+COPY --chown=solr target/anserini-0.2.1-SNAPSHOT-fatjar.jar lib/
```
r-clancy: We probably want a change (or copy) the final fatjar artifact to something with a constant name (i.e., `anserini.jar`).

lintool:
Hi @r-clancy - I've given this a bit of thought... why don't we drop this in a separate repo for now:
https://github.com/castorini/SolrAnserini

Conceptually, I think there's a clean separation:
+ The Anserini artifact itself
+ All the bindings that are necessary for working with Solr.

We can assume that both artifacts are cloned in the top-level directory, like

```
+
|-- Anserini
|-- SolrAnserini
```
Can you can reference files with relative paths.

Another argument for separation is that we're not familiar enough with what's ahead to understand how the integration is evolve... if the integration is tight, we can always merge repos. Untangling histories, however, is much more difficult.

Sound good?

r-clancy:
Closing this PR as it's been moved to https://github.com/castorini/SolrAnserini/pull/1

Issue #400, opened 2018-08-12
Anserini tagline: too generic?


lintool:
Currently, the tagline for Anserini says:
```
Anserini: An information retrieval toolkit built on Lucene
```
E.g., this is at the top of the license headers for our documentation.

This seems too generic to me... some better options:

1. A toolkit for information retrieval research built on Lucene
2. A toolkit for reproducible information retrieval research built on Lucene
3. A Lucene-based toolkit for information retrieval research
4. A Lucene-based toolkit for supporting information retrieval research
5. A Lucene-based toolkit for reproducible information retrieval research

Currently, I like the second one the best. Any additional thoughts? Additional suggestions?

Peilin-Yang:
You should make the list as an ordered one....
Vote for number 2

borislin:
+1 for number 2

lintool:
#439 address this. Closing.

PR #439, opened 2018-09-22
tweaked tagline and made sure all files have license headers

lintool:
Ref #400


lintool:
No, just copy and paste by hand.

Peilin-Yang:
Maybe we should consider to have a license header auto-generator?

lintool:
I don't expect to be doing this very often... don't think that's worth the effort...?

PR #451, opened 2018-10-02
Manually construct query instead of using QueryParser.

r-clancy:
@lintool I suspect the same issue with the tests as the Rm3 change (`cw09b` is failing) - I'll re-run the failed regressions once they've finished with more logging and update the regression results if that's the issue.

### DIFF
--- a/src/main/java/io/anserini/rerank/lib/AxiomReranker.java
+++ b/src/main/java/io/anserini/rerank/lib/AxiomReranker.java
@@ -31,7 +31,6 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.lucene.index.*;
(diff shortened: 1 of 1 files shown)


r-clancy:
@lintool The issue here was related to #452 - the query "403b" didn't have any similar terms in the query expansion. Previously the original query was fed into the query parser when this happened, so I've updated the code here to just do a TermQuery (instead of BoostQuery) when no similar terms are found.

The failing regression looks to be passing, I'll run the others over the next day for a sanity check.

Peilin-Yang:
Are the regression tests all passed?

r-clancy:
@Peilin-Yang cw09b was the only failing test, it looks to be good now. I'm running the others just to make sure the change didn't effect them.

lintool:
@r-clancy Did the regression numbers change for cw09b? I don't see a diff of AP here in this PR.

lintool:
Okay, thanks - once you confirm the other regressions pass okay, we'll be good to merge.

PR #453, opened 2018-10-10
Indexing and searching Freebase in Anserini

lintool:
This PR allows you to build an index over Freebase, where each entity is treated as a "document". With the index, we can access Freebase entities by mid or by free text search over labels.


lintool:
@Impavidity @MichaelAzmy

I've made the index available on `tuna`, e.g., this now works:

```
mvn exec:java -Dexec.mainClass="io.anserini.kg.freebase.LookupFreebase" -Dexec.args="-index freebase-index -mid fb:m.02mjmr"
```

PR #456, opened 2018-10-15
 Added scripts for fine tuning Axiomatic and RM3 reranking algorithms

Peilin-Yang:
See `src/main/python/fine_tuning/README.md` for runbook


src/main/resources/fine_tuning/models.yaml
```diff
@@ -0,0 +1,48 @@
+---
+models:
+  axiom:
+    name: axiom
+    fixed_params: "-axiom.n 30 -axiom.deterministic"
+    nexus_params:
```
lukuang: I think we could also have `nexus_params` for rm3?

src/main/resources/fine_tuning/models.yaml
```diff
@@ -0,0 +1,48 @@
...
+        type: int
+        lower: 1
+        upper: 50
+        pace: 1
+      rm3.originalQueryWeight:
+        type: float
+        lower: 0
+        upper: 1
+        pace: 0.1
+    expected:
```
lukuang: What does `expected` mean? Do we want to be more specific?

src/main/python/fine_tuning/evaluation.py
```diff
@@ -0,0 +1,83 @@
...
+                        for line in stdout.split('\n')[1:-1]:
+                            line = line.strip()
+                            if line:
+                                row = line.split(',')
+                                qid = row[-3]
+                                ndcg20 = row[-2]
+                                err20 = row[-1]
+                                o.write('ndcg20\t%s\t%s\n' % (qid if qid != 'amean' else 'all', ndcg20))
+                                o.write('err20\t%s\t%s\n' % (qid if qid != 'amean' else 'all', err20))
+                finally:
```
lukuang: I think we need a `catch` here (maybe IOExcpetion)?

src/main/python/fine_tuning/run_batch.py
```diff
@@ -0,0 +1,211 @@
...
+                if os.path.exists(os.path.join(input_root, yaml_data['index_path'])):
+                    index_path = os.path.join(input_root, yaml_data['index_path'])
+                    break
+    return index_path
+
+def batch_retrieval(collection_yaml, models_yaml, output_root):
+    all_params = []
+    program = os.path.join(collection_yaml['anserini_root'], 'target/appassembler/bin', 'SearchCollection')
+    index_path = get_index_path(collection_yaml)
+    this_output_root = os.path.join(output_root, collection_yaml['name'])
```
lukuang: This part seems to be redundant in all the `batch_` functions. Moving it out to main?

src/main/python/fine_tuning/run_batch.py
```diff
@@ -0,0 +1,211 @@
...
+
+def isclose(a, b, rel_tol=1e-09, abs_tol=0.0):
+    return abs(a-b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol)
+
+def get_index_path(yaml_data):
+    """
+    Find the possible index path
+    """
+    index_path = os.path.join('lucene-index.{0}.pos+docvectors{1}'.format(yaml_data['name'], \
+                                                                          '+rawdocs' if '-storeRawDocs' in yaml_data['index_options'] else ''))
```
lukuang: Indentation

src/main/python/fine_tuning/run_batch.py
```diff
@@ -0,0 +1,211 @@
...
+    input_fns = para[2:]
+    Effectiveness(index_path).output_effectiveness(output_fn, input_fns)
+
+def print_optimal_effectiveness(collection_yaml, models_yaml, output_root, metrics=['map']):
+    index_path = get_index_path(collection_yaml)
+    this_output_root = os.path.join(output_root, collection_yaml['name'])
+    logger.info('='*30+'Optimal Effectiveness for '+collection_yaml['name']+'='*30)
+    effectiveness = Effectiveness(index_path).load_optimal_effectiveness(this_output_root, metrics)
+    print(json.dumps(effectiveness, sort_keys=True, indent=2))
+    return
```
lukuang: Bug here? Since the code below would be dead code

src/main/python/fine_tuning/run_batch.py
```diff
@@ -0,0 +1,211 @@
...
+    success = True
+    for e in effectiveness:
+        expected = models_yaml[e['model']]['expected'][collection_yaml['name']][e['metric']][e['topic']]
+        if isclose(expected, e['actual']):
```
lukuang: There is not key named `actual`. The valid keys are `actual_optimal` and `actual_all`

src/main/resources/fine_tuning/models.yaml
Peilin-Yang: Which parameters should we be linking to?

src/main/resources/fine_tuning/models.yaml
Peilin-Yang: I think we can temporarily remove it - it was used in other regression tests where you put the expected values so that the regression script can compare the regression values with expected ones. If they are not equal the script will fail.

src/main/python/fine_tuning/evaluation.py
Peilin-Yang: yea, should be - although I do not think there will be fatal errors here since the eval program (be it `trec_eval` or `gdeval.pl`) has return success (code 0) state

src/main/python/fine_tuning/run_batch.py
Peilin-Yang: okay

src/main/python/fine_tuning/run_batch.py
Peilin-Yang: yea, this is related to the `expected` so I think we can currently remove.

src/main/resources/fine_tuning/models.yaml
lukuang: `rerankCutoff`？ I think it can also be used for rm3？

src/main/python/fine_tuning/run_batch.py
Peilin-Yang: yea, I will remove the code after the `return`

src/main/resources/fine_tuning/models.yaml
lukuang: `rerankCutoff`？ I think it can also be used for rm3？It would save a bit time?

Peilin-Yang: oh, i see

src/main/python/fine_tuning/run_batch.py
```diff
@@ -0,0 +1,193 @@
...
+logger.addHandler(ch)
+
+parallelism=1
+def batch_everything(all_params, func):
+    if len(all_params) == 0:
+        return
+    p = Pool(parallelism)
+    p.map(func, all_params)
+
+def isclose(a, b, rel_tol=1e-09, abs_tol=0.0):
```
lukuang: Since there is no "expected", we do not need `isclose` function as well?

Peilin-Yang: Let's keep it for now? We probably will use it in the future.

PR #461, opened 2018-10-21
Fix the incorrect implementation of -rm3.fbTerms

Peilin-Yang:
Bug in `FeatureVector.java` `pruneToSize` causes the incorrect expansion terms for RM3.
Still running regression tests....


src/test/java/io/anserini/util/FeatureVectorTest.java
```diff
@@ -0,0 +1,24 @@
...
+    FeatureVector fv = new FeatureVector();
+    fv.addFeatureWeight("a", 0.2f);
+    fv.addFeatureWeight("a", 0.3f);
+    fv.addFeatureWeight("b", 0.3f);
+    fv.addFeatureWeight("b", 0.5f);
+    fv.addFeatureWeight("c", 0.4f);
+  
+    assertEquals(fv.pruneToSize(2).getFeatures().size(), 2);
+    assertEquals(fv.pruneToSize(1).getFeatures().size(), 1);
+    assertEquals(fv.pruneToSize(2).getFeatures(), new HashSet<>(Arrays.asList(new String[]{"a", "b"})));
```
lukuang: When pruning, shouldn't we keep the highest weighted features? The `FeatureVector` needs to be changed for that?

lukuang: > When pruning, shouldn't we keep the highest weighted features? The `FeatureVector` needs to be changed for that?

Ok, the the order is handled by `getOrderedFeatures`, but then the result of the test case should be `"b","c"`, right?

lukuang: Should be good, my bad

PR #462, opened 2018-10-24
Expose the lookupMid and search function of LookupFreebase class

Impavidity:
Instead of print the results in the `lookupMid` and `search` function, returning the value (`Result` or `Document`) is more reasonable. This enables us to use `pyjnius` and get search results in python.


src/main/java/io/anserini/kg/freebase/LookupFreebase.java
```diff
@@ -155,12 +169,28 @@ public static void main(String[] args) throws Exception {
 
     if (searchArgs.mid != null) {
       // Lookup by mid.
-      lookup.lookupMid(searchArgs.mid);
+      Document doc = lookup.lookupMid(searchArgs.mid);
+      if (doc == null)
+        System.exit(-1);
+      doc.forEach(field -> {
```
lintool: There's no collection here, so `forEach` makes no sense?

lintool: nm

PR #465, opened 2018-11-02
Make it possible to use Jelinek-Mercer QL scoring model

tteofili:


src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -186,4 +186,11 @@
 
   @Option(name = "-model", metaVar = "[file]", required = false, usage = "ranklib model file")
   public String model = "";
+
+  @Option(name = "-qlmj", usage = "use Jelinek-Mercer query likelihood scoring model")
+  public boolean qlmj = false;
+
+  @Option(name = "-lambda", metaVar = "[value]", usage = "Jelinek Mercer smoothing parameter")
```
Peilin-Yang: Can we rename it to `qlmj.lambda`?
I know for BM25 and QL the parameters were `b` and `mu` and we should have changed them too. 
But those two were there from the very beginning of Anserini and breaking them will cause problem of regression tests.
If you look at other ranking functions like `pl2` you will see its parameter as `-pl2.c`.

Also, since `qlmj` is a _base_ ranking model could you please move these two options right under `ql`? 

Thanks

lintool:
Hi @tteofili thanks for your contributions! We'd love to know what you're using Anserini for?

src/main/java/io/anserini/search/SearchArgs.java
tteofili: sure, will do

tteofili:
> Hi @tteofili thanks for your contributions! We'd love to know what you're using Anserini for?

research / evaluations around usage of embeddings in IR

src/main/java/io/anserini/search/SearchArgs.java
lintool: Yes, and technically `-ql` should be `-qld` for (Query Likelihood Dirichlet). Such a change will likely break lots of regression testing scripts... :(

But perhaps worth filing an issue?

lintool:
@tteofili We've long discussed integrating word embeddings into Anserini directly... i.e., use a Lucene index as a simple key-value store for lookup up embedding vectors. Is this a feature you'd need? If so, interested in helping us build it out?

src/main/java/io/anserini/search/SearchArgs.java
tteofili: maybe `-qld` can be added, which would work exactly like `-ql` for backward compatibility.

tteofili:
@lintool sure, I would be very much interested.
I have a prototype `Reranker` I'm currently using, that can fetch existing stored embeddings or calculate them on the fly.
The not so nice part is having to choose one linear algebra / DL library unless we work with hundred dimensional extensions of Lucene `PointValues`.

src/main/java/io/anserini/search/SearchArgs.java
Peilin-Yang: This is a good idea.

lintool:
@tteofili opening this thread for discussion https://github.com/castorini/Anserini/issues/467

lintool:
@tteofili were you planning on updating the PR per review comments, or should I close for now?

tteofili:
yes, sure, I'll adjust the PR as per above comments.

lintool:
hi @tteofili I'm going to close this PR for now... this conflicts with a recent change made by @Peilin-Yang that allows much more flexible parameter sweeping. If you still want to contribute code, I think it'll be easier to start from scratch off the current master.

tteofili:
@lintool sure, thanks, it makes sense.

PR #466, opened 2018-11-02
Updates to documentation for JDIQ article

lintool:
JDIQ article published. Updating documentation accordingly.

### DIFF
--- a/src/main/resources/jdiq2018/doc.template
+++ b/src/main/resources/jdiq2018/doc.template
@@ -1,14 +1,19 @@
-### JDIQ2018 Effectiveness Scripts
+# Anserini: JDIQ 2018 Experiments
 
-The scripts calculate the optimal performances of all supported ranking models (by grid-searching all possible model parameters).
-The main purpose is to reproduce what we have in our JDIQ2018 paper:
+These script are used in the following article to compute optimal retrieval effectiveness by grid search over model parameters:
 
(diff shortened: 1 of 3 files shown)


src/main/resources/jdiq2018/doc.template
```diff
@@ -1,14 +1,19 @@
...
 
-_NOTICE: The query topics used in JDIQ paper are combined topics per collection while the numbers generated
-by the scripts here are separated_
+Note that the values produced by these scripts are _slightly_ different than those reported in the article.
+The reason for these differences stem from the fact that Anserini evolved throughout the peer review process; the values reported in the article where those generated when the manuscript was submitted.
+By the time the article was published, the implementation of Anserini has progressed.
+As Anserini continues to improve we will update these scripts, which will lead to further divergences between the published values.
+Unfortunately, this is an unavoidable aspect of empirical research on software artifacts.
+
+## Parameter Tuning
```
Peilin-Yang: Do you want to put the commit sha now or right before we upgrade to Lucene 7?

lintool: I don't think we need a commit sha here, but when we upgrade to Lucene 7 everything is going to change, and we'll document https://github.com/castorini/Anserini/blob/master/docs/regressions-log.md

Sg?

PR #470, opened 2018-11-05
Adding multi-threads support with multiple params to SearchCollection 

Peilin-Yang:
In this PR I fundamentally change how `SearchCollection` works.
1. With this PR one can provide multiple params for base models, for example `-b 0.2 0.75` for BM25 or `-mu 200 2000` for QL. One can also provide multiple params for reranking models, at the same time, for example `-rm3.fbDocs 10 20 -rm3.fbTerms 50 100` for RM3. As a result, there are N1*N2*N3*... run files being generated where N1..Nn are the values of the params.
2. Now the `SearchCollection` spawns new threads for all retrievals by default. This, together with the newly introduced `-inmem` option, make a bunch of retrieval less expensive since multiple retrievals need to load the index (potentially in the memory for better multi-threading) once.


src/main/java/io/anserini/search/similarity/AuxSimilarity.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.anserini.search.similarity;
+
+import org.apache.lucene.search.similarities.Similarity;
+
+public class AuxSimilarity {
```
lintool: How about we name this class `TaggedSimilarity`?

src/main/java/io/anserini/search/similarity/AuxSimilarity.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.anserini.search.similarity;
+
+import org.apache.lucene.search.similarities.Similarity;
+
```
lintool: Add top-level javadoc?

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -369,10 +485,9 @@ public static void main(String[] args) throws Exception {
 
     final long start = System.nanoTime();
     SearchCollection searcher = new SearchCollection(searchArgs);
-    int numTopics = searcher.runTopics();
+    searcher.runTopics();
     searcher.close();
     final long durationMillis = TimeUnit.MILLISECONDS.convert(System.nanoTime() - start, TimeUnit.NANOSECONDS);
-    LOG.info("Total " + numTopics + " topics searched in "
-        + DurationFormatUtils.formatDuration(durationMillis, "HH:mm:ss"));
+    LOG.info("Total run time: " + DurationFormatUtils.formatDuration(durationMillis, "HH:mm:ss"));
```
lintool: Why don't we print out number of topics anymore?

lintool:
High-level comments, for discussion:

1. Instead of depending on `toString`, maybe use explicit `tag` method?
2. Would it make sense to introduce a new abstraction called `RankingModel` that is a combination of the similarity and the rerankers?

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -80,17 +95,79 @@
...
+    final private SortedMap<K, Map<String, String>> topics;
+    final private AuxSimilarity auxSimilarity;
+    final private String cascadeTag;
+    final private RerankerCascade cascade;
+    final private String outputPath;
+    final private String runTag;
+    
+    private SearcherThread(IndexSearcher searcher, SortedMap<K, Map<String, String>> topics, AuxSimilarity auxSimilarity,
+                           String cascadeTag, RerankerCascade cascade, String outputPath, String runTag) throws IOException {
+      this.searcher = searcher;
```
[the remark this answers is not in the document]
Peilin-Yang: Most of the time we will probably have just 1 reranker and I think it is better to have just 1 tag.
Also, we have `TieBreaker` as the default last reranker and I am not sure if it is worthy to concatenate all tags.

src/main/java/io/anserini/search/SearchCollection.java
Peilin-Yang: Because now it is print inside each retrieval thread. 
Search for `LOG.info("Run " + topics.size()` in this PR

Peilin-Yang:
> Instead of depending on toString, maybe use explicit tag method? 
Will do
> Would it make sense to introduce a new abstraction called RankingModel that is a combination of the similarity and the rerankers?
Yes, this makes sense. But let's make it in another PR?

lintool:
Can you create a separate issue for above so we don't lose track of it?

Also, please make sure regressions pass before you merge?

Peilin-Yang:
ALL Regression tests passed on tuna, going to merge

PR #478, opened 2018-11-16
Fine tuning scripts update

Peilin-Yang:


src/main/python/fine_tuning/xfold.py
```diff
@@ -18,7 +18,8 @@
 import logging
 import json
 
-logging.basicConfig()
```
lukuang: You added `logging.basicConfig()` for other files but remove it from this file. Any particular reason?

Peilin-Yang: No idea why it was removed.
Reverted

src/main/resources/fine_tuning/models.yaml
```diff
@@ -102,24 +122,26 @@ models:
...
+          best_avg: 0.0
+          oracles_per_topic: 0.0
+          2-fold: 0.0
+          5-fold: 0.0
+        ndcg20:
+          best_avg: 0.0
+          oracles_per_topic: 0.0
+          2-fold: 0.0
+          5-fold: 0.0
   rm3:
```
lukuang: For `rm3` and `bm25+rm3`, `originalQueryWeight` can be `1.0`. I think we should have that. But then all combinations of other `rm3` parameters will be searched for `originalQueryWeight=1.0`. Then the whole program wastes some time on those? Can we get around it such as support having multiple parameter ranges? Not necessary needs to be changed in this PR but I think we probably should avoid that?

src/main/resources/fine_tuning/models.yaml
```diff
@@ -174,30 +204,20 @@ models:
...
-        pace: 0.1
     expected:
       robust04:
         map:
-          best_avg: 0.2532
-          oracles_per_topic: 0.2921
-          2-fold: 0.2253
-          5-fold: 0.2524
+          best_avg: 0.3020
+          oracles_per_topic: 0.4343
```
lukuang: should be `0.4402`? Please check out the newest version of the paper to updated numbers

src/main/resources/fine_tuning/models.yaml
```diff
@@ -174,30 +204,20 @@ models:
...
+          2-fold: 0.2973
+          5-fold: 0.2956
         P_20:
-          best_avg: 0.3614
-          oracles_per_topic: 0.4426
-          2-fold: 0.3604
-          5-fold: 0.3577
+          best_avg: 0.4012
+          oracles_per_topic: 0.5960
```
lukuang: `0.6054`

src/main/resources/fine_tuning/models.yaml
Peilin-Yang: I will leave it until we merge your PR? (actually you can update it?)

lukuang: I guess that's fine too

Issue #481, opened 2018-11-21
Indexing through symlinks


isoboroff:
This is initial playing, sorry if this is well-known.  I make a robust04 directory and symlinked in my normal locations for the CD45-cr subcollections, but it didn't work.  It seems that the filesystem walker either doesn't work over symlinks or crossing NFS boundaries:

```
$ cat log.robust04.pos+docvectors+rawdocs 
2018-11-21 15:21:56,782 INFO  [main] index.IndexCollection (IndexCollection.java:248) - DocumentCollection path: /Users/soboroff/robust04
2018-11-21 15:21:56,783 INFO  [main] index.IndexCollection (IndexCollection.java:249) - Index path: lucene-index.robust04.pos+docvectors
2018-11-21 15:21:56,783 INFO  [main] index.IndexCollection (IndexCollection.java:250) - CollectionClass: TrecCollection
2018-11-21 15:21:56,783 INFO  [main] index.IndexCollection (IndexCollection.java:251) - Generator: JsoupGenerator
2018-11-21 15:21:56,783 INFO  [main] index.IndexCollection (IndexCollection.java:252) - Threads: 16
2018-11-21 15:21:56,784 INFO  [main] index.IndexCollection (IndexCollection.java:253) - Stemmer: porter
2018-11-21 15:21:56,784 INFO  [main] index.IndexCollection (IndexCollection.java:254) - Keep stopwords? false
2018-11-21 15:21:56,784 INFO  [main] index.IndexCollection (IndexCollection.java:255) - Store positions? true
2018-11-21 15:21:56,784 INFO  [main] index.IndexCollection (IndexCollection.java:256) - Store docvectors? true
2018-11-21 15:21:56,784 INFO  [main] index.IndexCollection (IndexCollection.java:257) - Store transformed docs? false
2018-11-21 15:21:56,784 INFO  [main] index.IndexCollection (IndexCollection.java:258) - Store raw docs? true
2018-11-21 15:21:56,784 INFO  [main] index.IndexCollection (IndexCollection.java:259) - Optimize (merge segments)? false
2018-11-21 15:21:56,784 INFO  [main] index.IndexCollection (IndexCollection.java:260) - Whitelist: null
2018-11-21 15:21:56,799 INFO  [main] index.IndexCollection (IndexCollection.java:291) - Starting indexer...
2018-11-21 15:21:57,117 INFO  [main] index.IndexCollection (IndexCollection.java:314) - 4 files found in /Users/soboroff/robust04
[...]

PR #483, opened 2018-11-23
Adding the support for including symlink files when indexing

Peilin-Yang:


lintool:
Thanks @Peilin-Yang - how is this tested?

@isoboroff do you want to give this a shot and let us know if it works?

src/main/java/io/anserini/collection/DocumentCollection.java
```diff
@@ -108,6 +108,14 @@ public final Path getCollectionPath() {
       public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
         Path name = file.getFileName();
         boolean shouldAdd = true;
+        if (Files.isSymbolicLink(file)) {
```
lintool: Worth adding a comment to reference #481

Peilin-Yang:
@isoboroff 
Could you please try 
```
git fetch
git branch --track issue-481 origin/issue-481
git checkout issue-481
mvn clean package appassembler:assemble
```
and let us know?

Peilin-Yang:
@lintool actually there are somethings:

1. I haven't added the comments that you had suggested
2. I did not consider/test much about what if the symlink points to some files/dirs inside the root path

PR #495, opened 2018-12-07
Freebase indexing schema addition

lintool:
Previous, we can index Freebase with each "node" as a Lucene document. Added ability to index each individual triple as a Lucene document.

### DIFF
--- a/src/main/java/io/anserini/kg/FreebaseNode.java
+++ b/src/main/java/io/anserini/kg/FreebaseNode.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.anserini.kg.freebase;
+package io.anserini.kg;
 
 import org.openrdf.model.Literal;
 import org.openrdf.model.ValueFactory;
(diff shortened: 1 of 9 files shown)


src/main/java/io/anserini/kg/FreebaseNode.java
```diff
@@ -122,11 +122,15 @@ public static String normalizeObjectValue(String objectValue) {
         return removeEnclosingQuote(objectValue);
       }
     } else if (type.equals(FreebaseNode.RdfObjectType.TEXT)) {
-      Literal parsedLiteral = NTriplesUtil.parseLiteral(objectValue, valueFactory);
-      if (parsedLiteral.getLanguage().toString().equals(LANG_EN)) {
-        return parsedLiteral.stringValue();
-      }
-      return "";
+      // Previously, we were basically throwing away all non-English text...
```
Sara-Farazi: Since we're commenting out the logic in this if and just returning object value and the else does the exact same thing, maybe we can comment out this if entirely.

lintool:
> We probably need a different LookupFreebase for this new indexing method?

Indeed. Next PR!

src/main/java/io/anserini/kg/FreebaseNode.java
lintool: Yup, exactly what I had in mind - in my next iteration I'll add the ability to only index English literals.

PR #496, opened 2018-12-08
Fine tuning fix

Peilin-Yang:
All tests passed!

### DIFF
--- a/src/main/python/fine_tuning/run_batch.py
+++ b/src/main/python/fine_tuning/run_batch.py
@@ -53,15 +53,10 @@ def get_index_path(yaml_data):
     """
     Find the possible index path
     """
-    index_path = os.path.join('lucene-index.{0}.pos+docvectors{1}'.format(yaml_data['name'], \
-                                                                          '+rawdocs' if '-storeRawDocs' in yaml_data['index_options'] else ''))
-    if not os.path.exists(index_path):
-        index_path = yaml_data['index_path']
-        if not index_path or not os.path.exists(index_path):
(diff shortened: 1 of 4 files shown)


src/main/python/fine_tuning/run_batch.py
```diff
@@ -188,7 +183,7 @@ def verify_effectiveness(collection_yaml, models_yaml, output_root, use_drr_fold
                         logger.error('!'*5+'base model: %s model: %s fold: %d metric: %s expected: %f actual: %s ' % (basemodel, model, fold, metric, expected['%d-fold' % fold], x_fold_effectiveness[basemodel][model][metric])+'!'*5)
 
     if success_optimal and success_xfold:
-        logger.info("[Regression Tests Passed] Done...")
+        logger.info("[Regression Tests Passed] All Passed^^^")
```
lintool: Why `^^^`?

lintool:
lgtm but @lukuang can you take a look?

src/main/python/fine_tuning/run_batch.py
lukuang: maybe use single quotation marks for this and other `logger.info` calls in this file?

lukuang:
> lgtm but @lukuang can you take a look?

lgtm as well.

src/main/python/fine_tuning/run_batch.py
Peilin-Yang: @lintool no special reasons -- to separate it from `!!!`?

PR #499, opened 2018-12-09
TREC 2018 Runbook/Results Consolidation

Peilin-Yang:
Adding the results for: 

1. CENTRE track results of our runs (note that TREC released the new judgements)
2. Core track results of our runs

Still waiting for News track background linking judgements.


docs/experiments-trec2018.md
```diff
@@ -132,6 +178,30 @@ target/appassembler/bin/SearchCollection -index lucene-index.wash18.pos+docvecto
 target/appassembler/bin/SearchCollection -index lucene-index.wash18.pos+docvectors+rawdocs -topicreader Trec -topics 2018-test-topics.txt -ql -hits 10000 -runtag anserini_qlax17 -axiom -axiom.deterministic -rerankCutoff 0 -axiom.index lucene-index.core17.pos+docvectors+rawdocs -output wash18_ql_ax17.txt
```
lintool: Should we make all parameters explicit on the command line - e.g. b and k_1 for bm25, etc.
So that if we change defaults later (e.g., for RM3), it'll be easier to reconstruct?

lintool:
@Peilin-Yang We're just updating the runbook, but not converting it into a regression that we maintain, correct? This is okay with me, just wanted to check.

This means that people will be able to replicate our runs only if they check out this commit point. After we upgrade to Lucene7, we will no longer be able to replicate on master.

Peilin-Yang:
@lintool yea, I think the purpose of this PR is (trying) to reproduce what we did for TREC 2018 at the point of our submission.

docs/experiments-trec2018.md
Peilin-Yang: I think your suggestion applies more to the general regression test, but not the runbook of TREC?

PR #519, opened 2018-12-16
Add rm3 reranker

Victor0118:
Could @lintool review this PR?

#### QLD_1000
map                   	all	0.2469
recip_rank            	all	0.6719
P_30                  	all	0.3080

#### QLD_1000_RM3
map                   	all	0.2642
recip_rank            	all	0.6398
P_30                  	all	0.3139

#### BM25_0.9_0.4
map                   	all	0.2501
recip_rank            	all	0.6764
P_30                  	all	0.3124

#### BM25_0.9_0.4_RM3
map                   	all	0.2757
recip_rank            	all	0.6803
P_30                  	all	0.3250


src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -71,6 +81,26 @@ public SimpleSearcher(String indexDir) throws IOException {
     this.reader = DirectoryReader.open(FSDirectory.open(indexPath));
     this.similarity = new LMDirichletSimilarity(1000.0f);
     this.analyzer = new EnglishAnalyzer();
+    setNormalReRanker();
+  }
+
+  public void setRM3ReRanker() {
```
lintool: ReRanker -> Reranker

lintool: And below instances also.

lintool:
We can focus only on RM3 now - don't worry about the other models.
Any idea why results are slightly different from: https://github.com/castorini/Anserini/blob/master/docs/experiments-robust04.md

They should be *exactly* the same... can we reconcile now so differences don't creep in later?

src/main/java/io/anserini/search/SimpleSearcher.java
Victor0118: Thanks! Fixed.

Victor0118:
No idea what the exact reason is. We can pay attention to this and fix it in the future.

PR #520, opened 2018-12-18
Add Chinese wikipedia collection in sentence level

Victor0118:
TODO:
1. add paragraph and doc level indexer
2. solve the Chinese segmentation problem
   a. combine different analyzer
   b. processs before indexing


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -103,6 +105,9 @@
 
     @Option(name = "-whitelist", usage = "file containing docids, one per line; only specified docids will be indexed.")
     public String whitelist = null;
+    
+    @Option(name = "-chinese", usage = "boolean to use chinese analyser while indexing")
+    public boolean chinese = false;
```
Peilin-Yang: Change to `chineseAnalyzer`?
And all other occurrences.

Victor0118:
@Peilin-Yang Your comment fixed.

pom.xml
```diff
@@ -220,6 +220,11 @@
       <artifactId>lucene-test-framework</artifactId>
       <version>${LUCENE_VERSION}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.lucene</groupId>
+      <artifactId>lucene-analyzers-smartcn</artifactId>
+      <version>6.3.0</version>
```
Peilin-Yang: change to `${LUCENE_VERSION}`?

Victor0118: I did that but I got a wired error during CI (`${LUCENE_VERSION} is not a version number...`). I can compile that in my local machine. Let me try again.

lintool:
Hi @Victor0118 I'm not sure why you need a new class `WikipediaZHSentenceCollection` - why isn't the existing Json collection sufficient?

Also, Anserini has moved forward quite quickly over the last few days... our upgrade to Lucene 7 changed a lot. It might be easier for you to start over with the new HEAD and reapply your changes?

Victor0118:
@lintool There are some minor reasons for the new collection:

1. The output of wikiextracter is a binary file (and I haven't found the word to generate a text file). So I need FileStream to read it.
2. I index the Wikipedia in sentence level and split a document using `。`. JsonCollection can only do indexing in document level. In the future, we can add the function of indexing Wikipedia-paragraph to this collection (just change a splitter).

You are right. I need to redo the PR.

Issue #517, opened 2018-12-16
Call for records of successful replication


lintool:
Followup, generalizing #492 

@tteofili @catenamatteo @Chriskamphuis I think it'd be nice to keep records of other people who have replicate our results, so that newcomers can get a sense of how reliable our results are. Would you mind doing the following?

For each experiment that you've replicated (e.g., CORE 18), send a PR:
https://github.com/castorini/Anserini/blob/master/src/main/resources/docgen/templates/core18.template

Add something like (to the end of the file):

+ Results replicated by @USERNAME on DATE (commit `xxxxx`)

That way we can acknowledge your contributions, e.g., #514 

Thanks in advance!

tteofili:
I did it already but will re do it now and create a PR as soon as I have the numbers.

lintool:
Thanks all!

tteofili:
I've managed to replicate the results for Trec '17, I'm not sure though which commit should I reference when opening the PR.

PR #522, opened 2018-12-18
add Robust04 replication note

Chriskamphuis:
As requested in #517


Chriskamphuis:
It looks a bit messy, I will create a new one with the older commits not in the log.

Issue #524, opened 2018-12-18
Regressions for ECIR 2019 cross-collection relevance transfer runs


lintool:
We need to build regressions for the forthcoming ECIR paper.

lintool:
Hi @Kytabyte with #532 and #528 the ccrf code is in good shape - can you add in the configs for the ECIR experiments and let's try and get new results?

Thanks!

PR #525, opened 2018-12-18
Initial ccrf (cross-collection relevance feedback) code for ECIR 2019

lintool:
Per #524 - we're going to take our code from TREC 2018 as a starting point and start refactoring to clean up, working our way to building regressions for ECIR 2019.

This initial PR is just @Kytabyte's origin code, with minimal cleanup so that it doesn't rely on a weird directory structure per the runbook.

### DIFF
--- a/src/main/python/ccrf/build_test.py
+++ b/src/main/python/ccrf/build_test.py
@@ -1,0 +1,144 @@
+import argparse
+import logging
+import os
+import pickle
+import time
+
+from scipy.sparse import save_npz, csr_matrix
+import sklearn.preprocessing
+import numpy as np
+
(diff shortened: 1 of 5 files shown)


Kytabyte:
If we need all the code for ECIR 2019, that should be: train on Robust04, 05 and test on core 17. So, the main difference is that topic lists are not the same. Does that okay for now, and we refactor that later?

lintool:
Yes, let's refactor that for later. I want to redo the core runs because we have reference scores wrt previous code so we can be sure we're not breaking anything. Once we've refactored a bit, we can think about how to make the code more general?

Kytabyte:
Okay, sg

PR #528, opened 2018-12-19
Refactoring ccrf code

lintool:
Initial refactoring of ccrf code for ECIR 2019. Since I'm not very familiar with the code base, I'm going to refactoring chunk by chunk. This will no doubt be the first of many CRs.

### DIFF
--- a/src/main/python/ccrf/prepare_test_data.py
+++ b/src/main/python/ccrf/prepare_test_data.py
@@ -3,18 +3,16 @@ import logging
 import os
 import pickle
 import time
-
-from scipy.sparse import save_npz, csr_matrix
+import scipy.sparse
 import sklearn.preprocessing
 import numpy as np
-
-from utils import TfidfTgzReader
+import utils
 
 # Global variables, which we're later going to refactor into an external config file.
 
-topics = ['321', '336', '341', '347', '350', '362', '363', '367', '375', '378',
+topics = {'321', '336', '341', '347', '350', '362', '363', '367', '375', '378',
           '393', '397', '400', '408', '414', '422', '426', '427', '433', '439',
-          '442', '445', '626', '646', '690']
+          '442', '445', '626', '646', '690'}
 
 config = {
     'target': {'name': 'core18',
@@ -44,6 +42,7 @@ def read_vocab(path):
 
     return vocab
 
+
 def build_docid_idx_dict():
     logging.info('building docid idx dict...')
     cur_idx = 0
@@ -52,6 +51,16 @@ def build_docid_idx_dict():
         for line in f:
             topic, _, docid, _, _, _ = line.split(' ')
             if topic in topics and docid not in docid_idx_dict:
+                # docid_idx_dict and vocab_idx_dict actually store two different mappings:
+                # both docid (vocab) -> idx and idx -> docid (vocab).
+                #
+                # The docid -> idx mapping is to transform the docid to the row number of the matrix, and
+                # the idx -> docid mapping is to generate the final submission from the index.
+                # The vocab -> mapping is to transform the word into the column number of the matrix, and
+                # the idx -> vocab mapping is easy for us to check the importance of features.
+                #
+                # The docid and vocab are in string format and indices are integers, so there won't be any conflict.
+                # However, this isn't a good design, and should probably be refactored later.
                 docid_idx_dict[docid] = cur_idx
                 docid_idx_dict[cur_idx] = docid
                 cur_idx += 1
@@ -59,7 +68,7 @@ def build_docid_idx_dict():
 
 
 def write_docid_idx_dict(docid_idx_dict, filename):
-    logging.info(f"writing docid-idx-dict to {filename}")
+    logging.info(f'writing docid-idx-dict to {filename}')
     with open(filename, 'wb') as f:
         pickle.dump(docid_idx_dict, f)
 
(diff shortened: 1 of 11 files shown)


src/main/python/ccrf/prepare_test_data.py
```diff
@@ -10,61 +10,69 @@
...
+def build_docid_idx_dict():
     logging.info('building docid idx dict...')
     cur_idx = 0
     docid_idx_dict = {}
-    with open(rank_file, 'r') as f:
+    with open(config['target']['run'], 'r') as f:
         for line in f:
             topic, _, docid, _, _, _ = line.split(' ')
-            if topic in topic_list and docid not in docid_idx_dict:
+            if topic in topics and docid not in docid_idx_dict:
```
tiddler: Although the list is quite small, considering the judgement will be called for every line, I suggest use a set instead of list.

experiments
```
%%timeit
for topic in topics:
  if topic in topics:
    pass
>>> 100000 loops, best of 3: 14.4 µs per loop

%%timeit
for topic in topics:
  if topic in topics_set:
    pass
>>> 100000 loops, best of 3: 2.41 µs per loop
```

src/main/python/ccrf/prepare_test_data.py
```diff
@@ -10,61 +10,69 @@
...
                 cur_idx += 1
     return docid_idx_dict
 
-def write_docid_idx_dict(docid_idx_dict, filename):
-    """
 
-    """
-    logging.info(f"writting docid-idx-dict to {filename}")
+def write_docid_idx_dict(docid_idx_dict, filename):
+    logging.info(f"writing docid-idx-dict to {filename}")
```
tiddler: quote mark should be consistent, use ' instead.

src/main/python/ccrf/prepare_test_data.py
```diff
@@ -10,61 +10,69 @@
...
-    """
 
-    """
+def dump_docvectors():
+    run = config['target']['run']
+    source_name = config['target']['name']
+    index = config['target']['index']
+    docids_file = os.path.join(working_directory, f'docids.{source_name}')
+    logging.info(f'Extracting docvectors for {source_name} with Anserini')
+    cmd = f'cut -f 3 -d " " {run} | sort | uniq > {docids_file}'
```
tiddler: just be careful, f-format string method only works in 3.6+, if you consider Python 2/3 compatible code later, should use `''.format()` instead.

src/main/python/ccrf/prepare_training_data.py
```diff
@@ -10,34 +10,52 @@
...
-    for rank_file in rank_files:
+    for source in config['sources']:
+        rank_file = source['qrels']
         logging.info(f'building docid idx dict for {rank_file}')
-        with open(os.path.join(rank_file_folder, rank_file), 'r') as f:
+        with open(os.path.join(rank_file), 'r') as f:
             for line in f:
                 topic, _, docid, label = line.strip().split(' ')
-                if topic in topic_list:
+                if topic in topics:
```
tiddler: Same reason here, use `set` and get a speedup for free.

src/main/python/ccrf/prepare_training_data.py
```diff
@@ -51,23 +69,19 @@ def build_docid_idx_and_label(rank_file_folder):
     logging.info(f'{cur_idx} files found in total.')
     return docid_idx_dict, topic_docid_label
 
-def build_vocab_tfidf_dict(tfidf_raw_folder, docid_idx_dict, isalpha):
-    """
 
-    """
+def build_vocab_tfidf_dict(docid_idx_dict, isalpha):
```
tiddler: Our fault, please use is_alpha to keep naming consistency.

src/main/python/ccrf/prepare_test_data.py
```diff
@@ -10,61 +10,69 @@
...
     cur_idx = 0
     docid_idx_dict = {}
-    with open(rank_file, 'r') as f:
+    with open(config['target']['run'], 'r') as f:
         for line in f:
             topic, _, docid, _, _, _ = line.split(' ')
-            if topic in topic_list and docid not in docid_idx_dict:
+            if topic in topics and docid not in docid_idx_dict:
                 docid_idx_dict[docid] = cur_idx
                 docid_idx_dict[cur_idx] = docid
```
Kytabyte: Here the `docid_idx_dict` and the `vocab_idx_dict` actually stores two different mappings: both docid (vocab) -> idx and idx -> docid (vocab). 

The docid -> idx mapping is to transform the docid to the row number of the matrix, and the idx -> docid mapping is to generate the final submission from the index.

The vocab -> mapping is to transform the word into the column number of the matrix, and the idx -> vocab mapping is easy for us to check the importance of features.

The docid and vocab are in string format and indices are integers, so there won't be any conflict.

However, I know this is not a good design, but dumping too many files make things messy as well. Anyway, we probably want to refactor this.

src/main/python/ccrf/prepare_test_data.py
lintool: k

src/main/python/ccrf/prepare_test_data.py
lintool: Thanks for the head's up. We're assuming Python 3.

src/main/python/ccrf/prepare_training_data.py
lintool: k

src/main/python/ccrf/prepare_training_data.py
lintool: k

src/main/python/ccrf/prepare_test_data.py
lintool: added comment to code.

lintool:
@Kytabyte @tiddler thanks for review. lmk if there's anything else.

lintool:
@Kytabyte @tiddler Started refactoring of `train.py` - threw away everything out than lr2, lgb, and svm, since experiments show that ensemble7 isn't actually better than ensemble3.

Script now trains all three models.

PR #533, opened 2018-12-20
More ccrf refactoring

lintool:
Mostly minor tweaks...


tiddler:
In the rerank part, the function comes from Kyle's implementation which simply launches multiple jobs by bash. My implement uses `concurrent.futures.multiprocessing` launch the parallel tasks by mapping the topics. In practices, we compared the performance: bash version will suffer from some classifiers that are really slow, and `concurrent.futures.multiprocessing` is faster.

---

Example:
```python
import concurrent.futures
import math

PRIMES = [
    112272535095293,
    112582705942171,
    112272535095293,
    115280095190773,
    115797848077099,
    1099726899285419]

def is_prime(n):
    if n % 2 == 0:
        return False

    sqrt_n = int(math.floor(math.sqrt(n)))
    for i in range(3, sqrt_n + 1, 2):
        if n % i == 0:
            return False
    return True

def main():
    with concurrent.futures.ProcessPoolExecutor() as executor:
        for number, prime in zip(PRIMES, executor.map(is_prime, PRIMES)):
            print('%d is prime: %s' % (number, prime))
```

lintool:
@tiddler Where do you suggest we apply this? The experiments are bearable in terms of latency right now... runs in a few minutes... This is good to know but I think not low hanging fruit in terms of immediate needs?

Kytabyte:
> In the rerank part, the function comes from Kyle's implementation which simply launches multiple jobs by bash. My implement uses `concurrent.futures.multiprocessing` launch the parallel tasks by mapping the topics. In practices, we compared the performance: bash version will suffer from some classifiers that are really slow, and `concurrent.futures.multiprocessing` is faster.
> 
> Example:
> 
> ```python
> import concurrent.futures
> import math
> 
> PRIMES = [
>     112272535095293,
>     112582705942171,
>     112272535095293,
>     115280095190773,
>     115797848077099,
>     1099726899285419]
> 
> def is_prime(n):
>     if n % 2 == 0:
>         return False
> 
>     sqrt_n = int(math.floor(math.sqrt(n)))
>     for i in range(3, sqrt_n + 1, 2):
>         if n % i == 0:
>             return False
>     return True
> 
> def main():
>     with concurrent.futures.ProcessPoolExecutor() as executor:
>         for number, prime in zip(PRIMES, executor.map(is_prime, PRIMES)):
>             print('%d is prime: %s' % (number, prime))
> ```

Yes, Ruifan's impl was faster than mine, especially when training svm classifier, which takes much longer time to converge than LR and GB tree. However, I think this is the same question, that whether it is worth changing the code structure to speed up depends on whether we would like to take this as a part of regression. If we just would like to keep it as a runbook, it is fine to just keep it as for now.

lintool:
@Kytabyte @tiddler Yea, let's punt on this for now...

PR #535, opened 2018-12-21
Adding ECIR 2019 axiomatic reranking reproducibility runbook

Peilin-Yang:
This is mainly for parameter sensitivity results of `axiom.beta`


lintool:
hi @Peilin-Yang we discussed on Slack, but I still don't think we should check in the duplicate qrels into the repo... why not just add something like this?

```
filenames = ['file1.txt', 'file2.txt', ...]
with open('path/to/output/file', 'w') as outfile:
    for fname in filenames:
        with open(fname) as infile:
            outfile.write(infile.read())
```

ref: https://stackoverflow.com/questions/13613336/python-concatenate-text-files

That is, if the concatenated qrels doesn't exist, create them...

src/main/python/ecir2019_axiomatic/effectiveness.py
```diff
@@ -0,0 +1,94 @@
+# -*- coding: utf-8 -*-
+"""
```
lintool: See: https://github.com/castorini/Anserini/pull/538#issuecomment-452045711

src/main/python/ecir2019_axiomatic/effectiveness.py
```diff
@@ -0,0 +1,94 @@
...
+import json
+import ast
+from operator import itemgetter
+from inspect import currentframe, getframeinfo
+import logging
+
+logging.basicConfig()
+class Effectiveness(object):
+    """
+    Handle the performace. For example, get all the effectiveness of one method(has multiple parameters).
```
lintool: According to PEP 8, the summary should be on the line with `"""`
https://www.python.org/dev/peps/pep-0008/

src/main/python/ecir2019_axiomatic/effectiveness.py
```diff
@@ -0,0 +1,94 @@
...
+        for metric in all_results:
+            with open(os.path.join(output_root, self.effectiveness_root, 'axiom_paras_sensitivity_%s.csv' % metric), 'w') as f:
+                for basemodel in all_results[metric]:
+                    all_results[metric][basemodel].sort(key = itemgetter(0))
+                    for ele in all_results[metric][basemodel]:
+                        f.write('%s,%.1f,%.4f\n' % (basemodel, ele[0], ele[1]))
+
+    def read_eval_file(self, fn):
+        """
+        return {qid: {metric: [(value, para), ...]}}
```
lintool: same as above, and all instances.

PR #536, opened 2018-12-22
Add support for Solr indexing

r-clancy:
Initial support for indexing into Solr.

@lintool I chose to implement a `SolrIndexerThread` rather than mixing in the Solr logic with the existing Lucene indexing code... minor code dupe, but better than mixing responsibilities imo.

I still have a few things I want to test out before merging this, but for `SolrAnserini` you should be able to run it with something like:

```
target/appassembler/bin/IndexCollection -collection NewYorkTimesCollection \
    -generator JsoupGenerator \
    -threads 32 -input NYTcorpus -index lucene-index.core17.pos+docvectors+rawdocs \
    -storePositions -storeDocvectors -storeRawDocs \
    -solr -solr.batch 5000 -solr.index core17 -solr.url localhost:8983
```

I've tested with `core17` (just because it's quick to run), and indexing to the 9 Solr nodes takes a little over 4.5 mins on `himrod` whereas indexing to disk normally took just under 7.5 mins.


pom.xml
```diff
@@ -299,5 +299,10 @@
       <artifactId>trec-car-tools-java</artifactId>
       <version>13</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.solr</groupId>
+      <artifactId>solr-solrj</artifactId>
+      <version>${lucene.version}</version>
+    </dependency>
   </dependencies>
```
lintool: Move up to group with the rest of the Lucene dependencies?

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -89,13 +90,13 @@
...
+
     @Option(name = "-stemmer", usage = "Stemmer: one of the following porter,krovetz,none. Default porter")
     public String stemmer = "porter";
 
     @Option(name = "-uniqueDocid", usage = "remove duplicated documents with the same doc id when indexing. " +
-      "please note that this option may slow the indexing a lot and if you are sure there is no " +
-      "duplicated document ids in the corpus you shouldn't use this option.")
+        "please note that this option may slow the indexing a lot and if you are sure there is no " +
+        "duplicated document ids in the corpus you shouldn't use this option.")
     public boolean uniqueDocid = false;
```
lintool: this diff shouldn't be here?

lintool: Oh, hrm, is this auto code formatting?

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -233,6 +253,131 @@ public void run() {
...
+        }
+      } else {
+        this.solrClient = new ConcurrentUpdateSolrClient.Builder(args.solrUrl).withQueueSize(args.solrBatch).build();
+      }
+    }
+
+    @Override
+    public void run() {
+
+      try {
```
lintool: Hrm, I think there are more whitespaces in this method then there needs be?

lintool:
@r-clancy I want to let @Peilin-Yang chime in on this also.

@Peilin-Yang The idea here is that we want to use the Anserini "frontend" to be able to index directly into a Solr instance if desired.

> I've tested with core17 (just because it's quick to run), and indexing to the 9 Solr nodes takes a little over 4.5 mins on himrod whereas indexing to disk normally took just under 7.5 mins.

"Normally on disk" is on tuna or on console on a himrod node?
If not can we do this fair apples-to-apples comparison?

r-clancy:
> "Normally on disk" is on tuna or on console on a himrod node?
> If not can we do this fair apples-to-apples comparison?

Both tests were run from `himrod-10`:
- The Solr test was done pointing to the ZooKeeper for the SolrCloud install (on `himrod-11`)
- The Lucene test was done writing to local disk (on `himrod-10`)

src/main/java/io/anserini/index/IndexCollection.java
r-clancy: Yeah, auto code formatting - I can remove it, but it's standard Intellij indenting.

pom.xml
r-clancy: :+1:

src/main/java/io/anserini/index/IndexCollection.java
lintool: nah that's fine, let's go with IntelliJ

r-clancy:
> Does messing with the batch size make a difference?

Batches of 1k and 10k seem to be a bit slower (about 30 sec), 5k seems to be the sweet spot. I'll test with a few other different sizes though.

lintool:
@r-clancy Few more comments for you...
@Peilin-Yang In case you have any more comments...

PR #538, opened 2018-12-22
Analysis script to compare two runs

lintool:
@Peilin-Yang @Kytabyte @Victor0118 I want to start standardizing the way we do evaluation and analysis... here's a start.

@Kytabyte @Victor0118 can you show me how you create those better/worse per-topic bar charts that compare two runs?

Post code snippet here, and I'll try to synthesize something that looks reasonable - moving forward, we can use the same script so figures looks consistent in all our papers.

### DIFF
--- a/src/main/python/compare_runs.py
+++ b/src/main/python/compare_runs.py
@@ -1,0 +1,58 @@
+import argparse
+import os
+import numpy as np
+import scipy.stats
+import statistics
+
+def load_metrics(file):
+    metrics = {}
+    with open(file, 'r') as f:
+        for line in f:
+            metric, qid, score = line.split('\t')
+            metric = metric.strip()
+            qid = qid.strip()
+            score = score.strip()
+            if qid == 'all':
+                continue
+            if metric not in metrics:
+                metrics[metric] = {}
+            metrics[metric][qid] = float(score)
+
+    return metrics
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--base", type=str, help='base run', required=True)
+    parser.add_argument("--comparison", type=str, help='comparison run', required=True)
+    parser.add_argument("--qrels", type=str, help='qrels', required=True)
+    parser.add_argument("--metric", type=str, help='metric', required=True)
+
+    args = parser.parse_args()
+    base = args.base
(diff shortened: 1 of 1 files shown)


Victor0118:
I did the visualization for `diff` before.
https://github.com/Victor0118/IR_tool/blob/master/per-topic-analysis/per-topic-analysis.ipynb.

For T-test I use the script from Mark D. Smucker.
https://github.com/Victor0118/IR_tool/blob/master/significance_test/paired-randomization-test-v2.pl

src/main/python/compare_runs.py
```diff
@@ -0,0 +1,58 @@
...
+    comp = args.comparison
+    qrels = args.qrels
+
+    os.system(f'eval/trec_eval.9.0.4/trec_eval -q -M1000 {qrels} {base} > eval.base')
+    os.system(f'eval/trec_eval.9.0.4/trec_eval -q -M1000 {qrels} {comp} > eval.comp')
+
+    base_metrics = load_metrics('eval.base')
+    comp_metrics = load_metrics('eval.comp')
+
+    metric = 'map'
```
Victor0118: This is test code?

lintool: ah, good catch. Will fix.

lintool:
@Victor0118 hrm... instead of shelling out to `gnuplot`, it might be good to use ggplot instead?

@Peilin-Yang or @Kytabyte do you have code to share?

Peilin-Yang:
@lintool Add high level pydoc explaining what this script is?

lintool:
@Victor0118 @Peilin-Yang Addressed comments. Added docstring as high-level documentation. License header using '#' prefix.

lintool:
BTW, I did a bit of research. There doesn't seem to be a standard way of putting license information in a Python script, see: https://stackoverflow.com/questions/2048874/writing-copyright-information-in-python-code

*But*, I don't think the license information should be a docstring, as we currently have... https://github.com/castorini/Anserini/blob/master/src/main/python/run_regression.py

So I think a good convention is:

```
# License info
# License info
# License info
# ...

"""Docstring summary.

More docstring...
"""
```

Victor0118:
LGTM.

PR #543, opened 2019-01-04
Refactor simplesearcher to make RM3 work on MB corpus and results of simplesearcher consistent with searchCollection

Victor0118:
Add a field for tweet search to make sure it works with RM3. 

See: https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/rerank/lib/Rm3Reranker.java#L109

@lintool @Peilin-Yang Could you take a look at this PR?

### DIFF
--- a/src/main/java/io/anserini/search/SimpleSearcher.java
+++ b/src/main/java/io/anserini/search/SimpleSearcher.java
@@ -92,6 +92,7 @@ public class SimpleSearcher implements Closeable {
     this.reader = DirectoryReader.open(FSDirectory.open(indexPath));
     this.similarity = new LMDirichletSimilarity(1000.0f);
     this.analyzer = new EnglishAnalyzer();
+    this.searchtweets = false;
     this.isRerank = false;
     setDefaultReranker();
   }
@@ -173,6 +174,7 @@ public class SimpleSearcher implements Closeable {
     searchArgs.arbitraryScoreTieBreak = false;
     searchArgs.hits = k;
     searchArgs.searchtweets = searchtweets;
+
     TopDocs rs = new TopDocs(0, new ScoreDoc[]{}, Float.NaN);
     RerankerContext context;
     if (searchtweets) {
@@ -218,7 +220,7 @@ public class SimpleSearcher implements Closeable {
       return null;
     }
 
-    IndexableField field = doc.getField(LuceneDocumentGenerator.FIELD_RAW);
+    IndexableField field = doc.getField(FIELD_RAW);
     return field == null ? null : field.stringValue();
   }
 }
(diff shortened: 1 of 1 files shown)


src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -82,6 +84,16 @@ public SimpleSearcher(String indexDir) throws IOException {
     setNormalReranker();
```
lintool: this seems to be old code, but can re rename "Default" as opposed to "Normal"?

src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -82,6 +84,16 @@ public SimpleSearcher(String indexDir) throws IOException {
...
   }
 
+  public void setSearchTweets(boolean flag) {
+     this.searchtweets = flag;
+     if (flag) {
+        this.analyzer = new TweetAnalyzer();
+      }
+     else {
```
lintool: else on line above.

```
} else {
```

src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -82,6 +83,10 @@ public SimpleSearcher(String indexDir) throws IOException {
     setNormalReranker();
   }
 
+  public void setSearchTweets(boolean flag) {
```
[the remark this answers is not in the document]
lintool: I think the point here is for this class to be used from Python?

Victor0118:
@lintool @Peilin-Yang The refactor is done!

src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -139,24 +161,50 @@ public void close() throws IOException {
...
     SearchArgs searchArgs = new SearchArgs();
     searchArgs.arbitraryScoreTieBreak = false;
     searchArgs.hits = k;
-    RerankerContext context = new RerankerContext<>(searcher, null, query, null, q, queryTokens, null, searchArgs);
+    searchArgs.searchtweets = searchtweets;
+
+    TopDocs rs = new TopDocs(0, new ScoreDoc[]{}, Float.NaN);
+    RerankerContext context;
+    if (searchtweets) {
+		if (t > 0) {
```
lintool: fix two-space indentation here...

src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -139,24 +161,50 @@ public void close() throws IOException {
...
     ScoredDocuments hits = cascade.run(ScoredDocuments.fromTopDocs(rs, searcher), context);
 
     Result[] results = new Result[hits.ids.length];
     for (int i = 0; i < hits.ids.length; i++) {
       Document doc = hits.documents[i];
-      String docid = doc.getField(LuceneDocumentGenerator.FIELD_ID).stringValue();
-      IndexableField field = doc.getField(LuceneDocumentGenerator.FIELD_RAW);
+      String docid = doc.getField(FIELD_ID).stringValue();
+      IndexableField field = doc.getField(FIELD_RAW);
       String content = field == null ? null : field.stringValue();
```
lintool: Switch back to `LuceneDocumentGenerator.FIELD_RAW` - don't do static imports.

lintool:
Ref: https://github.com/castorini/Anserini/pull/519#issuecomment-447676537
According to @Victor0118 we addressed the issue referenced above ^^^^

Victor0118:
@lintool You comments resolved.

PR #544, opened 2019-01-11
ConcurrentUpdateSolrClient should be using multiple threads to drain document queue

lintool:
@r-clancy I think you missed this minor detail :)


r-clancy:
:ship:

PR #545, opened 2019-01-11
Implement object pooling for SolrClient...

r-clancy:
and also ensure only one of -index or -solr is passed.


r-clancy:
@lintool I want to do a bit more testing before this is merged, but take a look. Utilization seems -much- better than the previous implementation, I'll run some more experiments later.

PR #568, opened 2019-01-27
updated replication log for TREC 17

tteofili:
I've updated replication log for TREC 17, I got the very same results described at:
https://github.com/castorini/Anserini/blob/master/docs/experiments-core17.md

### DIFF
--- a/src/main/resources/docgen/templates/core17.template
+++ b/src/main/resources/docgen/templates/core17.template
@@ -40,4 +40,4 @@ ${effectiveness}
 
 ## Replicate Log
 
-* Results replicated by [@tteofili](https://github.com/tteofili) on 2019-01-27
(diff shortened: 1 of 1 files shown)


lintool:
Hi @tteofili thanks for this!

Can we keep to the same formatting as here for consistency?
https://github.com/castorini/Anserini/blob/master/docs/experiments-core18.md

For example, we want to keep track of the actual commit id you replicated...

tteofili:
@lintool I've added the commit id to the template.

tteofili:
thanks @lintool for merging this PR.
as a side note, I've also run experiments for `f2exp` and `f2log` models, which give better `MAP` than `bm25` and `ql` but still lower `P30` than `ql`

MAP        | BM25   | QL     | F2LOG  | F2EXP  |
:----------|--------|--------|--------|--------|
All Topics | 0.1977 | 0.1913 | 0.2003 | 0.1981 |

P30        | BM25   | QL     | F2LOG  | F2EXP  |
:----------|--------|--------|--------|--------|
All Topics | 0.4160 | 0.4373 | 0.4167 | 0.4133 |

PR #571, opened 2019-01-31
Small changes

tokee:
A few minor issues I came across when looking through the code.

### DIFF
--- a/src/main/java/io/anserini/index/IndexCollection.java
+++ b/src/main/java/io/anserini/index/IndexCollection.java
@@ -309,9 +309,9 @@ public final class IndexCollection {
 
           SolrInputDocument solrDocument = new SolrInputDocument();
 
-          // Add all STORED fields
+          // Copy all Lucene Document fields to Solr document
           for (IndexableField field : document.getFields()) {
-            if (field.fieldType().stored()) {
+            if (field.stringValue() != null) { // For some reason, id is multi-valued with null as one of the values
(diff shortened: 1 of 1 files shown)


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -536,7 +537,10 @@ public void run() throws IOException, InterruptedException {
     // Do a final commit
     if (args.solr) {
       try {
-        solrPool.borrowObject().commit(args.solrIndex);
+        SolrClient client = solrPool.borrowObject();
+        client.commit(args.solrIndex);
```
r-clancy: @tokee Is this any different from calling `solrPool.borrowObject().commit(args.solrIndex, false, false)`, where the two `false` are the `waitFlush` and `waitSearcher` parameters?

tokee: The point was to let the `solrPool` get the `SolrClient` back after use. If it does not, the pool treats the `SolrClient` as a still-borrowed object and refuses to close until the timeout of 1 minute has been reached.

Whether or not the `SolrClient` should wait for the Solr-commit to finish is independent of this. I have no suggestions there: It all depends on whether the user expects to see all the changes visible immediately after Anserini has finished or if it is fine with the given `commitWithin`-timeframe.

r-clancy: Ahh, gotcha - I missed the point of the change.

PR #574, opened 2019-02-06
Solr threads and exceptions

tokee:
Each `ConcurrentUpdateSolrClient` has an internal executor. When creating these clients, each client had their maximum thread count set to the overall thread count for Anserini. As each processing thread has their own `ConcurrentUpdateSolrClient`, this potentially meant `args.threads*args.threads` for the total amount of `ConcurrentUpdateSolrClient`s. Local experiments with `args.threads=40` resulted in exceptions being thrown due to this. This pull request makes the `ConcurrentUpdateSolrClient` maxThreads adjustable, with a default of 1.

Not related, exceptions from Solr document delivery using `ConcurrentUpdateSolrClient` were logged to some unknown log. This pull request redirects those to the `IndexCollection`-log, so that the pattern matches the Lucene code path.

### DIFF
--- a/src/main/java/io/anserini/index/IndexCollection.java
+++ b/src/main/java/io/anserini/index/IndexCollection.java
@@ -457,10 +457,20 @@ public final class IndexCollection {
         }
       }
       // Standlone
-      return new ConcurrentUpdateSolrClient.Builder(args.solrUrl)
+      ConcurrentUpdateSolrClient.Builder builder = new ConcurrentUpdateSolrClient.Builder(args.solrUrl)
           .withQueueSize(args.solrBatch)
-          .withThreadCount(args.solrClientThreads)
-          .build();
+          .withThreadCount(args.solrClientThreads);
+      return new ExceptionHandlingSolrClient(builder);
+    }
+    private class ExceptionHandlingSolrClient extends ConcurrentUpdateSolrClient {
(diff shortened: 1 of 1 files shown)


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -148,6 +148,12 @@
 
     @Option(name = "-solr.zkChroot", usage = "the ZooKeeper chroot, if using a ZooKeeper URL instead of Solr")
     public String solrZkChroot = null;
+
+    // Note: This is used for ConcurrentSolrClient, where each processing thread has its own instance
+    // If the architecture is changed to share one ConcurrentSolrClient among all threads, the default should likely be increased
+    @Option(name = "-solr.client.threads", metaVar = "[Number]", required = false, usage = "Number of Threads for each SolrClient")
+    public int solrClientThreads = 1;
```
r-clancy: Can we bump the default on this up to 16? https://github.com/castorini/Anserini/pull/573 will take us back to a single `ConcurrentSolrClient` among all threads.

tokee: Seems like we're crossing streams with pull requests here. Yes, the default should be increased if the client is shared. 16 is not a bad value: Most Solr installations should handle that without problems.

r-clancy: Yeah... might be easier if my PR is merged first than this one on top of it - fewer changes here.

PR #576, opened 2019-02-11
Buffer Documents for ConcurrentUpdateSolrClient

r-clancy:
(Re-)add object pooling for SolrClients... it helped throughput quite a bit for `CloudSolrClient` even though it's thread-safe.

Also, buffer docs for the `ConcurrentUpdateSolrClient` as it's queue is for update requests (i.e., it buffers `add(SolrInputDocument)` calls until the queue size is reached then drains the queue concurrently).

### DIFF
--- a/src/main/java/io/anserini/index/IndexCollection.java
+++ b/src/main/java/io/anserini/index/IndexCollection.java
@@ -153,14 +153,14 @@ public final class IndexCollection {
     /**
      * The number of {@link SolrClient#add(Collection) add} calls to buffer before draining pending update queue.
      */
-    @Option(name = "-solr.client.queue.size", usage = "the number of update requests to buffer before draining")
+    @Option(name = "-solr.clientQueueSize", usage = "the number of update requests to buffer before draining")
     public int solrClientQueueSize = 16;
 
     /**
      * The number threads used to drain the ConcurrentUpdateSolrClient queue
      */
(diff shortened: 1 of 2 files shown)


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -144,10 +150,23 @@
...
 
-    // Note: This is used for ConcurrentSolrClient, where each processing thread has its own instance
-    // If the architecture is changed to share one ConcurrentSolrClient among all threads, the default should likely be increased
-    @Option(name = "-solr.client.threads", metaVar = "[Number]", required = false, usage = "Number of Threads for each SolrClient")
-    public int solrClientThreads = 16;
+    /**
+     * The number of {@link SolrClient#add(Collection) add} calls to buffer before draining pending update queue.
+     */
+    @Option(name = "-solr.client.queue.size", usage = "the number of update requests to buffer before draining")
+    public int solrClientQueueSize = 16;
```
lintool: Should we use standard Java camelCasing here? e.g., `solr.clientQueueSize`? `solr.` is fine I think because that's how we've grouped the options...

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -144,10 +150,23 @@
...
+
+    /**
+     * The number threads used to drain the ConcurrentUpdateSolrClient queue
+     */
+    @Option(name = "-solr.client.thread.count", metaVar = "[NUMBER]", usage = "the number of threads used to drain the ConcurrentUpdateSolrClient queue")
```
lintool: `solr.clientThreads`?

PR #585, opened 2019-03-08
Add robust04 data generation code for reranking

Victor0118:
@lintool Could you take a look at this PR?

### DIFF
--- a/src/main/python/rerank/scripts/export_robust04_dataset.py
+++ b/src/main/python/rerank/scripts/export_robust04_dataset.py
@@ -23,18 +23,20 @@ args = parser.parse_args()
 searcher = JSearcher(JString(args.index))
 
 fqrel = "src/main/resources/topics-and-qrels/qrels.robust2004.txt"
+qid2reldocids = get_qid2reldocids(fqrel)
 best_rm3_parameters = [[47, 9, 0.3], [47, 9, 0.3], [47, 9, 0.3], [47, 9, 0.3], [26, 8, 0.3]]
 for split in range(1, 6):
     ftrain = json.load(open("src/main/resources/fine_tuning/drr_folds/rob04.train.s{}.json".format(split)))
     fdev = json.load(open("src/main/resources/fine_tuning/drr_folds/rob04.dev.s{}.json".format(split)))
     ftest = json.load(open("src/main/resources/fine_tuning/drr_folds/rob04.test.s{}.json".format(split)))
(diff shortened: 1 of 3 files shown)


src/main/python/rerank/scripts/export_robust04_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+            method = "BM25_0.9_0.5"
+        else:
+            print("Unsupported ranking method")
+            break 
+        prediction_fn = "predict_{}_robust04_split{}_{}.txt".format(method, split, mode)
+        output_dir = os.path.join(args.output_dir, "Robust04Corpus")
+        output_fn = os.path.join(output_dir, "split{}_{}_{}.txt".format(split, mode, method))
+        if not os.path.exists(output_dir):
+            os.makedirs(output_dir)
+            searcher.setBM25Similarity(0.9, 0.5)
```
lintool: I think BM25 defaults are 0.9, 0.4: https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/search/SearchArgs.java#L116

We should match defaults?

src/main/python/rerank/scripts/export_robust04_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+parser.add_argument('--output_dir', default='src/main/python/rerank/data', help='output directory')
+parser.add_argument('--K', default=1000, type=int, help='number of document retrieved')
+parser.add_argument('--index', default="/tuna1/indexes/lucene-index.robust04.pos+docvectors+rawdocs", help='index of Robust04 corpus')
+args = parser.parse_args()
+
+searcher = JSearcher(JString(args.index))
+
+fqrel = "src/main/resources/topics-and-qrels/qrels.robust2004.txt"
+qid2reldocids = get_qid2reldocids(fqrel)
+best_rm3_parameters = [[47, 9, 0.3], [47, 9, 0.3], [47, 9, 0.3], [47, 9, 0.3], [26, 8, 0.3]]
```
lintool: Add a comment here about these being the parameters for five-fold from SIGIR Forum 2018 paper.

Victor0118: Okay.

src/main/python/rerank/scripts/export_robust04_dataset.py
Victor0118: Right. Thanks!

PR #589, opened 2019-04-09
add interpolation code

Victor0118:
@lintool Could you review this code?

### DIFF
--- a/src/main/python/rerank/scripts/interpolate.py
+++ b/src/main/python/rerank/scripts/interpolate.py
@@ -1,5 +1,15 @@
 '''scripts to aggregate the retrieval scores and the reranking scores by linear interpolation
 
+This scipt requires five input files:
+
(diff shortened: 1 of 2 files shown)


lintool:
Can you write some documentation, both in the file and in this PR, on what this code is used for?

Victor0118:
This is script to aggregate the retrieval scores and the reranking scores by linear interpolation.

The hyper-parameter \lamda is tuned on the dev set.

src/main/python/rerank/scripts/interpolate.py
```diff
@@ -0,0 +1,123 @@
+'''scripts to aggregate the retrieval scores and the reranking scores by linear interpolation
+
```
lintool: Please add a note here to say the input format... Is this standard TREC output format?
What's the output of this script? A ranking in the same format?

PR #593, opened 2019-04-18
Only store paragraph text in content field

r-clancy:
When using `-storeTransformedDocs`, the `contents` field had some malformed text (title wrapped with `Optional[]` and kicker) - the content should be the body text of the document.

### DIFF
--- a/src/main/java/io/anserini/index/generator/WapoGenerator.java
+++ b/src/main/java/io/anserini/index/generator/WapoGenerator.java
@@ -39,7 +39,7 @@ public class WapoGenerator extends LuceneDocumentGenerator<WashingtonPostCollect
   public static final String FIELD_RAW = "raw";
   public static final String FIELD_BODY = "contents";
   public static final String FIELD_ID = "id";
-  
+
   private static final String PATTERN = "<.+>";
   public static final List<String> CONTENT_TYPE_TAG = Arrays.asList("sanitized_html", "tweet");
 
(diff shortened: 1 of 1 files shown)


lintool:
Can you describe what we're currently storing and what we're changing it to? Perhaps an example?

I'm thinking we should actually store as much information as possible, and then strip it out downstream? Otherwise, wouldn't be potentially losing information for other use cases?

r-clancy:
Currently, we store something like this:
```
Optional[Bubbly planet Venus starts off New Year]\n\nSet down your champagne and gaze west, as the bubbly planet Venus kicks off Sunday night’s first evening of the New Year.\nSee this splendid planet...
```

The `title` is present before the content (and is wrapped in an optional, but that's beside the point). The `title` is stored separately in another field called `title` so it's duplicate information. There are some other fields (`kicker`, `fullCaption`, etc.) that follow a similar pattern - stuck in with the text, but available in another field.

and the new version stores only the text:
```
Set down your champagne and gaze west, as the bubbly planet Venus kicks off Sunday night’s first evening of the New Year. See this splendid planet...
```

There's also a problem with some extractions with all the newlines. For example, the same doc has `\nJan. 14 — Stargazing at the National...` extractions on the first version miss `Jan 14` as a date, but the second version (where we just append paragraphs with a space), catches this.

Thoughts? Anything we should change?

lintool:
Should we also store title?

Or, as an alternative, should we just store the entire raw document (JSON string, right?)... and let downstream apps parse out the JSON and do whatever it wants?

r-clancy:
`title` is already stored in a different field (above was just for the `contents` field).

I believe when using `-storeRawDocs` it stores the entire JSON string - imo the `-storeTransformedDocs` (as above) should just store the text.

lintool:
I see. Gotcha.

lgtm then.

PR #600, opened 2019-04-23
Add Solr configuration and install script

r-clancy:
This PRs add the Solr configsets for Anserini (normal and twitter) and provides an install script.


lintool:
1. All the stopwords are standard right? If so, can we just pack in a tarball, and extract into the right location? I'll make our directory structure look a lot cleaner. (okay for XML configs to be in plain text, so we can track diffs if we tweak.)
2. Can we assume Solr's already been downloaded and packed? I.e., don't download in script - I believe the common case is to re-index a collection into Solr... and I only want to do this once.

r-clancy:
> 1. All the stopwords are standard right? If so, can we just pack in a tarball, and extract into the right location? I'll make our directory structure look a lot cleaner. (okay for XML configs to be in plain text, so we can track diffs if we tweak.)

Yep - I've archived them and made the script extract them.

> 2. Can we assume Solr's already been downloaded and packed? I.e., don't download in script - I believe the common case is to re-index a collection into Solr... and I only want to do this once.

Sure... I think we make the assumption that Solr is already running and this script just bootstraps an existing Solr with the relevant anserini bits, seem reasonable?

lintool:
> Sure... I think we can make the assumption that a single node SolrCloud is already running locally and this script just bootstraps it with the relevant anserini bits, seem reasonable?

+1

src/main/resources/solr/solr.sh
```diff
@@ -19,10 +17,6 @@ mkdir ${SOLR_DIR}/lib && cp ././../../../../target/anserini-*-fatjar.jar ${SOLR_
 unzip anserini/conf/lang.zip -d anserini/conf
```
lintool: You probably don't mean `././`?

PR #601, opened 2019-04-23
Add Solrini documentation

r-clancy:


r-clancy:
@lintool Should be good to go with the instructions [here](https://github.com/castorini/anserini/blob/solr-config/docs/solrini.md).

PR #603, opened 2019-04-25
Refactoring of MS MACRO retrieve script

lintool:
- changed to argparse
- update to PEP8

@rodrigonogueira4


rodrigonogueira4:
Thanks for this refactoring!

PR #614, opened 2019-04-30
Added multifield indexing with JSON collection

lintool:
This was implemented for MS MARCO doc2query experiments. Doesn't appear to help, but this code should be useful for other applications as well, so good to commit to repo.

### DIFF
--- a/src/main/java/io/anserini/search/SimpleSearcher.java
+++ b/src/main/java/io/anserini/search/SimpleSearcher.java
@@ -208,6 +208,48 @@ public class SimpleSearcher implements Closeable {
     return results;
   }
 
+  // searching both the defaults contents fields and another field with weight boost
+  // this is used for MS MACRO experiments with query expansion.
+  public Result[] searchFields(String q, String f, float boost, int k) throws IOException {
+    IndexSearcher searcher = new IndexSearcher(reader);
+    searcher.setSimilarity(similarity);
+
+    Query queryContents = new BagOfWordsQueryGenerator().buildQuery(LuceneDocumentGenerator.FIELD_BODY, analyzer, q);
+    Query queryField = new BagOfWordsQueryGenerator().buildQuery(f, analyzer, q);
+    BooleanQuery query = new BooleanQuery.Builder()
+        .add(queryContents, BooleanClause.Occur.SHOULD)
+        .add(new BoostQuery(queryField, boost), BooleanClause.Occur.SHOULD).build();
+
+    List<String> queryTokens = AnalyzerUtils.tokenize(analyzer, q);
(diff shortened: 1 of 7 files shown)


src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -208,6 +208,48 @@ public void close() throws IOException {
...
+  public Result[] searchFields(String q, String f, float boost, int k) throws IOException {
+    IndexSearcher searcher = new IndexSearcher(reader);
+    searcher.setSimilarity(similarity);
+
+    Query queryContents = new BagOfWordsQueryGenerator().buildQuery(LuceneDocumentGenerator.FIELD_BODY, analyzer, q);
+    Query queryField = new BagOfWordsQueryGenerator().buildQuery(f, analyzer, q);
+    BooleanQuery query = new BooleanQuery.Builder()
+        .add(queryContents, BooleanClause.Occur.SHOULD)
+        .add(new BoostQuery(queryField, boost), BooleanClause.Occur.SHOULD).build();
+
```
Victor0118: So lines below are the same with those in the function `search()`, right? Packing them into one function would be better?

lintool: Do you mean an overloaded method `search`?

src/test/java/io/anserini/collection/JsonDocumentObjectTest.java
```diff
@@ -23,7 +23,12 @@
 import java.util.HashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
-
+// A file in a JsonCollection can either be:
+// (1) A single JSON object (i.e., a single document)
+// (2) An array of JSON objects
+// (3) JSON Lines (i.e., one JSON object per line)
```
Victor0118: IIUC, JsonCollection supports a line/lines of:
(1). JSON object
(2). JSON array

So it supports JSON array lines beside your list, right?

lintool: > So it supports JSON array lines beside your list, right?

What do you mean by this?

JSON Lines is JSONL: http://jsonlines.org/

Victor0118: I see. Then make sense. No need to modify.

src/main/java/io/anserini/search/SimpleSearcher.java
Victor0118: I mean wrapping line 225 to line 250 into one function and call it in both search() and searchFields().

PR #621, opened 2019-05-01
Fix tweaks for Windows compatibility

emmileaf:
In PR [#620](https://github.com/castorini/anserini/pull/620) I replaced `:` with `=` in `SearchCollection` at the file writer. 
https://github.com/castorini/anserini/blob/70bf6022c06948f642009f657094a54fd52790cf/src/main/java/io/anserini/search/SearchCollection.java#L137
This might cause issues when the following code checks for whether file already exists. 
https://github.com/castorini/anserini/blob/70bf6022c06948f642009f657094a54fd52790cf/src/main/java/io/anserini/search/SearchCollection.java#L323-L326

Submitting a fix to replace the character during construction of similarity tags instead.


lintool:
@emmileaf in the future add @r-clancy also to speed reviews...

PR #635, opened 2019-05-09
Checksum Change

edwardhdlu:
Just to check - I get a different checksum for `collectionandqueries.tar.gz` but end up with the same baseline results.

### DIFF
--- a/docs/experiments-msmarco.md
+++ b/docs/experiments-msmarco.md
@@ -12,7 +12,7 @@ wget https://msmarco.blob.core.windows.net/msmarcoranking/collectionandqueries.t
 tar -xvf ${DATA_DIR}/collectionandqueries.tar.gz -C ${DATA_DIR}
 ```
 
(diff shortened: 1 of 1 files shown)


lintool:
Woah. This is odd. I just redownloaded the tarball and confirm @edwardhdlu 's checksum.
I also have my tarball from previous, which confirms the checksum in the doc.

So the tarball's changed recently?!

@Victor0118 @rodrigonogueira4 do you know anything about this?

Victor0118:
I think @rodrigonogueira4 downloaded the MSMARCO corpus quite long ago. I guess they have added some other files to the tarball since then but do not change `collection.tsv`.

rodrigonogueira4:
MS Marco organizers added some files to the collection tar file:
https://github.com/dfcf93/MSMARCO/issues/29

lintool:
Okay, great! Thanks @edwardhdlu

PR #640, opened 2019-05-10
Add CAR 2.0 regression

lintool:
The regressions now replicate the BM25 baseline in https://github.com/nyu-dl/dl4marco-bert
(and provides additional models)


docs/experiments-car17v1.5.md
```diff
@@ -11,7 +11,7 @@ lucene-index.car17v1.5.pos+docvectors+rawdocs -storePositions -storeDocvectors \
 -storeRawDocs >& log.car17v1.5.pos+docvectors+rawdocs &
 ```
```
rodrigonogueira4: nit: CAR is all upper case.

docs/experiments-car17v1.5.md
```diff
@@ -20,7 +20,7 @@ For additional details, see explanation of [common indexing options](common-inde
 Topics and qrels are stored in `src/main/resources/topics-and-qrels/`, downloaded from NIST:
 
 + `topics.car17v1.5.test200.txt`: [Topics for the test200 subset (TREC 2017 Complex Answer Retrieval Track)](http://trec-car.cs.unh.edu/datareleases/v1.5/test200-v1.5.tar.xz)
-+ `qrel: qrels.car17v1.5.test200.txt`: [adhoc qrels (TREC 2017 Complex Answer Retrieval Track)](http://trec-car.cs.unh.edu/datareleases/v1.5/test200-v1.5.tar.xz)
++ `qrels.car17v1.5.test200.txt`: [adhoc qrels (TREC 2017 Complex Answer Retrieval Track)](http://trec-car.cs.unh.edu/datareleases/v1.5/test200-v1.5.tar.xz)
```
rodrigonogueira4: should we use benchmarckY1test as these are the official test queries of 2017?

docs/experiments-car17v1.5.md
lintool: I think for all the file names we downcase? E.g., robust04 not Robust04, nyt instead of NYT, etc.

docs/experiments-car17v1.5.md
lintool: yes, I'll circle around to do this, but there seems to be some problem with the 1.5 topics, which I gotta ask Laura about...

Victor0118: By the time I build this doc, the qrel file for benchmarckY1test hadn't been out yet. So I use the test200 queries. 

I think I noticed a little difference between queries between 2018 and 2017 before. We can handle them together in one topic reader if we have plan to add benchmarckY1test queries.

Issue #639, opened 2019-05-10
Reimplement MS MARCO retrieve.py script in Java


lintool:
Currently, MS MARCO retrieval is performed by this Python script `src/main/python/msmarco/retrieve.py`

I'd make sense to reimplement this in Java, where it's probably easier to add in features like multithreading.

Would also be an opportunity to understand #604

lintool:
Completed!

PR #644, opened 2019-05-13
Java rewrite of retrieve.py

edwardhdlu:
See https://github.com/castorini/anserini/issues/639

You can run this in IntelliJ by going to Run > Edit Configurations and setting Program arguments: 
```
-index=msmarco_data/lucene-index-msmarco -qid_queries=msmarco_data/queries.dev.small.tsv -output=msmarco_data/run.dev.small.tsv -hits=1000 -k1=0.82 -b=0.72 -topicreader= -topics=
```

On my machine, this script gets `0.015 s/query` while `retrieve.py` only gets `0.062 s/query`. Is all this due to python overhead? Running the evaluation scripts gives the same results in both cases. Also not sure what to name this and where to put it.

@lintool


lintool:
@Victor0118 @rodrigonogueira4 
wow, it seems like Python adds a lot of overhead...

src/main/java/io/anserini/search/RetrieveMain.java
```diff
@@ -0,0 +1,71 @@
+package io.anserini.search;
```
lintool: Let's rename this `SearchMsmarco`? Looks a bit ugly that's what camel case convention would dictate?

Let's add a binding to generate a launch script, per https://github.com/castorini/anserini/blob/master/pom.xml#L106

src/main/java/io/anserini/search/RetrieveMain.java
```diff
@@ -0,0 +1,71 @@
...
+import org.kohsuke.args4j.ParserProperties;
+
+import java.io.File;
+import java.util.List;
+
+/*
+ * Java rewrite of retrieve.py
+ */
+public class RetrieveMain {
+    public static void main(String[] args) throws Exception {
```
lintool: repo convention is two space indent; please fix all throughput.

src/main/java/io/anserini/search/RetrieveMain.java
```diff
@@ -0,0 +1,71 @@
...
+        SearchArgs searchArgs = new SearchArgs();
```
lintool: Create a new args object, since this is completely different code path and most of the `SearchArgs` fields aren't relevant.

lintool: Line up the args to be the same as the python script.

lintool:
Also, update https://github.com/castorini/anserini/blob/master/docs/experiments-msmarco.md
Keep old Python docs, just add in a Java option.

lintool:
Great work, thanks for the contribution!

Issue #602, opened 2019-04-25
Build Python wrapper to Anserini collection API


lintool:
Which will allow us to, for example, iterate through documents in a collection, without reimplementing all the collection parsing gunk.

lintool:
@emmileaf as you're working your way through the collection API, please add test cases per #371

lintool:
Hi @emmileaf this is a good start here:
https://github.com/emmileaf/anserini/tree/python-collection/src/main/python/segment_collection

So, I'm thinking that the collection wrapping code can start moving into a central lib in `src/main/python/io/anserini/collection`?

Such that we'll be able to do things like:

```
collection = Collection('Robust04', '/path/to/collection')

for segment in collection.segments:
   for doc in segment.docs:
      ...
```

And support the appropriate idioms for using threadpools...
I think it's important to hide all the Java-specific stuff so that it's completely magic from the Python perspective...

And the code to do paragraph retrieval will use the libraries you've built.

Of course, both will co-evolve, but send a PR so we can have concrete discussion on how to shape the APIs?

lintool:
Initial implementation completed, closing for now.

PR #645, opened 2019-05-13
Python collections wrapper and passage retrieval setup

emmileaf:
Keeping an open PR with initial setup for iterating over collections with Pyjnius and segmenting documents, as well as discussion for adapting this to generally address [#602](https://github.com/castorini/anserini/issues/602).


emmileaf:
Currently the setup uses the java document generator classes for collection-specific content parsing logic. This is more interlaced with other parts of the code such as IndexCollection:


https://github.com/castorini/anserini/blob/32e01911f34766b1d87bbf3a3c5efc31ecb4f206/src/main/java/io/anserini/index/generator/LuceneDocumentGenerator.java#L75-L79

An alternative to this approach could be providing a Python interface for only the collections classes, and skip using generator classes to convert `SourceDocument` into `LuceneDocument`? This would move the content parsing and HTML cleaning logic over to be addressed or re-written in Python instead.

src/main/python/segment_collection/utils.py
```diff
@@ -0,0 +1,54 @@
+import jnius_config
```
lintool: And you're planning on moving common code to something like `src/main/python/io/anserini/collection/`? We probably want to mirror the java package structure?

src/main/python/segment_collection/README.md
```diff
@@ -0,0 +1,40 @@
+### Segmenting collection
```
lintool: Let's call this `passage_retrieval`?
So move everything to `src/main/python/passage_retrieval/`?

emmileaf:
PR changes:

- moved to  [`src/main/python/passage_retrieval`](https://github.com/emmileaf/anserini/tree/python-collection/src/main/python/passage_retrieval)
- used Python enums for specifying collection and generator classes
- factored out a common collections wrapper into [`passage_retrieval/collection`](https://github.com/emmileaf/anserini/tree/python-collection/src/main/python/passage_retrieval/collection) for now, which can support usage similar to:

```
collection = Collection(collection_class, input_path)

for (i, fs) in enumerate(collection.segments):
    for (i, doc) in enumerate(fs):
         foo(doc)
```

[`passage_retrieval/collection`](https://github.com/emmileaf/anserini/tree/python-collection/src/main/python/passage_retrieval/collection) is what I can move into `src/main/python/io/anserini/collection/` for later…  In terms of mirroring the java package structure, are we expecting to have other subdirectories fall into `src/main/python/io/anserini/` in the future? @lintool

src/main/python/passage_retrieval/collection_iterator.py
```diff
@@ -0,0 +1,123 @@
...
+    logger.info("# Final Counter Values");
+    logger.info("indexable:     {:12d}".format(collection.counters.indexable.value))
+    logger.info("unindexable: {:12d}".format(collection.counters.unindexable.value))
+    logger.info("skipped:     {:12d}".format(collection.counters.skipped.value))
+    logger.info("errors:      {:12d}".format(collection.counters.errors.value))
+    
+    logger.info("Total duration: %s", str(datetime.timedelta(seconds=elapsed)))
+    
+
+#IterCollection('C:/cygwin64/home/Emily/usra/collection/disk45',
```
lintool: probably want to remove this?

lintool:
> passage_retrieval/collection is what I can move into src/main/python/io/anserini/collection/ for later… In terms of mirroring the java package structure, are we expecting to have other subdirectories fall into src/main/python/io/anserini/ in the future? @lintool

Yes, please move over? I anticipate that we'll gradually mirror and "export" Java classes as Python gets more use. For example, we probably want to expose the topic readers in Python also... I can imagine doing learning-to-rank in Python also...

Let's do the `src/main/python/io/anserini/collection/` move, and we should merge this...

PR #650, opened 2019-05-20
Added multithreaded indexing through Elasticsearch (with ES 7.0.0)

charW:
This is part of the effort of integrating Elasticsearch into Anserini. It's based off the `lucene8` branch to avoid class path errors.

The current approach in this PR is to have bulks of indexing requests (of a specified batch size) sent to Elasticsearch high level REST clients taken from a pool; this is very similar to the approach of the Solr integration.

Here, we set up the ELK stack using [docker-elk](https://github.com/deviantony/docker-elk).

When testing, adjust the `es` parameters as you see fit. Depending on the documents you are indexing, you probably also have to increase the ELK stack's heap size (adjust the corresponding `JAVA_OPTS` in `docker-compose.yml` in `docker-elk`).

Note: As of May 19, 2019, this PR is not meant for an immediate merge; instead, it's a source for continued suggestions, improvements, and collaborations.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -226,6 +226,31 @@
       <version>${lucene.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpcore</artifactId>
+      <version>4.4.11</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>4.5.8</version>
+    </dependency>
+    <dependency>
+      <groupId>org.elasticsearch.client</groupId>
+      <artifactId>elasticsearch-rest-high-level-client</artifactId>
+      <version>7.0.0</version>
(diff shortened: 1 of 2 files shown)


lintool:
@charW are the issues w/ trying to patch the main trunk?
Upgrading to Lucene8 is going to be a huge operation... and I think I'd like the ES patch to go in before that...

But otherwise I assume @r-clancy has got this covered in terms of CR, etc.?

r-clancy:
There's runtime errors with transitive dependencies with the latest ELK stack (which uses Lucene 8)... @charW want to give ELK `6.X` a try instead of `7.X`? It should be trivial to upgrade ELK in the future.

@lintool Yep, I've got the CR covered for this one - this PR will be a WIP until we're ready to merge.

charW:
@lintool @r-clancy Good point! I'll look into using another version of Elasticsearch...

r-clancy:
Closing this one as we've got https://github.com/castorini/anserini/pull/655

PR #655, opened 2019-05-28
Multithreaded indexing through Elasticsearch (with ES 7 & Lucene 8)

charW:
This is part of the effort of integrating Elasticsearch into Anserini. 

**Update:** we have recently rebased the branch `add-elasticsearch-lucene7` so it shares the same Lucene 8 dependency as the master branch (this makes the branch name a misnomer), and we have also updated this branch to use Elasticsearch 7.0.0.

### DIFF
--- a/docs/elastirini.md
+++ b/docs/elastirini.md
@@ -1,14 +1,55 @@
 # Elastirini - Elasticsearch integration with Anserini
 
-## Setup
+## Deploying ELK stack locally
 
-We use [docker-elk](https://github.com/deviantony/docker-elk) to set up the ELK stack locally; please see the link for more setup and configuration instructions.
+We use [docker-elk](https://github.com/deviantony/docker-elk) to set up the ELK stack locally.
 
-Note: depending on the documents you are indexing, you probably also have to increase the ELK stack's heap size by adjusting the corresponding `ES_JAVA_OPTS` (for Elasticsearch) and `LS_JAVA_OPTS` (for Logstash) in `docker-compose.yml` in `docker-elk`.
+Before we start, make sure you have Docker and Docker Compose installed and running.
 
-## Index
+First, we clone the repository and switch into the directory.
 
-To index with Elasticsearch, instead of passing in `-index` (to index with Lucene directly) or `-solr` (to index with Solr), we pass in `-es`. For example, to index [robust04](https://github.com/castorini/Anserini/blob/master/docs/experiments-robust04.md), we could run:
-`sh target/appassembler/bin/IndexCollection -collection TrecCollection -generator JsoupGenerator -es -es.index [index name] -threads 16 -input [absolute path to disk45] -storePositions -storeDocvectors -storeRawDocs`
+`git clone https://github.com/deviantony/docker-elk.git && cd docker-elk`
 
-There are additional `-es` parameters for using Elasticsearch in Anserini; you could specify them as you see fit.
+Depending on the documents you are indexing, you probably also have to increase the ELK stack's heap size in `docker-compose.yml`:
+
+To increase Elasticsearch's heap size (on MacOS):
+
+`sed -i '' 's/ES_JAVA_OPTS: "-Xmx256m -Xms256m"/ES_JAVA_OPTS: "-Xmx1g -Xms512m"/' docker-compose.yml`
+
+To increase Logstash's heap size (on MacOS):
+
+`sed -i '' 's/LS_JAVA_OPTS: "-Xmx256m -Xms256m"/LS_JAVA_OPTS: "-Xmx1g -Xms512m"/' docker-compose.yml`
+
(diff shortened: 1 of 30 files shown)


r-clancy:
When documents are indexed, they are placed in an index based on the name of the file they came from. Can we add a `-es.index` option (just like `-solr.index`) to specify the name of the index?
image

r-clancy:
Currently, both the `raw` and `contents` fields are indexed into ES regardless of the values of `-storeRawDocs` (which could be raw HTML) and `-storeTransforedDocs` (which could be parsed text from HTML), respectively. This isn't an issue in the Solr version (which the ES one follows) as it's done via configuration and not code in Solr.

When the `-storeRawDocs` flag is passed, `raw` should be indexed and when the `-storeTransformedDocs` flag is passed, `contents` should be stored.

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -400,6 +449,121 @@ private void flush() {
...
+          builder.startObject();
+          for (IndexableField field : document.getFields()) {
+            if (field.stringValue() != null) {
+              builder.field(field.name(), field.stringValue());
+            } else if (field.numericValue() != null) {
+              builder.field(field.name(), field.numericValue());
+            }
+          }
+          builder.endObject();
+          bulkRequest.add(new IndexRequest(input.getFileName().toString(), "doc", sourceDocument.id()).source(builder));
```
r-clancy: When documents are indexed, they are placed in an index based on the name of the file they came from. Can we add a -es.index option (just like -solr.index) to specify the name of the index?

image

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -400,6 +449,121 @@ private void flush() {
...
+            continue;
+          }
+          if (whitelistDocids != null && !whitelistDocids.contains(sourceDocument.id())) {
+            counters.skipped.incrementAndGet();
+            continue;
+          }
+
+          XContentBuilder builder = XContentFactory.jsonBuilder();
+          builder.startObject();
+          for (IndexableField field : document.getFields()) {
```
r-clancy: Currently, both the raw and contents fields are indexed into ES regardless of the values of -storeRawDocs (which could be raw HTML) and -storeTransforedDocs (which could be parsed text from HTML), respectively. This isn't an issue in the Solr version (which the ES one follows) as it's done via configuration and not code in Solr.

When the -storeRawDocs flag is passed, raw should be indexed and when the -storeTransformedDocs flag is passed, contents should be stored.

r-clancy:
The `-storeRawDocs` and `-storeTransformedDocs` changes look good!

A couple things I noticed...

The performance of ES is quite slow in comparison to Lucene (expected), but also Solr - take a look into this? From a quick search, there appears to be a `index.refresh_interval` option, which controls how quickly documents are available for searching, that defaults to 1 second. In the Solr implementation, we control similar behavior with the `solr.commitWithin` option. Let's have some more fine grained control over when documents are committed within ES? I'm not sure what the `RequestOptions` mechanism exposes, if not that maybe update the refresh interval for the index to `60s` by default like Solr?

Once indexing is complete, the program doesn't appear to exit and I need to Ctrl-C to get out of it. I'll try on another PC just to make sure, but maybe some resource isn't being closed properly? I don't see the ES client pool being closed...

r-clancy:
@charW Can we also rebase this branch to the current `master` which is now Lucene 8.0.0?

r-clancy:
@emmileaf @lintool Can one of you follow the `elastirini.md` and sanity check this PR before we merge?

docs/elastirini.md
```diff
@@ -0,0 +1,61 @@
...
+
+`curl --user elastic:changeme -XPUT "localhost:9200/yourindexname"`
+
+Here, the username and password are those defaulted by `docker-elk`. You can change these if you like.
+
+You can further specify the settings associated with this index. For example, if you would like to change `index.refresh_interval` from the default 1 second to 60 seconds:
+
+`curl --user elastic:changeme -XPUT -H 'Content-Type: application/json' 'localhost:9200/yourindexname/_settings' -d '{ "index": { "refresh_interval": "60s"}}'`
+
+Now, we can start indexing through Elastirini. Here, instead of passing in `-index` (to index with Lucene directly) or `-solr` (to index with Solr), we pass in `-es`. For example, to index [robust04](https://github.com/castorini/Anserini/blob/master/docs/experiments-robust04.md), we could run:
```
r-clancy: URL has changed to https://github.com/castorini/anserini/blob/master/docs/regressions-robust04.md

emmileaf:
> @emmileaf @lintool Can one of you follow the `elastirini.md` and sanity check this PR before we merge?

Was able to setup and index `robust04` according to doc. LGTM!

PR #656, opened 2019-05-28
Updated Python collections wrapper and passage retrieval setup

emmileaf:
Updates to #645:
* Fixed pyjnius type issues and file segment naming
* Updated tokenizer and score aggregation code
* Updated documentation and results


src/main/python/passage_retrieval/document_scorer.py
```diff
@@ -0,0 +1,65 @@
...
+                       dtype={"topic": "object", 
+                              "query": "object", 
+                              "id": "object",
+                              "rank": "int64",
+                              "score": "float64",
+                              "tag": "object"})
+    
+    new = data["id"].str.split(".", n=1, expand=True)
+    data["docid"] = new[0] # get source document id
+#    data["segid"] = new[1] # get segment id
```
zeynepakkalyoncu: Remove if we don't want to set `segid` in `data`?

src/main/python/passage_retrieval/segment.py
```diff
@@ -23,16 +21,22 @@
...
+                    help='minimum size value for tokenizer, defaults to zero', required=False)
     parser.add_argument("--raw", '-r', action='store_true',
                     help='defaults to false for transformed text', required=False)
 
     args = parser.parse_args()
     
+    logfile = os.path.join(os.path.dirname(os.path.abspath(args.output)), 
+                           os.path.basename(os.path.abspath(args.output)) + '.log')
+    
+    logging.basicConfig(level=logging.DEBUG,
```
zeynepakkalyoncu: If you want to reuse the logger, you could create a `Logger` class to set these attributes in one place (like `log4j.properties` in Java)?

emmileaf:
Thanks @zeynepakkalyoncu for the help! Just made some edits to address the comments, and added more docstrings overall.

For re-using the logger, did you mean across different modules? I believe that calling `logger = logging.getLogger(__name__)` within the package hierarchy is already doing this, but please correct me if I misunderstood?

lintool:
@emmileaf do you sentences align with sentences from @zeynepakkalyoncu for birch? if so, great, if not, we should make them consistent, otherwise it will be difficult to match BERT scores with BM25 scores...

zeynepakkalyoncu:
@emmileaf Yes, you are right - I thought the log file was repeated within the same directory, but what you have is fine. Looks good!

emmileaf:
> @emmileaf do you sentences align with sentences from @zeynepakkalyoncu for birch? if so, great, if not, we should make them consistent, otherwise it will be difficult to match BERT scores with BM25 scores...

@lintool  Right now they are not using the same code, but resulting sentences are mostly similar. I was able to do an approximate matching on the sentences to get a rough idea of score differences, but there are probably some sentences that don't exactly align (e.g. birch breaking up sentences greater than 512 chars, and likely other small inconsistencies). 

For the sake of exact comparison, I'm thinking of defining something like `split_sentence_birch` in the tokenizer options here and try to replicate the birch sentence logic… but it would probably make sense to have it separate from this PR?

lintool:
@emmileaf yes, separate PR makes sense...

Issue #654, opened 2019-05-25
Improved processing of WashPost fields?


lintool:
A follow up on: https://github.com/castorini/anserini/issues/646#issuecomment-495253072

> Around 4% of Core18 documents have empty body content (e.g. images) so no term vector is created, but other parsed fields such as title are stored and the document is not skipped over.

It might make sense to try to append the title to the body content so that at least something is indexed?

@emmileaf can you look in the qrels file and see if any of docids are relevant?

emmileaf:
I've had some issues replicating [experiments-core18.md](https://github.com/castorini/anserini/blob/master/docs/experiments-core18.md) and think it might be related to this...seems like body content had more fields appended to it pre-#593?

@r-clancy Do you have any other dependencies on a paragraph-only content field? It looks like term vectors are only created from the contents field (FIELD_BODY), and removing the `title`/`caption`/`kicker` fields alters the index + retrieval results from before.

Perhaps we can append title/caption/etc. back into the content field, but just try to clean up the original parsing issues - thoughts?

Previous:

```
BM25
map                     all     2491
P_30                    all     3580

Index statistics
----------------
documents:             595037
documents (non-empty): 595037
unique terms:          840929
total terms:           317882653
stored fields:
  article_url (indexOption: DOCS, hasVectors: false)
  author (indexOption: DOCS, hasVectors: false)
  contents (indexOption: DOCS_AND_FREQS_AND_POSITIONS, hasVectors: true)
  fullCaption (indexOption: DOCS, hasVectors: false)
  id (indexOption: DOCS, hasVectors: false)
  kicker (indexOption: DOCS, hasVectors: false)
  title (indexOption: DOCS, hasVectors: false)
```
Current:


```
BM25
map                     all     0.2444 
P_30                    all     0.3520 

Index statistics
----------------
documents:             595037
documents (non-empty): 585493
unique terms:          817578
total terms:           304740621
stored fields:
  article_url (indexOption: DOCS, hasVectors: false)
  author (indexOption: DOCS, hasVectors: false)
  contents (indexOption: DOCS_AND_FREQS_AND_POSITIONS, hasVectors: true)
  id (indexOption: DOCS, hasVectors: false)
  title (indexOption: DOCS, hasVectors: false)
```

lintool:
Hrm. So a regression broke, huh? And made results worse?

Yes, I agree... clean up but add those fields back in?

r-clancy:
[...]

Issue #658, opened 2019-05-31
Core18 (Washington Post) regression failure


lintool:
@emmileaf reported a regression failure in #654 but exposing here at top level so this doesn't get buried.

@emmileaf can you copy and paste the regression log in this issue?

emmileaf:
Previous index:

```
BM25
map                     all     0.2491
P_30                    all     0.3580

Index statistics
----------------
documents:             595037
documents (non-empty): 595037
unique terms:          840929
total terms:           317882653
stored fields:
  article_url (indexOption: DOCS, hasVectors: false)
  author (indexOption: DOCS, hasVectors: false)
  contents (indexOption: DOCS_AND_FREQS_AND_POSITIONS, hasVectors: true)
  fullCaption (indexOption: DOCS, hasVectors: false)
  id (indexOption: DOCS, hasVectors: false)
  kicker (indexOption: DOCS, hasVectors: false)
  title (indexOption: DOCS, hasVectors: false)
```
Current index:
```
BM25
map                     all     0.2444 
P_30                    all     0.3520 

Index statistics
----------------
documents:             595037
documents (non-empty): 585493
unique terms:          817578
total terms:           304740621
stored fields:
  article_url (indexOption: DOCS, hasVectors: false)
  author (indexOption: DOCS, hasVectors: false)
  contents (indexOption: DOCS_AND_FREQS_AND_POSITIONS, hasVectors: true)
  id (indexOption: DOCS, hasVectors: false)
  title (indexOption: DOCS, hasVectors: false)

**Comment by r-clancy** [comment]
I reverted the commit (https://github.com/castorini/anserini/commit/c4ab6bfbe38648337ad824503e8945f5111ac673) and removed the `Optional` wrapper around the title, but there appears to be some documents that have no title and no paragraph content. When unwrapping the `Optional` for the `title` and adding it to `contents`, we now have additional empty documents (where before it likely indexed `Optional[]`, the empty `Optional` string value):
```
2019-05-31 01:48:05,053 - regression_test - INFO - ==========Verifying Index==========
2019-05-31 01:48:05,053 - regression_test - INFO - [Index]: lucene-index.core18.pos+docvectors+rawdocs
2019-05-31 01:48:05,829 - regression_test - INFO - documents: 595037
documents (non-empty): expected=595037, actual=594483
Traceback (most recent call last):
File "src/main/python/run_regression.py", line 237, in <module>
verify_index(yaml_data, args.index, args.dry_run)
[...]

PR #661, opened 2019-05-31
Fix WaPo Regressions

r-clancy:
- Only add the `title` if it's present (non-empty `Optional`)
- Add kicker to the `contents` field
- Update regression numbers for slight changes


r-clancy:
Fixes #654 and #658

PR #662, opened 2019-05-31
Increase timeout for Solr to 5 mins...

r-clancy:
some of the himrod tests were timing out with the default value of 10 seconds.


r-clancy:
@lintool Slipping this one in before 0.5.0...

lintool:
Will merge once travis turns green.

Issue #705, opened 2019-06-12
Master build is broken


lintool:
See build: https://travis-ci.org/castorini/Anserini/builds/544720828

```
Installing oraclejdk8
$ export JAVA_HOME=~/oraclejdk8
$ export PATH="$JAVA_HOME/bin:$PATH"
$ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"
install-jdk.sh 2019-05-02
Expected feature release number in range of 9 to 13, but got: 8
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .
```

Related: https://travis-ci.community/t/the-travis-ci-build-could-not-be-completed-due-to-an-error/1345

lintool:
Fixed by #709

PR #706, opened 2019-06-12
Update .travis.yml to oraclejdk9

lintool:


lintool:
Issue reported in #705, fixed by #709

PR #708, opened 2019-06-12
Update .travis.yml to oraclejdk12

lintool:
Let's just try one that works...


lintool:
Issue reported in #705, fixed by #709

PR #711, opened 2019-06-12
Added Pyjnius configuration helper to add jar to classpath

### LINKED ISSUE #678: Simplify pyjnius configuration - jar import
Currently, we need to do something like this:

```
import jnius_config
jnius_config.set_classpath("target/anserini-0.4.1-SNAPSHOT-fatjar.jar")
...
```

Our scripts have the jar hardcoded... is there some auto-config magic we can do?

emmileaf:
- Added setup script `src/main/python/pyjnius_setup.py` that searches for the latest matching jar in target
- Replaced existing instances of the following code to call `pyjnius_setup.py` so that version is not hardcoded 
```
import jnius_config
jnius_config.set_classpath("target/anserini-0.6.0-SNAPSHOT-fatjar.jar")
```
- Closes #678.

### DIFF
--- a/src/main/python/msmarco/retrieve.py
+++ b/src/main/python/msmarco/retrieve.py
@@ -18,12 +18,16 @@ limitations under the License.
 import argparse
 import time
 
-import jnius_config
-jnius_config.set_classpath('target/anserini-0.6.0-SNAPSHOT-fatjar.jar')
+# Pyjnius setup
+import sys
+sys.path += ['src/main/python']
+from pyjnius_setup import configure_classpath
+configure_classpath()
 
 from jnius import autoclass
 JString = autoclass('java.lang.String')
 JSearcher = autoclass('io.anserini.search.SimpleSearcher')
+print("Pyjnius check passed")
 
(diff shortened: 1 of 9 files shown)


src/main/python/msmarco/retrieve.py
```diff
@@ -18,12 +18,16 @@
...
+# Pyjnius setup
+import sys
+sys.path += ['src/main/python']
+from pyjnius_setup import configure_classpath
+configure_classpath()
 
 from jnius import autoclass
 JString = autoclass('java.lang.String')
 JSearcher = autoclass('io.anserini.search.SimpleSearcher')
+print("Pyjnius check passed")
```
bazingagin: this line can be deleted :)

emmileaf: oops, thanks for catching this!

lintool:
@emmileaf take a look at this line: https://github.com/castorini/anserini/blob/master/.travis.yml#L19

I believe it's supposed to test for the Python interface... can you make sure that's still the right thing to do?

Issue #667, opened 2019-06-01
Refactoring Collection API


lintool:
Looking at the current structure of the Collection API, I'm a bit puzzled as to why it was designed this way... perhaps it's accumulated debt?

It seems like the conceptual organization is as follows: A collection is comprised of segments. Each segment is backed by a file and provides an iterator of documents.

So, this definition makes complete sense:

```
public interface Segment<T extends SourceDocument> extends Iterator<T>
```

But why doesn't `DocumentCollection` extend an iterator over Segments, or at the very least have a `getSegments` method? Instead, there's this `SegmentProvider` class, and a collection, like `ClueWeb09Collection`, has the following definition:

```
public class ClueWeb09Collection extends DocumentCollection
    implements SegmentProvider<ClueWeb09Collection.Document>
```

Why? Why implement an interface as oppose to extend an abstract class?

@emmileaf can you take a look at this and think through what the Collection API _should_ be? While we're developing the Python bridge, we might as well see if we need to clean up the API on the Java end.

This should probably happen post v0.5.0 release...

emmileaf:
`SegmentProvider` offers something similar to `getSegments`, represented as a list of paths:
https://github.com/castorini/anserini/blob/7fd1f648dcdd196cda436bb6af129c015dbdb9b6/src/main/java/io/anserini/collection/SegmentProvider.java#L29-L38

I agree that these methods can move to the `DocumentCollection` abstract class instead, getting rid of the interface. However, from #331  and #332  it seems like the original intent of having the `SegmentProvider` interface was to make the `collection-segment-document` relationship more explicit in signatures. Do we still want this to be the case?

lintool:
I agree we should make the collection/segment/document relationship explicit, but I don't think the current design does that... for example, in theory we could have a collection without a provider... and classes other than collections can implement the provider... which may be confusing.

IIRC, we originally considered the possibility of other types of segments... segments backed by a file, segments backed by a stream (e.g., live Twitter feed), etc. We were probably overly ambitious, as non-file-backed segments never materialized... so `FileSegment` and `Segment` can probably be collapsed.

Now that I think about it, I remember why we designed it this way... `Iterator` is an interface, so a collection would need to be something like:

```
public interface DocumentCollection<D> extends Iterator<Segment<D>>
```

[...]

PR #717, opened 2019-06-13
Preliminary refactoring of collections API

emmileaf:
(For issue #667 - not ready for merge into master, but any feedback would be super helpful 🙂 )

In this PR:

- Refactored classes within the collection API, summarized by the following signatures:

```
public abstract class DocumentCollection<T extends SourceDocument> implements Iterable<FileSegment<T>>
public abstract class FileSegment<T extends SourceDocument> implements Iterable<T>
public interface SourceDocument
```

- This is a bit different than what we proposed in the issue, but I kept `DocumentCollection` and `FileSegment` as abstract classes that implement the Iterable interface since there is quite a bit of common code repeated with each implementation of a collection subclass. We can iterate over collections and segments through the `Iterable` interface (given collectionClass and collectionPath):

```
DocumentCollection collection = (DocumentCollection) collectionClass.newInstance();
testCollection.setCollectionPath(collectionPath);

for (Object s : collection){
  FileSegment segment = (FileSegment) s; 
  for (Object d : segment){
    SourceDocument doc = (SourceDocument) d; 
    foo(d);
  }
}
```
- `IndexCollection`, `MapCollections`, and tests were modified so that they are compatible with the refactored code, but major changes to the way they iterate over collections are held off for now since the current diff is pretty big with a bunch of entangled changes already.


src/main/java/io/anserini/collection/FileSegment.java
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Base implementation for a {@link FileSegment}.
+ * A collection is comprised of one or more file segments. Note that implementations may have independent
+ * existence outside a collection, and in principle multiple collections might share the same
+ * {@code FileSegment} implementation.
+ */
+public abstract class FileSegment<T extends SourceDocument> implements Iterable<T> {
+
+  public enum Status {
+    SKIPPED, ERROR, VOID
```
lintool: does `VOID` actually mean `EOF`?

src/main/java/io/anserini/collection/FileSegment.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  /**
+   * For concrete classes to implement depending on desired iterator behaviour
+   *
+   * @throws IOException if reader error encountered
+   */
+
+  protected abstract void readNext() throws IOException;
```
lintool: Since we're refactoring here, does it make sense to have `readNext()` here anymore?

src/main/java/io/anserini/collection/DocumentCollection.java
```diff
@@ -88,19 +93,72 @@ public final Path getCollectionPath() {
...
   }
 
+  /**
+   * Creates a {@code FileSegment} from a path.
+   *
+   * @param p path
+   * @return {@code FileSegment} with the specified path
+   * @throws IOException if file access error encountered
+   */
+  public abstract FileSegment<T> createFileSegment(Path p) throws IOException;
```
lintool: Do we still need this? Or shall we let the iterator figure out how to this itself?

lintool:
@emmileaf thanks, this looks great!

A few comments for you to think about... at this point you probably have a better sense of all the details...

Have you run some of the regressions to make sure things still work? (At least the small ones... we can save the large ones for when `tuna` is less busy...)

emmileaf:
> @emmileaf thanks, this looks great!
> 
> A few comments for you to think about... at this point you probably have a better sense of all the details...
> 
> Have you run some of the regressions to make sure things still work? (At least the small ones... we can save the large ones for when `tuna` is less busy...)

Thanks for the help! I've only had a chance to run the robust04 regression for now, but will try to run a few more later (probably after iterating over some changes) and update.

src/main/java/io/anserini/collection/FileSegment.java
emmileaf: If we keep both abstract `readNext()` and abstract` createFileSegment()`, different concrete classes can share the same iterator implementations defined in `DocumentCollection` and `FileSegment`. This way, each time we add a new collection, it can just implement these two helper methods instead of overriding the entire iterator?

lintool: Okay, makes sense!

emmileaf:
Added some exception handling changes to the previous `FileSegment` iterator:
- Three statuses `VOID`, `SKIPPED`, and `ERROR` were used to keep track of different cases of exception handling within the iterator. `TrecwebCollection` and `TweetCollection` were throwing unchecked `RunTimeException` for skipping documents, because of limitations in the `hasNext()` signature

Changed to:
- `readNext()` now throws three types of exceptions which are caught in `hasNext()`: `IOException` (error), `NoSuchElementException` (EOF), and `ParseException` (skip). Each segment instance keeps count of whether error occurred (iteration stopped) as well as number of skips (iteration continued)

So far the following regressions pass (still working on some of the larger ones):

```
nohup python src/main/python/run_regression.py --collection disk12 --index >& log.disk12 &
nohup python src/main/python/run_regression.py --collection robust04 --index >& log.robust04 &
nohup python src/main/python/run_regression.py --collection robust05 --index >& log.robust05 &
nohup python src/main/python/run_regression.py --collection core17 --index >& log.core17 &
nohup python src/main/python/run_regression.py --collection core18 --index >& log.core18 &
nohup python src/main/python/run_regression.py --collection wt10g --index >& log.wt10g &
nohup python src/main/python/run_regression.py --collection mb11 --index >& log.mb11 &
nohup python src/main/python/run_regression.py --collection car17v1.5 --index >& log.car17v1.5 &

emmileaf:
I'll go ahead and merge this into the `collection-refactor` branch for now. Will create a separate PR later for merge into `master`, so that the accumulated changes won't get too messy.

PR #736, opened 2019-07-02
Fixed "neural hype" tuning experiments after Lucene 8.0 upgrade

lintool:
+ Retuned retrieval models for Lucene 8.0

Refactored tuning script:
+ made command-line parameters more consistent
+ broke fold settings into external config files for greater generality
+ removed unintuitive distinction between "model" and "basemodel": there's just "model" now.


src/main/python/fine_tuning/evaluation.py
```diff
@@ -62,13 +62,13 @@ def output_all_evaluations(self, qrel_programs, qrel_file_path, result_file_path
...
                 try:
                     if i == 0:
-                        o = open( output_path, 'w')
+                        o = open(output_path, 'w')
                     else:
-                        o = open( output_path, 'a')
+                        o = open(output_path, 'a')
                     if 'trec_eval' in qrel_program:
-                        o.write(stdout)
+                        o.write(stdout.decode("utf-8"))
```
Peilin-Yang: Why we need a `decode` here?

src/main/python/fine_tuning/run_batch.py
```diff
@@ -207,11 +215,11 @@ def del_method_related_files(method_name):
...
     parser.add_argument('--collection', required=True, help='the collection key in yaml')
-    parser.add_argument('--basemodel', default='bm25', choices=['bm25', 'ql'], help='the ranking model')
-    parser.add_argument('--model', default='axiom', choices=['bm25', 'ql', 'axiom', 'rm3', 'bm25+axiom', 'bm25+rm3'], help='the higher level model')
-    parser.add_argument('--n', dest='parallelism', type=int, default=16, help='number of parallel threads for retrieval/eval')
+    parser.add_argument('--model', required=True, help='model')
+    parser.add_argument('--threads', dest='parallelism', type=int, default=16, help='number of parallel threads for retrieval and evaluation')
     parser.add_argument('--output_root', default='fine_tuning_results', help='output directory of all results')
-    parser.add_argument('--use_drr_fold', action='store_true', help='if specified, use the 5-folds from the deep-relevance-ranking paper')
+    parser.add_argument('--fold_settings', default='', help='fold setting')
```
Peilin-Yang: Can we make the help text more informational? something like `json file contains the fold settings, format e.g. http://xyz.com/abc.json`

src/main/python/fine_tuning/evaluation.py
lintool: If I don't have `decode` there, I get the following error:

```
"""
Traceback (most recent call last):
  File "/anaconda3/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/anaconda3/lib/python3.7/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "src/main/python/fine_tuning/run_batch.py", line 112, in atom_eval
    Evaluation.output_all_evaluations(*params)
  File "/Users/jimmylin/workspace/Anserini/src/main/python/fine_tuning/evaluation.py", line 69, in output_all_evaluations
    o.write(stdout)
TypeError: write() argument must be str, not bytes
"""
```

src/main/python/fine_tuning/run_batch.py
lintool: Done.

lintool:
@Peilin-Yang Addressed CR.

src/main/python/fine_tuning/evaluation.py
Peilin-Yang: interesting -- might be a python2 vs python3 compatibility issue.

PR #737, opened 2019-07-03
Add BM25PRF Reranker

matthew-z:
The ranking model used in https://github.com/osirrc/anserini-bm25prf-docker

[Stephen E. Robertson, and Karen Spärck Jones. Simple, proven approaches to text retrieval. University of Cambridge Computer Laboratory, 1994.](https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-356.pdf)


matthew-z:
Sorry…. it seems that there is some problem after rebasing the latest Anserini….

matthew-z:
Fixed

src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -189,6 +189,34 @@
       usage = "RM3 parameter: print original and expanded queries")
   public boolean rm3_outputQuery = false;
 
+  // BM25 PRF Options
+  @Option(name = "-bm25prf", usage = "use bm25 PRF query expansion model")
```
lintool: Can you document where you get these default parameters from? Just like how we document the source of the other parameter values?

src/main/java/io/anserini/rerank/lib/BM25PRFReranker.java
```diff
@@ -0,0 +1,336 @@
...
+
+    @Override
+    // idf is not needed in BM25PRF
+    protected float idf(long docFreq, long docCount) {
+        return 1;
+    }
+}
+
+
+public class BM25PRFReranker implements Reranker {
```
lintool: Would `BM25PrfReranker` look better? And better match Java conventions? I'm agnostic about this... I think both look ugly-ish.

src/main/java/io/anserini/rerank/lib/BM25PRFReranker.java
```diff
@@ -0,0 +1,336 @@
...
+import org.apache.lucene.search.similarities.BM25Similarity;
+import org.apache.lucene.util.BytesRef;
+
+import java.io.IOException;
+import java.util.*;
+
+import static io.anserini.index.generator.LuceneDocumentGenerator.FIELD_BODY;
+
+
+class BM25PRFSimilarity extends BM25Similarity {
```
lintool: Can you change to two-space indent, per rest of the project?

src/main/java/io/anserini/rerank/lib/BM25PRFReranker.java
```diff
@@ -0,0 +1,336 @@
+package io.anserini.rerank.lib;
```
lintool: Add the Apache boilerplate please? Just copy from other files.

src/main/java/io/anserini/rerank/lib/BM25PRFReranker.java
```diff
@@ -0,0 +1,336 @@
...
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.Term;
+import org.apache.lucene.index.Terms;
+import org.apache.lucene.index.TermsEnum;
+import org.apache.lucene.search.*;
+import org.apache.lucene.search.similarities.BM25Similarity;
+import org.apache.lucene.util.BytesRef;
+
+import java.io.IOException;
+import java.util.*;
```
lintool: Please use explicit imports, no *'s.

src/main/java/io/anserini/search/SearchArgs.java
matthew-z: The k1 and b are copied from the bm25's default options. For others, we set them arbitrarily (actually suggested by Prof.Sakai, I will ask him if there is any ref about it).

src/main/java/io/anserini/rerank/lib/BM25PRFReranker.java
matthew-z: I think `BM25PrfReranker ` is better, so I will replace `PRF`with `Prf`.

matthew-z:
Thank you for the review, I have fixed most of them (except the default parameters)

src/main/java/io/anserini/search/SearchArgs.java
matthew-z: I talked to Prof.Sakai, and it seems that these values came from his experience on TREC collections, so there is no reference about it. These values are quite "conservative", so users are expected to perform some tuning to maximize its performance.

lintool: Okay, great!

Issue #726, opened 2019-06-21
Speeding Up Search


Ricocotam:
Hi,
I'm reproducing some scripts on MS-Marco but on the whole data. I have a big cluster available and I'd like to use it properly. Currently I can only use 1 core on 1 machine (anserini doesn't exploit more), how can I change that ?

Cheers

lintool:
It's been on our TODO list to write some multi-threaded code... we're thinking, probably on the Python end... contributions are welcome!

For example, a multi-threaded version of https://github.com/castorini/anserini/blob/master/src/main/python/msmarco/retrieve.py would be a good start..

cc/ @rodrigonogueira4 @emmileaf

Ricocotam:
It is not possible to do multi-threaded search within Lucene ?
I'll give it a try but I guess a simple use of the `multiprocessing` module should be enough

lintool:
Yes, it is possible to do multi-threading on the Java end, and in fact, would be more efficient.

However, it is likely that the results will be integrated with DL code on the Python end, and so for tighter integration Python might be a good idea.

Ricocotam:
Mmh, I don't know.
I think sending a batch of queries is easier, even from Python's size, rather than handling multiple search and aggregation. What ES does is actually great with the msearch interface : you send multiple queries and a json formatted answer is given.

lintool:
In that case, it makes sense to make `SimpleSearcher` multithreaded and add a batch query interface.

Would you be willing to contribute this?

Ricocotam:
Why not, I'm not fully available from next week for 2 weeks but since it's July I guess it's not urgent.

I can give it a try at least :)

lintool:
sg, thanks!

lintool:
Closing.

PR #741, opened 2019-07-06
Added multithread support for SimpleSearcher and MSMARCO retrieval

emmileaf:
In this PR (per @Ricocotam 's suggestion in #726):
- Implemented batch query interface and multithreading support in `SimpleSearcher`
- Extended `SearchMsmarco.java` and `msmarco/retrieve.py` for multithreaded search
- Updated msmarco-passage experiment doc

Because the multithreading setup logs time per query within the Java batch call, I added logging for total end-to-end retrieval time to better estimate speed differences. Here are some speed results from running on `tuna` with 16 threads:

```
# Java version
...
Retrieving query 6700 (0.019 s/query)
Retrieving query 6800 (0.019 s/query)
Retrieving query 6900 (0.019 s/query)
Total retrieval time: 140.628 s
Done!
```

```
# Python version
...
Retrieving query 6700 (0.017 s/query)
Retrieving query 6800 (0.017 s/query)
Retrieving query 6900 (0.017 s/query)
Total retrieval time: 262.055 s
Done!
```

Python is close to 2x slower which is a larger difference than the single-threaded implementation... I think the wrapper for `SimpleSearcher`'s batch query interface probably contributes to most of the overhead here?


emmileaf:
> Thanks for implementing this!
> 
> I still could not understand why Python is ~2x slower than Java.
> It seems very unlikely that the initialization of SimpleSearch would take 120s. Calls to batchSearch should also have a small overhead, right?
> 
> Maybe the JVM initialization from python was not set to use all the threads?

Yeah I agree, the speed difference was more than what we saw in the single-threaded implementation. 
There might also be some extra overhead introduced across the python code that could be reduced? 
e.g. Java `batchSearch` returning a hashmap of  `<qid, results>` that python `batch_search` explicitly converts into a dictionary (for consistent output write order).

I'll make it a todo to look into this and see if we can improve on the python multithreading.

PR #750, opened 2019-07-26
Fix FileSegment to implement closable interface

### LINKED ISSUE #749: FileSegment should implement Closeable
hi @emmileaf - shouldn't `FileSegment` implement `Closeable` for proper cleanup?

https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/collection/FileSegment.java#L32

emmileaf:
- Resolves #749.

### DIFF
--- a/src/main/java/io/anserini/collection/FileSegment.java
+++ b/src/main/java/io/anserini/collection/FileSegment.java
@@ -22,6 +22,7 @@ import java.nio.file.Path;
 import java.text.ParseException;
 import java.util.NoSuchElementException;
 import java.util.Iterator;
+import java.io.Closeable;
 
 /**
  * Base implementation for a {@link FileSegment}.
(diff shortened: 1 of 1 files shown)


lintool:
hey @emmileaf don't we need to modify the collection and indexing classes to actually call close to properly clean up?

emmileaf:
Ah, sorry I misunderstood this... Currently `IndexCollection` does call `close()` on each segment, but are we looking to have `FileSegment`/`DocumentCollection` code further ensure this clean up?

https://github.com/castorini/anserini/blob/33d630cec4d07e926a022731046a0553aa928fb5/src/main/java/io/anserini/index/IndexCollection.java#L330-L332

lintool:
> but are we looking to have `FileSegment/DocumentCollection` code further ensure this clean up?

Yes, I'm wondering whether there are guardrails we should put up to improve hygiene. I think at this point you're more familiar with the whole collection API than anyone else... that I'll just trust whatever you recommend. If you think this PR as is sufficient, that's fine also!

emmileaf:
@lintool Sounds good, we can probably merge this PR for the v.0.6.0 release but leave the issue open? I'll take a look at how/where we can better safeguard the closable and submit a separate PR once that's implemented.

lintool:
@emmileaf sg

PR #757, opened 2019-08-01
tweaks to TREC CAR doc2query documentation

lintool:
Minor tweaks, fixed broken path.


lintool:
Broken path on line 112:

Screen Shot 2019-08-02 at 3 58 50 AM

`trec_car` instead of `treccar`

PR #764, opened 2019-08-07
Updated runbook trec 2018 background linking

Chriskamphuis:
Updated search commands
Removed duplicate information

### DIFF
--- a/docs/runbook-trec2018-anserini.md
+++ b/docs/runbook-trec2018-anserini.md
@@ -83,27 +83,18 @@ Metric            | AP*    | AP+    | NDCG*  | NDCG+  | P10*   | P10+   |
 The same as Core Track
 
 ### Submitted Runs
-```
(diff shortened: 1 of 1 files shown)


docs/runbook-trec2018-anserini.md
```diff
@@ -83,28 +83,19 @@ Metric            | AP*    | AP+    | NDCG*  | NDCG+  | P10*   | P10+   |
 The same as Core Track
 
 ### Submitted Runs
-```
```
Peilin-Yang: Why remove the commands to generate the submitted runs? 
I think it is extremely important for reproducing.

Chriskamphuis: I thought lines 87 - 91 were essentially the same as 100 - 104.

Peilin-Yang: I see... thanks!

PR #765, opened 2019-08-07
Run anserini+BM25 baseline on PubMed and DBLP

bazingagin:


docs/experiments-openresearch.md
```diff
@@ -142,3 +142,80 @@ The table below compares our BM25 results against Bhagavatula's et. al (2018):
...
+The output of using DBLP title + abstract as query
+
+```
+map                     all     0.1687
+recip_rank              all     0.5851
+P_20                    all     0.1586
+recall_20               all     0.2511
+recall_1000             all     0.6913
+f1_20                   all     0.194
+```
```
rodrigonogueira4: Could you please add a table summarizing F1@20 and MRR for our BM25 variations (title, title+abstract, key terms) and Bhagavatula's BM25? Like you did for Open Research..

docs/experiments-openresearch.md
```diff
@@ -142,3 +142,80 @@ The table below compares our BM25 results against Bhagavatula's et. al (2018):
...
 
+## Extra Baseline on PubMed and DBLP
+
+### PubMed and DBLP dataset
+
+Follow [citeomatic's repo](https://github.com/allenai/citeomatic/tree/44dc210c82515b5d4c5a96f5aafcb9b6e48206af) to download the necessary data.
+
+The steps are similar to run baseline on OpenResearch, to run all three experiments on PubMed and DBLP quickly, run
+
+`./src/main/python/openresearch/run_pubmed_dblp.sh -citeomatic_data <YOUR CITEOMATIC_DATA_ROOT> -output_folder <YOUR_OUTPUT_FOLDER>`
```
rodrigonogueira4: Awesome that you created a script for this!

src/main/python/openresearch/convert_pubmed_dblp_to_anserini_format.py
```diff
@@ -0,0 +1,196 @@
+import argparse
+import gzip
+import json
+import os
+import time
+from collections import defaultdict
+
+
+def clean(text):
+	return text.replace('\n', ' ').replace('\t', ' ')
```
rodrigonogueira4: It seems there are more than 4 spaces of identation here, right?

src/main/python/openresearch/retrieve_pubmed_dblp.py
```diff
@@ -0,0 +1,82 @@
+# -*- coding: utf-8 -*-
```
rodrigonogueira4: Could you just have used openresearch/retrieve.py since the input and output files have the same format? We could just add a check to restrict the results if a candidates_file is passed (I would call this argument `valid_doc_ids` or something similar).

src/main/python/openresearch/retrieve_pubmed_dblp_keyterms.py
```diff
@@ -0,0 +1,87 @@
+from whoosh import scoring, qparser
```
rodrigonogueira4: Same here: could we reuse openresearch/retrieve_with_key_terms.py?

src/main/python/openresearch/convert_pubmed_dblp_to_anserini_format.py
bazingagin: actually i used tabs... and tab size is set to be 4 spaces, so i think it should be all right

src/main/python/openresearch/retrieve_pubmed_dblp.py
bazingagin: yes that will make directory clean, i've changed that

src/main/python/openresearch/convert_pubmed_dblp_to_anserini_format.py
rodrigonogueira4: But please use only 4 spaces as mixing tabs and spaces breaks the code in Python 3:
https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces

bazingagin: so the official resolved this battle, sad :(  
i used all tabs in the file so it wouldn't break, but all right i will follow pep8 in the future

PR #779, opened 2019-08-13
Add Replication Log for BM25 on MS MARCO (Passage)

MathBunny:
Successfully replicated on:
OS: macOS High Sierra 10.13.3
JAVA: java version "11.0.1" 2018-10-16 LTS
Python: 3.6.4


lintool:
I don't think you want that extra space between the bullets, because otherwise MD renders weirdly? I'm not sure, but please check.

MathBunny:
You're right it looks awkward (extra space). Let me fix it

lintool:
Thanks for your contribution!

Issue #758, opened 2019-08-01
Elastirini collection retrieval for BM25


emmileaf:
Opening this issue to document some work on the [`search-es-collection`](https://github.com/castorini/anserini/tree/search-es-collection) branch, since I'll be gone by the time @r-clancy returns... This branch sets up `SearchESCollection` for BM25 retrieval, similar to `SearchSolrCollection` in #746.

Comparing BM25 retrieval in ES with Anserini:
- `robust04`: overall ranking/metrics match Anserini regressions, but document scores are different. BM25Similarity calculation for ES is higher by factor of (k1+1), likely due to this change in Lucene: https://issues.apache.org/jira/browse/LUCENE-8563
- `mb11`: TweetCollections can be indexed and searched. BM25 retrieval scores/regressions are not replicated since indexing/searching uses ES's english analyzer instead of Anserini's custom TweetAnalyzer, but metrics are not too far off.

Updating index creation settings (in documentation)
- I'm not sure if this is the best approach to this, but I also made a change to `storeRawDocs` and `storeTransformedDocs` logic so that they are set up in configs during index creation, replicating Solrini's configs (`contents` is indexed but not stored, and `raw` is stored but not indexed). The intended behaviour for this change is as follows:
	
  + If we don't pass `-storeRawDocs`,  `raw` is not added as a field by `LuceneDocumentGenerator`
  + If we pass `-storeRawDocs`, the `raw` field is stored, but not indexed in ES
  + `contents` should be indexed in ES even if we don't pass `-storeTransformedDocs`

r-clancy:
I'm able to replicate the results, except twitter, looks good to me!

>     * I'm not sure if this is the best approach to this, but I also made a change to `storeRawDocs` and `storeTransformedDocs` logic so that they are set up in configs during index creation, replicating Solrini's configs (`contents` is indexed but not stored, and `raw` is stored but not indexed). The intended behaviour for this change is as follows:
I'm good with this approach to stay consistent with the Solr implementation.

>     * `mb11`: TweetCollections can be indexed and searched. BM25 retrieval scores/regressions are not replicated since indexing/searching uses ES's english analyzer instead of Anserini's custom TweetAnalyzer, but metrics are not too far off.
@lintool should I tackle this now or circle back for it later after this is merged?

r-clancy:
Merging this as per discussion with Jimmy, will circle back later to fix Twitter experiments.

PR #785, opened 2019-08-21
Replicate search experiments with ES

r-clancy:
Will circle back later to fix Twitter experiments as per https://github.com/castorini/anserini/issues/758


r-clancy:
Quick +1 here so I can merge?

PR #786, opened 2019-08-21
Add date filter to background linking reranker

Chriskamphuis:
Date is saved as an additional field
Date filter can be toggled using command line argument

Without datefilter (BM25 + RM3)
```
target/appassembler/bin/SearchCollection -searchnewsbackground -index lucene-index.core18.pos+docvectors+rawdocs -topicreader NewsBackgroundLinking -topics ~/topics/newsir18-topics.txt -bm25 -rm3 -hits 100 -backgroundlinking.k 100 -runtag bm25_rm3 -output test/bm25-rm3.txt
```
will result in `NDCG@5 = 0.3526`  

With datefilter (BM25 + RM3)
```
target/appassembler/bin/SearchCollection -searchnewsbackground -index lucene-index.core18.pos+docvectors+rawdocs -topicreader NewsBackgroundLinking -topics ~/topics/newsir18-topics.txt -bm25 -rm3 -hits 100 -backgroundlinking.k 100 -backgroundlinking.datefilter -runtag bm25_rm3_df -output outputs/bm25-rm3-df.txt
```
will result in `NDCG@5 = 0.4171`

### DIFF
--- a/src/main/java/io/anserini/index/generator/WapoGenerator.java
+++ b/src/main/java/io/anserini/index/generator/WapoGenerator.java
@@ -45,8 +45,7 @@ public class WapoGenerator extends LuceneDocumentGenerator<WashingtonPostCollect
   public static final String FIELD_RAW = "raw";
   public static final String FIELD_BODY = "contents";
   public static final String FIELD_ID = "id";
-  public static final String FIELD_DATE = "published_date";
-  
+
   private static final String PATTERN = "<.+>";
   public static final List<String> CONTENT_TYPE_TAG = Arrays.asList("sanitized_html", "tweet");
 
@@ -86,7 +85,7 @@ public class WapoGenerator extends LuceneDocumentGenerator<WashingtonPostCollect
     doc.add(new StringField(FIELD_ID, id, Field.Store.YES));
     // This is needed to break score ties by docid.
     doc.add(new SortedDocValuesField(FIELD_ID, new BytesRef(id)));
-    doc.add(new LongPoint(FIELD_DATE, wapoDoc.getPublishDate()));
+    doc.add(new LongPoint(WapoField.PUBLISHED_DATE.name, wapoDoc.getPublishDate()));
     doc.add(new StoredField(WapoField.PUBLISHED_DATE.name, wapoDoc.getPublishDate()));
     wapoDoc.getAuthor().ifPresent(author -> {
       doc.add(new StringField(WapoField.AUTHOR.name, author, Field.Store.NO));
--- a/src/main/java/io/anserini/index/generator/WapoGenerator.java
+++ b/src/main/java/io/anserini/index/generator/WapoGenerator.java
@@ -45,6 +45,7 @@ public class WapoGenerator extends LuceneDocumentGenerator<WashingtonPostCollect
   public static final String FIELD_RAW = "raw";
   public static final String FIELD_BODY = "contents";
   public static final String FIELD_ID = "id";
+  public static final String FIELD_DATE = "published_date";
   
   private static final String PATTERN = "<.+>";
   public static final List<String> CONTENT_TYPE_TAG = Arrays.asList("sanitized_html", "tweet");
@@ -85,7 +86,8 @@ public class WapoGenerator extends LuceneDocumentGenerator<WashingtonPostCollect
     doc.add(new StringField(FIELD_ID, id, Field.Store.YES));
     // This is needed to break score ties by docid.
     doc.add(new SortedDocValuesField(FIELD_ID, new BytesRef(id)));
-    doc.add(new LongPoint(WapoField.PUBLISHED_DATE.name, wapoDoc.getPublishDate()));
+    doc.add(new LongPoint(FIELD_DATE, wapoDoc.getPublishDate()));
+    doc.add(new StoredField(WapoField.PUBLISHED_DATE.name, wapoDoc.getPublishDate()));
     wapoDoc.getAuthor().ifPresent(author -> {
       doc.add(new StringField(WapoField.AUTHOR.name, author, Field.Store.NO));
     });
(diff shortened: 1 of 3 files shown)


Chriskamphuis:
The datefilter filters out articles published after the topic article.

lintool:
@Peilin-Yang can you take a look?

src/main/java/io/anserini/index/generator/WapoGenerator.java
```diff
@@ -86,6 +86,7 @@ public Document createDocument(WashingtonPostCollection.Document wapoDoc) {
     // This is needed to break score ties by docid.
     doc.add(new SortedDocValuesField(FIELD_ID, new BytesRef(id)));
     doc.add(new LongPoint(WapoField.PUBLISHED_DATE.name, wapoDoc.getPublishDate()));
+    doc.add(new StoredField(WapoField.PUBLISHED_DATE.name, wapoDoc.getPublishDate()));
```
Peilin-Yang: I am not sure if `StoreField` is needed. Can you try to remove it to see if it still works?

src/main/java/io/anserini/rerank/lib/NewsBackgroundLinkingReranker.java
```diff
@@ -54,29 +55,43 @@ public ScoredDocuments rerank(ScoredDocuments docs, RerankerContext context) {
...
         if (computeCosineSimilarity(docsVectorsMap.get(i), docsVectorsMap.get(j)) >= 0.9) {
-          duplicates.add(j);
+          toRemove.add(j);
+        }
+      }
+    }
+
+    if(context.getSearchArgs().backgroundlinking_datefilter){
+      try{
+        Document queryDoc = reader.document(NewsBackgroundLinkingTopicReader.convertDocidToLuceneDocid(reader, queryDocId));
```
Peilin-Yang: You can use `LongPoint.newRangeQuery` here. 
See http://lucene.apache.org/core/8_2_0/core/org/apache/lucene/document/LongPoint.html#newRangeQuery-java.lang.String-long-long-

src/main/java/io/anserini/index/generator/WapoGenerator.java
Chriskamphuis: Storefield is needed: 

`An indexed long field for fast range filters. If you also need to store the value, you should add a separate StoredField instance.`

according to http://lucene.apache.org/core/8_2_0/core/org/apache/lucene/document/LongPoint.html

src/main/java/io/anserini/rerank/lib/NewsBackgroundLinkingReranker.java
Chriskamphuis: I will try this, but I expect if you do it in the initial retrieval step the results will be different as the RM3 constructed query will be different.

Peilin-Yang: This is actually interesting. 
I think we should keep the filter same for both initial ranking and the relevance feedback otherwise the results are biased. right?

arjenpdevries: If the filter models "no information from the future should be used", then the filter should be everywhere (and technically, IDF values should also depend on time...).

If it is just reflecting how judging took place, then maybe that is a different situation.

Peilin-Yang: How about let's make this consistent with what we did for microblog track? 
That is, to only filter the "future" documents in the initial ranking? 
https://github.com/castorini/anserini/blob/453a9d0307bfb331e83a8d9c48eef269379ccefb/src/main/java/io/anserini/search/SearchCollection.java#L505

Chriskamphuis: Ok the results are:

BM25 with datefilter in initial ranking:
`NDCG@5 : 0.3735`

BM25 + RM3 with datefilter in initial ranking:
`NDCG@5 : 0.3433`

BM25 + RM3 with datefilter in both initial ranking and reranking:
`NDCG@5 : 0.3444`

BM25 + RM3 with datefilter in only reranking (initial implementation)
`NDCG@5 : 0.4171`

Chriskamphuis: So it seems that datefilter in initial ranking is good when no reranking is applied, however when trying to rerank the query construction of RM3 does suffer from missing the documents that are filtered out.

Chriskamphuis:
More a general comment on this PR: The reason for the big increase in performance is also likely due to the fact that initially the task guidelines stated that articles could only relevant if they were published before the topic article; resulting in a bias in the judgments because the pooling consists of more articles published before the topic article.

lintool:
Do the official guidelines say anything about using "future evidence"? Obviously, in the task scenario, "documents in the future" haven't been written yet... so it's not really a "bias" in judgments... it's reflecting the reality of the task?

src/main/java/io/anserini/rerank/lib/NewsBackgroundLinkingReranker.java
lintool: Well, it depends on the answer to my question below about the task guidelines... do we want to "win" or model the task accurately? :)

Chriskamphuis:
So initially the idea was like you mentioned, it would not make sense to consider future documents relevant as they are not published yet at the moment the topic article is published. However, after discussion, it was decided that future documents _can_ be relevant:

There is time between when the article is published and when someone is reading it. In the meantime another article that is important for the context of the query article can be published. In this case that article would be considered relevant to. 

So using future evidence would make sense, it can even contain relevant documents. But it turns out that filtering out all documents that are published after the query article helps the effectiveness.

Chriskamphuis:
Re: bias in the judgement. I think because runs were submitted not taking future documents into account the pool of scored documents might be biased. (Although future docs can be relevant)

lintool:
I understand now. The key is that article publication time and user reading time may not be the same.

Since @Peilin-Yang has been doing CR, I'm happy to merge after he gives 👍 

Are we just making the date filter command-line parameters?

lintool:
@Peilin-Yang to be precise, it is because the qrels only include documents before the article time.

PR #790, opened 2019-09-05
Accurate BM25

Chriskamphuis:
Accurate BM25 as proposed by Leo Boytsov


src/main/java/io/anserini/search/similarity/AccurateBM25Similarity.java
```diff
@@ -0,0 +1,164 @@
+package io.anserini.search.similarity;
```
lintool: Can we add in standard Apache License header boilerplate?

lintool:
Please add license header, then go ahead and merge.

PR #791, opened 2019-09-06
Extract doc lengths

Chriskamphuis:
Script that produces tsv file which contains the length of all documents, the unique term count and the lossy unique term count as used in the BM25similarity class.


lintool:
Please add license header, update branch, then go ahead and merge.

PR #793, opened 2019-09-06
Replication Core18

Chriskamphuis:


lintool:
hi @Chriskamphuis the replication info should go here:

https://github.com/castorini/anserini/blob/master/src/main/resources/docgen/templates/core18.template

the docs are auto-gened by Maven as part of our build chain... thanks!

PR #796, opened 2019-09-08
Refactoring run_regression script

lintool:
Refactoring of script so that we can override collection location and number of indexing threads from the command line - this allows us to run regressions on different machines without having to manually change the YAML configurations.

### DIFF
--- a/src/main/python/run_regression.py
+++ b/src/main/python/run_regression.py
@@ -65,29 +65,43 @@ def get_index_path(yaml_data):
     return index_path
 
 
-def construct_indexing_command(yaml_data):
+def construct_indexing_command(yaml_data, args):
     """Construct the Anserini indexing command.
 
     Args:
         yaml_data (dict): The yaml config read from config file.
+        args: the command-line arguments.
 
     Returns:
         (:obj:`list` of :obj:`str`): the command as a list that can be executed by calling subprocess.call(command)
     """
     logger.info('='*10+'Indexing'+'='*10)
-    corpus_input_path = None
+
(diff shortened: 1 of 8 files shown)


src/main/python/run_regression.py
```diff
@@ -65,29 +65,43 @@ def get_index_path(yaml_data):
...
+    collection_path = None
     for input_root in yaml_data['input_roots']:
         if os.path.exists(os.path.join(input_root, yaml_data['input'])):
-            corpus_input_path = os.path.join(input_root, yaml_data['input'])
+            collection_path = os.path.join(input_root, yaml_data['input'])
             break
-    if not corpus_input_path:
+
+    # Allow command-line override of the input collection path
+    if args.collection_path != '':
```
Peilin-Yang: How about use explicit `if..else` block here? 
```
if args.collection_path: 
  collection_path = args.collection_path
else:
  for ... if break
```

src/main/python/run_regression.py
```diff
@@ -65,29 +65,43 @@ def get_index_path(yaml_data):
...
+        collection_path = args.collection_path
+
+    if not collection_path:
         raise RuntimeError("All corpus inputs are not existing, please check!")
+
+    # Read number of threads from YAML, but allow command-line override
+    threads = yaml_data['threads']
+    if args.indexing_threads != -1:
```
Peilin-Yang: ditto: `if..else`?

lintool:
@Peilin-Yang Addressed CR comments.

PR #804, opened 2019-09-13
support chinese indexing and search

Impavidity:
- Add CJKAnalyzer in the Indexing class.
- Add `language`  argument in indexing argument.
- Add `setLanguage` method and CJKAnalyzer in SimpleSearch class.
- Add experiments on NTCIR-8 ZH dataset.

TODO:

- Add more langugage support. 
- The language argument will follow ISO-639-2 code.


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -656,10 +656,12 @@ public void run() throws IOException {
     if (indexPath != null && !args.dryRun) {
 
       final Directory dir = FSDirectory.open(indexPath);
+      final CJKAnalyzer chineseAnalyzer = new CJKAnalyzer();
       final EnglishStemmingAnalyzer analyzer = args.keepStopwords ?
           new EnglishStemmingAnalyzer(args.stemmer, CharArraySet.EMPTY_SET) : new EnglishStemmingAnalyzer(args.stemmer);
       final TweetAnalyzer tweetAnalyzer = new TweetAnalyzer(args.tweetStemming);
-      final IndexWriterConfig config = args.collectionClass.equals("TweetCollection") ? new IndexWriterConfig(tweetAnalyzer) : new IndexWriterConfig(analyzer);
+      final IndexWriterConfig config = args.collectionClass.equals("TweetCollection") ?
+          new IndexWriterConfig(tweetAnalyzer) : args.language.equals("zh") ? new IndexWriterConfig(chineseAnalyzer) : new IndexWriterConfig(analyzer);
```
r-clancy: Maybe this should be a normal nested if statement to improve readability?

src/main/java/io/anserini/index/IndexArgs.java
```diff
@@ -76,6 +76,9 @@
   @Option(name = "-bm25.accurate", usage = "Switch to use the accurate BM25 similarity)")
   public boolean bm25Accurate = false;
 
+  @Option(name = "-language", usage = "boolean to use CJK analyser while indexing")
```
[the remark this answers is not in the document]
lintool: Either actually implement the ISO-639-2 codes or change the option to, say, `-useCJK`?

src/main/java/io/anserini/index/IndexCollection.java
lintool: +1

lintool:
@Impavidity also, Travis failed...

src/main/java/io/anserini/search/topicreader/TsvStringKeyTopicReader.java
```diff
@@ -0,0 +1,61 @@
...
+ * 174249 does xpress bet charge to deposit money in your account
+ * 320792 how much is a cost to run disneyland
+ * 1090270  botulinum definition
+ * 1101279	do physicians pay for insurance from their salaries?
+ * 201376 here there be dragons comic
+ * 54544  blood diseases that are sexually transmitted
+ * ...
+ * </pre>
+ */
+public class TsvStringKeyTopicReader extends TopicReader<String> {
```
lintool: This is mostly duplicate with https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/topicreader/TsvTopicReader.java

Seems like the only different is either treating the key as an int or the key as a string...

I don't know if the distinction is important, but let's keep it for now.

So, refactor: `TsvTopicReader` -> `TsvIntTopicReader`, this class to `TsvStringTopicReader`?

src/main/resources/topics-and-qrels/topics.ntcir8zh.eval.txt
```diff
@@ -0,0 +1,73 @@
+ACLIA2-CS-0002	《千里走单骑》和张艺谋是什么关系？
```
lintool: Can we make the qrels and topics have the same name? Why `ntcir8zh` vs. `nctir8` above?

src/main/python/clir/convert_collection_to_jsonl.py
```diff
@@ -0,0 +1,95 @@
+# -*- coding: utf-8 -*-
```
lintool: Why do we need this?
The collection is in TREC doc format... the existing collection should be able to parse it?
We should avoid rewriting collections unless absolutely necessary...

src/main/java/io/anserini/search/topicreader/TsvStringKeyTopicReader.java
```diff
@@ -0,0 +1,61 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+/**
+ * Topic reader for queries in tsv format, such as the MS MARCO queries.
```
lintool: Please fix the comments.

docs/experiments-clir.md
```diff
@@ -0,0 +1,52 @@
+# Cross-lingual Information Retrieval Experiments
```
lintool: See https://github.com/castorini/anserini/blob/master/src/main/resources/docgen/templates/robust04.template for an example of how this is actually done.

Please make consistent?

Title should be collection-specific (e.g., NTCIR).

lintool:
@Impavidity bunch of comments for you to address... since we're going to build on this foundation, let's get it right...

src/main/java/io/anserini/search/topicreader/TsvStringKeyTopicReader.java
Impavidity: Any reason we want to use int instead of string?

lintool: I think it's the sort order? 100 sorts before 99 based on String.

Actually, you know what? Refactor into a link list map, and let's just preserve whatever order comes in the topic file.

That way we don't need two separate implementations.

The key can be String in this case.

What do you think?

src/main/python/clir/convert_collection_to_jsonl.py
Impavidity: Is this trec doc format ?
<DOC>
<DOCNO> XIN_CMN_20020101.0001 </DOCNO>
    新华社一月一日发稿目录(二)
</DOC>

PR #806, opened 2019-09-15
A script to convert Yahoo Answers collection.

searchivarius:
A couple of notes. 
1. It's not super lightweight, b/c I use numpy. Sorry, I was lazy to reimplement sampling without replacement in Python.
2. The relevance grade is either 4 (for best answers) or 3. I believe this is compatible with the TREC NIST gdeval.pl utility.

### DIFF
--- a/src/main/python/yahoo_answers/convert_collection_to_jsonl.py
+++ b/src/main/python/yahoo_answers/convert_collection_to_jsonl.py
@@ -1,0 +1,211 @@
+# -*- coding: utf-8 -*-
+"""
+Anserini: A Lucene toolkit for replicable information retrieval research
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


searchivarius:
@Chriskamphuis I will make it compliant.

searchivarius:
Hi @Chriskamphuis I fixed the script, but not types of formatting. Are you sure it's against pep8? pep8 checker doesn't complain about it.

Chriskamphuis:
You're right, that's not PEP8.

searchivarius:
@Chriskamphuis thanks for checking. BTW, what do you use to format according to pep8 standards?

Chriskamphuis:
@searchivarius `flake8` which you can also install as a vim plugin. But for bigger projects I use pycharm, which has pep8 checking build in.

PR #810, opened 2019-09-17
A script to convert Yahoo Answers collection

searchivarius:


lintool:
I assume a "guide" like https://github.com/castorini/anserini/blob/master/docs/experiments-msmarco-passage.md is coming?

searchivarius:
Sure @lintool I will add one.

PR #834, opened 2019-10-24
Update approximate-nearestneighbor.md

lintool:


tteofili:
+1, thanks @lintool it looks much better now

PR #838, opened 2019-10-25
Minor documentation tweak, forgot to check in as part of #837

lintool:
Ref: #837


lintool:
This is indexing only... the corpus only has 1 file, and Anserini indexing is one thread per file... so 40 threads means 39 are not doing anything :)

rodrigonogueira4:
I see, th

> This is indexing only... the corpus only has 1 file, and Anserini indexing is one thread per file... so 40 threads means 39 are not doing anything :)

I see, thanks!

Issue #819, opened 2019-10-02
Add integation test for Elastirini


r-clancy:
We need to create an end-to-end integration testing script written in Python that: sets up Solr, indexes a number of collections, runs search and outputs the trec_eval results, and tears down Solr.

The name of the collections to run on should be configurable (the `generator` and `collection` can be fixed like [here](https://github.com/osirrc/anserini-docker/blob/master/index)) and we should be able to either specify an existing Solr server (i.e., via a `--solr` parameter) or setup/teardown a new one (as above) based on the Solrini instructions.

lintool:
For this, we need a non-Docker solution to starting up ES. Once we've figured this out, docs [here](https://github.com/castorini/anserini/blob/master/docs/elastirini.md) should be updated to provide an alternative.

lintool:
Might be useful: https://www.elastic.co/guide/en/elasticsearch/reference/current/integration-tests.html

lintool:
Closed by #911

PR #841, opened 2019-10-26
Elastirini ms passage update

w329li:


lintool:
Can we rewrite the documentation to reduce redundancy?

You can simply refer to https://github.com/castorini/anserini/blob/master/docs/experiments-msmarco-passage.md for data prep.

Actually, why aren't these instructions just integrated into https://github.com/castorini/anserini/blob/master/docs/elastirini.md directly?

w329li:
I removed the elastic-msmarco-passage.md and integrated new instruction directly into Elastirini.md file

lintool:
I'll let @r-clancy take a look!

r-clancy:
I'll run through this on Monday, validate I can replicate, and merge if it looks good.

lintool:
@r-clancy or whoever does the merging, please write a more informative commit message.

@w329li #819 next.

r-clancy:
@w329li the expected results are:
```
map                   	all	0.1957
recall_1000           	all	0.8573
```

but I get:
```
map                all     0.1956
recall_1000    all     0.8573
```

Small difference, but it doesn't seem to be re-producible. I can supply the run files for you to compare against your results.

w329li:
I think it is because we have different version of Lucene. Check the table at the bottom of this link(https://github.com/castorini/anserini/blob/master/docs/experiments-msmarco-passage.md), it explains the slight difference of results.

r-clancy:
> I think it is because we have different version of Lucene. Check the table at the bottom of this link(https://github.com/castorini/anserini/blob/master/docs/experiments-msmarco-passage.md), it explains the slight difference of results.

I am using the latest Anserini, so version 8.0.0 of Lucene - the results should match second-from-last able.

w329li:
testing Elastirini on Msmarco-doc with Apoorv but we both have same results that different from expected result. Trying to find reason behind it , and also working on #819 at same time.

lintool:
@w329li nudge?

lintool:
Another bump on this... @r-clancy @x389liu 
Could it be due to https://dl.acm.org/citation.cfm?doid=3331184.3331339 ?

r-clancy:
That might be a likely culprit @lintool, as the results are only `0.0001` off I'd be fine with pointing to that paper in the README and merging this.

I had asked @w329li and @apoorvp16 to look into the query tokenization and document scores between Lucene and Elastic, but haven't heard anything back yet - how's that going guys?

w329li:
> That might be a likely culprit @lintool, as the results are only `0.0001` off I'd be fine with pointing to that paper in the README and merging this.
> 
> I had asked @w329li and @apoorvp16 to look into the query tokenization and document scores between Lucene and Elastic, but haven't heard anything back yet - how's that going guys?

w329li:
I double -checked the command and re-run it with dubug file @r-clancy sent. It is still the same result(single decimal difference)

lintool:
Are we abandoning this PR? I thought @r-clancy thought this was good to go re: merging?

r-clancy:
I was happy with it, not sure why it was closed - @w329li ?

w329li:
Just compared indexing and searching result between Lucene and Elastirini. Ii think the reason behind this is: some of the data content are empty, so they are not indexed in Elastirini, these group of data might affect search and eval process. However this does not happen in Lucene indexing step.
i remember Solr also have same result (empty content will not be indexed). However Lucene will index all content even thought they are  empty. Since there is only a very small group of empty content in Msmarco-passage, it does not make too much difference. Thats why it results in a 0.0001 difference.

Issue #845, opened 2019-11-01
Solr config tweaks: "contents" should be searchable and stored


lintool:
`dstlr` wants to tweak Solr configs: https://github.com/dstlry/dstlr#download-and-build-anserini

```
sed -i.bak 's/field name="contents" type="text_en_anserini" indexed="true" stored="false" multiValued="false"/field name="contents" type="text_en_anserini" indexed="true" stored="true" multiValued="false"/g' src/main/resources/solr/anserini/conf/managed-schema
```

Why wouldn't we just want to change in Anserini directly? @r-clancy 

Hey @x389liu can you send a PR?

r-clancy:
This comes down to the difference between `-storeRawDocs` (e.g., raw HTML) and `-storeTransformedDocs` (e.g., text from HTML page). Indexing with Lucene, this can be specified in code ([raw](https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/generator/LuceneDocumentGenerator.java#L130) and [transformed](https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/generator/LuceneDocumentGenerator.java#L158)) but with Solr this is done via modifying the schema ([raw](https://github.com/castorini/anserini/blob/master/src/main/resources/solr/anserini/conf/managed-schema#L518) and [transformed](https://github.com/castorini/anserini/blob/master/src/main/resources/solr/anserini/conf/managed-schema#L521)).

We can change this in Anserini if we want Solr to store transformed docs by default, which might make more sense as it's the "interactive" back-end document store. @lintool

lintool:
@r-clancy wouldn't it make sense for Solr to store _both_? The raw docs are useful for actual presentation (e.g., HTML rendering), and the transformed docs are useful for downstream NLP.

With Anserini, depending on application, we allow the user to choose... and they can always get the transformed from the raw by just rerunning the transformation, right?

In Solr, we want to support _both_ viewing and downstream applications? Unless there's a way to apply the transformer from the Solr end?

r-clancy:
@lintool I'm fine with having them both, it just means the indexing will take a bit longer with Solr.

PR #847, opened 2019-11-01
Make "contents" be searchable and stored

x389liu:
PR for issue #845


src/main/resources/solr/anserini/conf/managed-schema
```diff
@@ -518,7 +518,7 @@
   <field name="raw" type="text_general" indexed="false" stored="true" multiValued="false" />
   
   <!-- "contents" searchable but not stored. -->
```
lintool: Forgot to change comment...

PR #852, opened 2019-11-03
Fixes error when input file has non-ASCII characters

rodrigonogueira4:


lintool:
@rodrigonogueira4 does this have any impact on MMR? If so, we should update regressions as well...

rodrigo-eai:
> @rodrigonogueira4 does this have any impact on MMR? If so, we should update regressions as well...

Good point, I didn't measure MRR but I will do it.

rodrigonogueira4:
I confirm that I've ran it on the dev queries and got the same MRR@10 from https://github.com/castorini/anserini/blob/master/docs/experiments-msmarco-passage.md:
0.1874

PR #859, opened 2019-11-09
Added ability to read topics directly from jar

lintool:
Also cleaned up resource inclusion in jar; skips CACM collection.


lintool:
@emmileaf Thanks! I'll merge myself.

PR #872, opened 2019-11-13
Add Indexing file for CORE corpus

### LINKED ISSUE #856: Write indexer for CORE Dataset
We need to write an indexer for the [CORE dataset](https://core.ac.uk/services/dataset/).

We can base on the current JSON collection: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/collection/JsonCollection.java

bazingagin:
to close #856

### DIFF
--- a/src/test/java/io/anserini/collection/CoreDocumentTest.java
+++ b/src/test/java/io/anserini/collection/CoreDocumentTest.java
@@ -32,108 +32,100 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 
-// A file in a JsonCollection can either be:
-// (1) A single JSON object (i.e., a single document)
-// (2) An array of JSON objects
-// (3) JSON Lines (i.e., one JSON object per line)
-//
-// This is the test case for (3)
-//
-// Note that we're testing the multifield capability here and only here, since the codepath is shared.
 public class CoreDocumentTest extends DocumentTest {
 
-    @Before
-    public void setUp() throws Exception {
-        super.setUp();
+  @Before
+  public void setUp() throws Exception {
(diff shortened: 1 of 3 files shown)


lintool:
Hi @bazingagin can you please add a simple test?

E.g., store a small sample of `.xz` file in `src/test/resources/sample_docs/` somewhere and write some test cases that demonstrate parsing the collection?

Thanks!

lintool:
Actually, can you follow the template here: https://github.com/castorini/anserini/blob/master/src/test/java/io/anserini/collection/JsonLineObjectTest.java

(Which doesn't involve checking `xz` data into the repo.

src/test/java/io/anserini/collection/CoreDocumentTest.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+// A file in a JsonCollection can either be:
+// (1) A single JSON object (i.e., a single document)
+// (2) An array of JSON objects
+// (3) JSON Lines (i.e., one JSON object per line)
+//
+// This is the test case for (3)
+//
+// Note that we're testing the multifield capability here and only here, since the codepath is shared.
```
lintool: Comment doesn't make sense in this context? Just copy and paste, right?
Either rewrite or remove.

src/test/java/io/anserini/collection/CoreDocumentTest.java
```diff
@@ -0,0 +1,139 @@
...
+public class CoreDocumentTest extends DocumentTest {
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
```
lintool: Change to two-space indent please.

Issue #368, opened 2018-07-27
Confusion HtmlCollection: does it accept tgz?


lintool:
The javadoc says:
https://github.com/castorini/Anserini/blob/master/src/main/java/io/anserini/collection/HtmlCollection.java#L37

> The file name (excluding the extension) will be the docid and the stripped contents will be the contents.

However, the impl accepts `tgz` files. So what's the behavior when indexing `tgz` files? What's used as the docid? Some clarification seems to be needed?

Peilin-Yang:
The docid will still be the filename (the files in the tar archive).

lintool:
This is lower priority, but we should add a test case... things that aren't tested or part of regressions that we run regularly have a tendency to break...

lintool:
Closed by #873

PR #873, opened 2019-11-13
Added tests for HtmlCollection

lintool:


lintool:
Note that we can index CACM without decompressing the collection.

lintool:
This PR will close #368 and

lintool:
hi @bazingagin - starting to involve you... please help w/ my CR also? thanks!

src/test/java/io/anserini/collection/HtmlCollectionTest.java
```diff
@@ -0,0 +1,44 @@
...
+public class HtmlCollectionTest {
+
+  @Test
+  public void testCACM() throws IOException {
+    HtmlCollection cacm = new HtmlCollection();
+    FileSegment<HtmlCollection.Document> segment =
+        cacm.createFileSegment(Paths.get("src/main/resources/cacm/cacm.tar.gz"));
+    Iterator<HtmlCollection.Document> iter = segment.iterator();
+
+    AtomicInteger cnt = new AtomicInteger();
```
r-clancy: Is there some parallelism that's going on here so that we need `AtomicInteger`? Why not just an `int`?

lintool: I learned this a while ago, but don't remember the exact details. It relates to auto boxing/unboxing of int/Integers in Java.

You obviously can't do `i++`, the `Integer` object doesn't have an increment method. `AtomicInteger` ensures we have a wrapper object, and we can increment the value inside.

PR #875, opened 2019-11-15
Remove evaluation code

lintool:
There's an entire evaluation package in anserini that we don't use. Refactoring out.

Test cases check the run files directly instead of checking eval metrics.


lintool:
haha @Peilin-Yang :)

PR #877, opened 2019-11-17
SimpleSearch refactoring in coordination with Pyserini

lintool:
hi @emmileaf the broader context of this PR is that I'm adjusting the Pyserini bindings on the Java end.

+ Makes BM25 default
+ Makes the ability to switch similarities actually work :)
+ Adds the ability to unset RM3
+ Removes a bunch of bindings to models we don't use


lintool:
@emmileaf Thanks! I have one more minor modification to push, and then I'll merge.

lintool:
Yes, correct - lots of changes on the Pyserini side.
Currently local and on branch. I will consolidate and send PR from that end.

PR #882, opened 2019-11-19
Added test case demonstrating how to get term positions out of the index

lintool:
Helping @cmacdonald on this...


lintool:
hey @zeynepakkalyoncu I've merged, but you should still take a look - we might want to expose this from Python.

PR #890, opened 2019-11-23
Port IndexUtil in Pyserini

alirezamirzaeiyan:
This piece of code is changed for pyserini.


lintool:
Thanks for your contribution @alirezamirzaeiyan !

If you can describe how you're using Anserini/Pyserini, perhaps we can coordinate feature development?

PR #893, opened 2019-11-25
Update IndexReaderUtils functions

### LINKED ISSUE #883: Refactor IndexUtils to facilitate porting to Python

zeynepakkalyoncu:
- Moves dumpDocumentVectors into IndexReaderUtils for consistency
- Adapts printTermCounts and printDocumentVector to be used with Pyserini (includes term counts, document vector, postings lists)

Resolves #883


lintool:
hey @zeynepakkalyoncu -

I think `Map<Integer, Integer> getPostingsList` should probably be something like `List<Posting> getPostingsList`.

Create a specific `Posting` class, should have methods like:

+ `int getTf()`
+ `int getDocid()`
+ `int[] getPositions()` (if available)

lintool:
Also, we need test cases!

src/main/java/io/anserini/index/Posting.java
```diff
@@ -0,0 +1,54 @@
+/**
```
lintool: I would make this a static inner class of `IndexReaderUtils` since it's not supposed to have independent existence. We don't want to cause confusion with native Lucene classes.

src/main/java/io/anserini/index/Posting.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.lucene.index.PostingsEnum;
+
+import java.io.IOException;
+
+/*
+ Basic Posting class used to construct postings lists in IndexReaderUtils
+ */
+
+public class Posting {
+    private int docId;
```
lintool: two space indent place.

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -17,23 +17,223 @@
...
+      GzipCompressorInputStream gzIn = new GzipCompressorInputStream(in);
+      return gzIn;
+    } else if (path.endsWith(".zip")) {
+      GzipCompressorInputStream zipIn = new GzipCompressorInputStream(in);
+      return zipIn;
+    }
+    return in;
+  }
+
+  public static Map<String, String> getTermCounts(IndexReader reader, String termStr) throws IOException, ParseException {
```
lintool: Let's separate into two methods? `analyzeTerm` to return the stemmed form?
Then, `getTermCounts` can return a`Map<String, Long>`, so we don't need turn numbers into strings...

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -17,23 +17,223 @@
...
+    EnglishAnalyzer ea = new EnglishAnalyzer(CharArraySet.EMPTY_SET);
+    QueryParser qp = new QueryParser(LuceneDocumentGenerator.FIELD_BODY, ea);
+    TermQuery q = (TermQuery)qp.parse(termStr);
+    Term t = q.getTerm();
```
lintool: I think Java convention has space after cast, so `(TermQuery) qp.parse(termStr)`

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -17,23 +17,223 @@
...
+
+    Map<String, String> termInfo = Map.ofEntries(
+            Map.entry("rawTerm", termStr),
```
lintool: This would be a continuation indent, so 4 spaces.

PR #896, opened 2019-11-26
Refactored code to extract doc lengths and norms

lintool:
+ Fixed issues w/ code to extract doc lengths from doc vectors
+ Added code to extract norms directly from index

### DIFF
--- a/src/main/java/io/anserini/util/ExtractNorms.java
+++ b/src/main/java/io/anserini/util/ExtractNorms.java
@@ -18,18 +18,24 @@ package io.anserini.util;
 
 import org.apache.lucene.index.DirectoryReader;
 import org.apache.lucene.index.IndexReader;
-import org.apache.lucene.index.Terms;
+import org.apache.lucene.index.LeafReader;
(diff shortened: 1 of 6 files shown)


src/main/java/io/anserini/util/ExtractNorms.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    Directory dir = FSDirectory.open(Paths.get(myArgs.index));
+    IndexReader reader = DirectoryReader.open(dir);
+
+    PrintStream out = new PrintStream(new FileOutputStream(new File(myArgs.output)));
+    out.println("docid\tnorm");
+    for (LeafReaderContext context : reader.leaves()) {
+      LeafReader leafReader = context.reader();
+      NumericDocValues docValues = leafReader.getNormValues("contents");
```
jpountz: you might want to check docValues for null here to return a better error if the field doesn't exist, or if it has been indexed without norms

lintool:
@jpountz thanks for the CR!

PR #897, opened 2019-11-26
Compute accurate BM25 weights

lintool:


zeynepakkalyoncu:
Again for my own clarification, which tests go under BasicIndexOperationsTest and which ones under IndexReaderUtilsTest?

lintool:
Actually, good call, this is a method in `IndexReaderUtils` so it should go under `IndexReaderUtilsTest`. Will send update.

PR #898, opened 2019-11-26
Add method to get IndexReader

zeynepakkalyoncu:
We need to be able to instantiate an IndexReader directly in Pyserini without interacting with Lucene


lintool:
Flaky test case. Will look into it, but merging in the meantime.

PR #902, opened 2019-11-27
Multilingual regression fixes

lintool:
@Impavidity this is all on `tuna`.

This one works fine:

```
nohup python src/main/python/run_regression.py --collection clef06-fr --index >& log.clef06-fr &
```

These work fine, but I caught a few typos:

```
nohup python src/main/python/run_regression.py --collection fire12-bn --index >& log.fire12-bn &
nohup python src/main/python/run_regression.py --collection fire12-hi --index >& log.fire12-hi &
nohup python src/main/python/run_regression.py --collection fire12-en --index >& log.fire12-en &
```

This appears to be broken:

```
nohup python src/main/python/run_regression.py --collection trec02-ar --index >& log.trec02-ar &
```

Also, the generated regression docs are different.

Can you please check?


lintool:
@Impavidity verified that AR fixed.

PR #905, opened 2019-11-29
Python integration test

w329li:
Made following updates:
1. Input format updated. user need to input collection name, and collection path. There are also
  some optional inputs(user can choose to turn off Solr server at the end)
2. program execute "restart" command to ensure we will turn on Solr server in a safe way no
    matter it is on or off previously
3.program will check whether we have correct path when we execute a new command
4. program will check whether the collection already exists on Solr server. If it exists, check
    whether it has correct number of indexed document, then let user decide whether skip
    indexing step or overwrite this collection
5.An assertion for trec_eval result. If eval result(MPA,P_30) is different from expected result, we 
   throw exception 
6. make all string quote consistent. All codes have appropriate length at each line


src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+        command2 = "sh target/appassembler/bin/IndexCollection -collection {}  -generator {} -threads {} -input {} \
+            -solr -solr.index {} -solr.zkUrl localhost:9983  -storePositions -storeDocvectors -storeTransformedDocs \
+            ".format(collection,generator,thread_num,input_path,name)
+        self.command_execution("target/appassembler/bin/IndexCollection", command2)   
+        ###subprocess.run(command2, shell = True)  
+
+    # if data collection already exists on solr server, check whether it has correct number of indexed document
+    # return true/false
+    def check_indexing(self):
+        subprocess.run("wget http://localhost:8983/solr/core18/query?q=*:*",shell=True)
```
lintool: don't use `wget` in a process... use something like `requests`: https://2.python-requests.org//en/latest/

lintool:
@w329li need to resolve conflict.

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+            content = f.read().splitlines()   
+        res = []
+        for keyword in keywords:
+            for n in content:
+                if (keyword in n):res.append(n)
+    print(res) 
+    return res
+
+
+class Solr_command:
```
lintool: wouldn't `SolrClient` be a better name?

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+        "generator": "JsoupGenerator", 
+        "thread_num": 8,
+        "topic_reader":"Trec",
+        "topic_path":"src/main/resources/topics-and-qrels/topics.robust04.txt",
+        "qrel_path":"src/main/resources/topics-and-qrels/qrels.robust04.txt"
+    }
+}
+
+# keyword_search works with trec_eval() to extract actual MAP, P_30 result from "this_is_temp_file.txt" file
+def keyword_search(keywords):
```
lintool: wouldn't `read_file_into_dictionary` or something like that be a better name?

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+    return res
+
+
+class Solr_command:
+    def __init__(self,name,input_path):
+        if (name == None): raise Exception("Need collection name, but receive None")
+        if (input_path == None): raise Exception("Need path of collection, but receive None")
+        self.name = name 
+        self.input_path = input_path
+        self.output_path = "run.solr.{}.bm25.topics.{}.301-450.601-700.txt".format(name,name)
```
lintool: Remove the `301-450.601-700` part.

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+    def command_execution(self, command_path, command):
+        print(command)
+        if (os.path.exists(command_path) == False):
+            sys.exit("default solr path: {} does not exist! Cannot execute the command. Please check:\n \
+            1.Solr is installed under correct path \
+            2.You run this program under correct path".format(command_path))    
+        output = subprocess.run(command, shell = True)
+
+    
+    def turn_on_server(self):
```
lintool: how about `startup`? or `start`?

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+        print("we turn on solr server")
+        #### we check whether we have correct solr path first
+        solr_on = "solrini/bin/solr restart -c -m 8G"
+        solr_config = "pushd src/main/resources/solr && ./solr.sh ../../../../solrini localhost:9983 && popd"
+        self.command_execution("solrini/bin/solr", solr_on)
+        self.command_execution("src/main/resources/solr", solr_config)  
+        print("Solr server is on")
+
+
+    def turn_off_server(self):
```
lintool: how about `shutdown` or `stop`?

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+        subprocess.run("solrini/bin/solr stop -all",shell=True)        
+
+    # we consider 9983 as default local port number
+    # if this collection already exists on solr and has correct number of indexed document from previous test,we ask user 
+    # whether overwrite it, final MAP, P_30 result will be different for some reason. Thus we delete it and re-build it   
+    def indexing(self,name,input_path,thread_num):
```
lintool: how about `index`? method names should be verbs.

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+    def check_indexing(self):
+        subprocess.run("wget http://localhost:8983/solr/core18/query?q=*:*",shell=True)
+        # it takes time to download the file,so we run a loop to make sure 
+        while(os.path.exists("query?q=*:*") == False): pass
+        ##time.sleep(5)
+        with open("query?q=*:*") as json_file:
+            indexing_result = json.load(json_file)
+        print("Expected indexed number: 595037, actual indexed number", indexing_result["response"]["numFound"])
+        total_indexed_doc = int(indexing_result["response"]["numFound"])
+        subprocess.run("rm -rf query\?q\=\*\:\*", shell = True)
```
lintool: don't shell out for this. use Python commands.

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+        return total_indexed_doc == 595037
```
lintool: but what if it's a different collection?

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+        total_indexed_doc = int(indexing_result["response"]["numFound"])
+        subprocess.run("rm -rf query\?q\=\*\:\*", shell = True)
+        return total_indexed_doc == 595037 
+
+
+    
+    # we consider 8983 as default local port number
+    def searching(self,name,topic_reader, topic_path):
+        command = "sh target/appassembler/bin/SearchSolr -topicreader {} -solr.index {} -solr.zkUrl localhost:9983 -topics {} \
+                -output run.solr.{}.bm25.topics.{}.301-450.601-700.txt""".format(topic_reader,name,topic_path,name,name)
```
lintool: if you're specifying output file here, why do you have `self.output_path`?

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+        print("\nWe start trec_eval: \n")
+        command = "eval/trec_eval.9.0.4/trec_eval -m map -m P.30 {} {}".format(qrel_path,output_path)
+        self.command_execution("eval/trec_eval.9.0.4/trec_eval", command)   
+        save_result = command + " > this_is_temp_file.txt"    
+        subprocess.run(save_result , shell = True)
+        keywords = ["map", "P_30"]   
+        res = keyword_search(keywords)
+        MAP,P_30 = res[0].split()[-1], res[1].split()[-1]    
+        idea_MAP,idea_P_30 = index_options[self.name]["MAP"],index_options[self.name]["P30"]
+        if (MAP != idea_MAP): raise Exception("MAP: {}, which is not equal to theoritical MAP value {} ".format(MAP,idea_MAP))
```
lintool: typo, and I think you actually mean "ground truth"?

src/main/python/solr_integration_test.py
```diff
@@ -0,0 +1,210 @@
...
+    # if this collection already exists on solr and has correct number of indexed document from previous test,we ask user 
+    # whether overwrite it, final MAP, P_30 result will be different for some reason. Thus we delete it and re-build it   
+    def indexing(self,name,input_path,thread_num):
+        thread_num = str(thread_num)
+        # if this collection already exists
+        collection_location = "solrini/server/solr/" + name + "_shard1_replica_n1"  
+        # if check_indexing return False, it mean previous indexed doc is incorrect,then we should overwrite this collection
+        # re-execute indexing, searching step.
+        if (os.path.exists(collection_location) and self.check_indexing() ):
+            response = input("collection already exists on Solr server and indexed doc number is same as expectancy\n \
```
lintool: this script should not be interactive - this should be controlled via a command line option.

lintool:
Lots of commands. More, but please address this first.

w329li:
Thanks for suggestions, I will finish all them today!

lintool:
@w329li you still need to resolve conflicts with master... just replace with this version?

w329li:
Updates is made, conflict should be resolved now.

PR #907, opened 2019-11-29
Refactoring of the Elastirini documentation

lintool:
Tweaked schema mappings and placed configs in their own files.
Verified Robust04 and MS MARCO passage results.


r-clancy:
LGTM - I am able to replicate the results.

Using the other PR with `"fielddata": true` I was also able to replicate the `robust04`, strange it was different for you.

PR #916, opened 2019-12-05
Refactoring indexer: cleaned up options and logging

lintool:
Per thread logging is very verbose for collections that have lots of file segments. Added option to control, off by default.

Got ride of dryrun option... never used; simplifies code path.


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -229,8 +229,10 @@ public void run() {
               inputFile.getFileName().toString() + ": error iterating through segment.");
         }
 
-        LOG.info(inputFile.getParent().getFileName().toString() + File.separator +
-            inputFile.getFileName().toString() + ": " + cnt + " docs added.");
+        if (verbose) {
```
r-clancy: Instead of marking up the code with if statements for verbosity, maybe move the finer-grained logging to the DEBUG level and have a `-verbose` flag that sets the logger to this level?

lintool: Good call. Will do.

lintool:
Hi @r-clancy - I adopted your suggestion of setting different log levels.

I was going through the code again, and decided to remove the "unindexed" counter. The rationale here is that the counter code path was really confusing. Some things were being doubled counted - for example, a "skipped" doc inside a generator was also counted as an "unindexed". But a whitelisted doc was counted as "skipped" but not "unindexed".

With "unindexed" counts removed, everything is only counted in one category, which I believe is the desired behavior.

What do you think?

Also added more tests.

lintool:
@x389liu @bazingagin can you follow along also please to start learning about Anserini's indexing internals?

r-clancy:
> I was going through the code again, and decided to remove the "unindexed" counter. The rationale here is that the counter code path was really confusing. Some things were being doubled counted - for example, a "skipped" doc inside a generator was also counted as an "unindexed". But a whitelisted doc was counted as "skipped" but not "unindexed".
> 
> With "unindexed" counts removed, everything is only counted in one category, which I believe is the desired behavior.
> 
> What do you think?

Agreed, the counters are less confusing now.

LGTM, verbose/quiet logging is much cleaner and simpler now.

lintool:
hi @r-clancy I rewrote all the index usage descriptions to be consistent grammatically, with correct metavars, etc.

I rewrote all `-es` options to begin with "Elasticsearch" and all `-solr` options to begin with "Solr" so it's very clear.

A second pair of eyes would be appreciated.

```
$ target/appassembler/bin/IndexCollection
Option "-collection" is required
 -bm25.accurate               : Boolean switch to use AccurateBM25Similarity (computes accurate
                                document lengths). (default: false)
 -collection [class]          : Collection class in package 'io.anserini.collection'.
 -es                          : Indexes into Elasticsearch. (default: false)
 -es.batch [n]                : Elasticsearch batch index requests size. (default: 1000)
 -es.connectTimeout [ms]      : Elasticsearch (low level) REST client connect timeout (in ms).
                                (default: 600000)
 -es.hostname [host]          : Elasticsearch host. (default: localhost)
 -es.index [name]             : Elasticsearch index name.
 -es.password [password]      : Elasticsearch password. (default: changeme)
 -es.poolSize [num]           : Elasticsearch client pool size. (default: 10)
 -es.port [port]              : Elasticsearch port number. (default: 9200)
 -es.socketTimeout [ms]       : Elasticsearch (low level) REST client socket timeout (in ms).
                                (default: 600000)
 -es.user [username]          : Elasticsearch user name. (default: elastic)
 -generator [class]           : Document generator class in package 'io.anserini.index.generator'.
 -index [path]                : Index path.
 -input [path]                : Location of input collection.
 -keepStopwords               : Boolean switch to keep stopwords. (default: false)
 -language [language]         : Analyzer language (ISO 3166 two-letter code). (default: en)
 -memorybuffer [mb]           : Memory buffer size (in MB). (default: 2048)
 -optimize                    : Boolean switch to optimize index (i.e., force merge) into a single
                                segment; costly for large collections. (default: false)
 -quiet                       : Turns off all logging. (default: false)
 -shard.count [n]             : Number of shards to partition the document collection into.
                                (default: -1)
 -shard.current [n]           : The current shard number to generate (indexed from 0). (default: -1)
 -solr                        : Indexes into Solr. (default: false)
 -solr.batch [n]              : Solr indexing batch size. (default: 1000)
 -solr.commitWithin [s]       : Solr commitWithin setting (in seconds). (default: 60)
 -solr.index [name]           : Solr index name.
 -solr.poolSize [n]           : Solr client pool size. (default: 16)
 -solr.zkChroot [path]        : Solr ZooKeeper chroot (default: /)
 -solr.zkUrl [urls]           : Solr ZooKeeper URLs (comma separated list).
 -stemmer [stemmer]           : Stemmer: one of the following {porter, krovetz, none}; defaults to
                                'porter'. (default: porter)
 -storeDocvectors             : Boolean switch to store document vectors; needed for (pseudo)
                                relevance feedback. (default: false)
 -storePositions              : Boolean switch to index store term positions; needed for phrase
                                queries. (default: false)
 -storeRawDocs                : Boolean switch to store raw document text. (default: false)
 -storeTransformedDocs        : Boolean switch to store transformed document text. (default: false)
 -threads [num]               : Number of indexing threads.
 -tweet.deletedIdsFile [file] : File that contains deleted tweet ids (longs), one per line; these
                                tweets will be skipped during indexing. (default: )
 -tweet.keepRetweets          : Boolean switch to index retweets. (default: false)
 -tweet.keepUrls              : Boolean switch to keep URLs. (default: false)
 -tweet.maxId [id]            : Max tweet id to index (long); all tweets with larger tweet ids will
                                be skipped. (default: 9223372036854775807)
 -tweet.stemming              : Boolean switch to apply Porter stemming while indexing tweets.
                                (default: false)
 -uniqueDocid                 : Removes duplicate documents with the same docid during indexing.
                                This significantly slows indexing throughput but may be needed for
                                tweet collections since the streaming API might deliver a tweet
                                multiple times. (default: false)
 -verbose                     : Enables verbose logging for each indexing thread; can be noisy if
                                collection has many small file segments. (default: false)
 -whitelist [file]            : File containing list of docids, one per line; only these docids
                                will be indexed.
Example: IndexCollection -collection [class] -generator [class] -input [path] -threads [num]
```

PR #921, opened 2019-12-07
Expose iterator over terms in IndexReaderUtils

### LINKED ISSUE #912: Expose iterator over terms in dictionary in IndexReaderUtils
See: https://github.com/castorini/anserini/blob/master/src/test/java/io/anserini/index/BasicIndexOperationsTest.java#L54

We should probably expose a method in IndexReaderUtils to provide an iterator over terms - so that we can access from Pyserini. From this we can do the standard log-log counts of df and related plots...

zeynepakkalyoncu:
Resolves #912 

Also renames the analyzeTerm function with the analyzer parameter to solve Pyjnius issue in Pyserini (unable to differentiate between methods by signature only)


src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -165,7 +165,7 @@ public static IndexReader getReader(String path) throws IOException {
    * @return stemmed form of the term or <code>null</code> if error encountered
    */
   public static String analyzeTerm(String term) {
-    return analyzeTerm(term, DEFAULT_ANALYZER);
+    return analyzeTermWithAnalyzer(term, DEFAULT_ANALYZER);
```
lintool: Why rename? Overloading seems fine?

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -198,6 +198,15 @@ public static String analyzeTerm(String term, Analyzer analyzer) {
     return termInfo;
   }
 
+  /**
+   * Returns iterator pointing to the first term in index
```
lintool: "Returns iterator over all terms in the collection."

More accurate?

src/test/java/io/anserini/index/IndexReaderUtilsTest.java
```diff
@@ -79,6 +81,36 @@ public void testTermCounts() throws Exception {
...
+    IndexReader reader = DirectoryReader.open(dir);
+
+    TermsEnum iter = IndexReaderUtils.getTermIterator(reader);
+    BytesRef bytesRef = iter.next();
+    while (bytesRef != null) {
+      // This is the current term in the dictionary.
+      String token = bytesRef.utf8ToString();
+      Term term = new Term("contents", token);
+
+      // How to dump out positional info as well, from test case at:
```
lintool: I think you can remove this part about the positions... but I think you want to add some actual `assertEquals`?

src/main/java/io/anserini/index/IndexReaderUtils.java
zeynepakkalyoncu: It works fine in Java but Pyjnius can't seem to distinguish between the two based on signature and throws an error.

lintool: Ah. 👍

lintool: To document the issue:
https://forum.image.sc/t/minimum-working-example-for-pyjnius-method-overload-issues/11329
https://github.com/kivy/pyjnius/issues/427

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -198,6 +198,15 @@ public static String analyzeTerm(String term, Analyzer analyzer) {
     return termInfo;
   }
 
+  /**
+   * Returns iterator over all terms in the collection.
+   * @param reader index reader
+   * @throws IOException if error encountered during access to index
+   */
+  public static TermsEnum getTermIterator(IndexReader reader) throws IOException {
```
lintool: I see, you're just exposing the `TermsEnum` directly... does it make sense to wrap in a custom class, like `Term`, which parallels how we wrap the postings?

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -131,6 +132,55 @@ public int getDocid() {
     }
   }
 
+  public static class IndexTerm {
```
lintool: I think just `Term` is fine?

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -198,6 +248,16 @@ public static String analyzeTerm(String term, Analyzer analyzer) {
     return termInfo;
   }
 
+  /**
+   * Returns iterator over all terms in the collection.
+   * @param reader index reader
+   * @return IndexTerm wrapper over TermsEnum
+   * @throws IOException if error encountered during access to index
+   */
+  public static IndexTerm getTermIterator(IndexReader reader) throws IOException {
```
lintool: Don't we want something like `public static Iterator<Term> getTerms(IndexReader reader)`?
The point is that `TermsEnum` is a weird Lucene construct... if we wrap into a normal Java Iterator, we can do standard Java stream operations like `map` and `forEach`...

Make sense?

src/main/java/io/anserini/index/IndexReaderUtils.java
zeynepakkalyoncu: `Term` clashes with `org.apache.lucene.index.Term`, but we could use it with the package name if we really want `Term`?

src/main/java/io/anserini/index/IndexReaderUtils.java
zeynepakkalyoncu: It does. Sorry for the back and forth, could you clarify this for me? To implement a custom iterator to do what you suggested, I would do something like below, but the `iterator` method isn't supposed to have any parameters as far as I know? How do we get around that?

```
public Iterator<IndexTerm> iterator(IndexReader reader) {
  return new IndexTermIterator(reader);
}

public class IndexTermIterator<IndexTerm> implements Iterator<IndexTerm> {
  TermsEnum curTerm;

  public IndexTermIterator(IndexReader reader) {
    this.curTerm = MultiTerms.getTerms(reader, "contents").iterator();
  }

  @Override
  public boolean hasNext() {
    return this.curTerm != null;
  }

  @Override
  public IndexTerm next() {
    return new IndexTerm(this.curTerm);
  }
}
```

lintool: Ah, `iterator()` is only necessary if you want to use it something like a `for (String term : X)`.

In this case, `Iterator<Term> getTerms(IndexReader reader)` is fine...

Also, you can use an anonymous inner class instead of defining another class...

src/main/java/io/anserini/index/IndexReaderUtils.java
lintool: sg, keep as `IndexTerm`

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -198,6 +240,52 @@ public static String analyzeTerm(String term, Analyzer analyzer) {
...
+          // Make sure iterator is positioned.
+          if (this.bytesRef == null) {
+            return true;
+          }
+
+          BytesRef originalPos = BytesRef.deepCopyOf(this.bytesRef);
+          if (this.curTerm.next() == null) {
+            return false;
+          }
+          else {
```
lintool: code formatting convention: else with closing brace above, same with catch below

lintool:
@zeynepakkalyoncu please fix final minor issues and merge.

Remember to write a more meaningful commit message - the default is just a glob of all individual commits, which isn't very helpful...

PR #922, opened 2019-12-09
Solr integration test script

x389liu:
This is a revised version of Solr integration test which matches the Elasticsearch one.

The available commands are shown below:

```
python src/main/python/run_solr_regression.py --ping
python src/main/python/run_solr_regression.py --check-index-exists core18
python src/main/python/run_solr_regression.py --delete-index core18
python src/main/python/run_solr_regression.py --create-index core18
python src/main/python/run_solr_regression.py --insert-docs core18 --input /tuna1/collections/newswire/WashingtonPost.v2/
python src/main/python/run_solr_regression.py --evaluate core18

# Run end to end
python src/main/python/run_solr_regression.py --regression core18 --input /tuna1/collections/newswire/WashingtonPost.v2/
```


src/main/python/run_solr_regression.py
```diff
@@ -0,0 +1,204 @@
...
+ch = logging.StreamHandler()
+ch.setFormatter(logging.Formatter('%(asctime)s %(levelname)s - %(message)s'))
+logger.addHandler(ch)
+logger.setLevel(logging.INFO)
+
+class SolrClient:
+    def __init__(self):
+        pass
+
+    def run_shell_command(self, command, echo=False, capture=False):
```
lintool: Can you abstract this method into a module, something like `regression_utils`? That both the Solr and ES tests use?

x389liu: That would be a good idea! Will do.

lintool:
Also, can you make sure the documentation here is up to date? https://github.com/castorini/anserini/blob/master/docs/solrini.md

Let's just leave the directory name as something like `solr-8.3.0` since we don't rename the ES directory either? To be consistent.

x389liu:
> Also, can you make sure the documentation here is up to date? https://github.com/castorini/anserini/blob/master/docs/solrini.md
> 
> Let's just leave the directory name as something like `solr-8.3.0` since we don't rename the ES directory either? To be consistent.

Sure I will change the directory name and double check commands there. Also I'm thinking of adding the instruction for running solr integration test there.

x389liu:
> Also, can you make sure the documentation here is up to date? https://github.com/castorini/anserini/blob/master/docs/solrini.md
> 
> Let's just leave the directory name as something like `solr-8.3.0` since we don't rename the ES directory either? To be consistent.

Just thought a bit more about using `solr-8.3.0` vs `solrini`. 

Currently we're running `pushd src/main/resources/solr && ./solr.sh ../../../../solrini localhost:9983 && popd` to upload the configset, and we're also using `solrini/bin/solr start -c -m 8G` to turn on solr server. Although we could update our existed scripts/commands to accommodate the name change, I feel it may be better to always rename `solr-x.x.x` to a constant name `solrini`, so that the name would be consistent overtime regardless of anserini's lucene version. 

To be consistent with elasticsearch, I would prefer to just add one step in `elastirini.md` to rename the `elasticsearch-7.5.0` to `elastirini`.

lintool:
> To be consistent with elasticsearch, I would prefer to just add one step in `elastirini.md` to rename the `elasticsearch-7.5.0` to `elastirini`.

👍

x389liu:
- Revised solr and es integration tests 
- Updated `solrini.md`

Will update `elastirini.md` as well.

x389liu:
- Updated `elastirini.md`

@lintool Could you take a look please? Thanks!

docs/elastirini.md
```diff
@@ -83,3 +90,25 @@ $ ./eval/trec_eval.9.0.4/trec_eval -c -mrecall.1000 -mmap src/main/resources/top
 map                   	all	0.1956
 recall_1000           	all	0.8573
 ```
+
+## Elasticsearch integration test
```
lintool: capital I, capital T

lintool:
Hi @x389liu - Here's the issue I'm encountering: First unpacked tarball - running:

```
python src/main/python/run_solr_regression.py --regression core18 --input /tuna1/collections/newswire/WashingtonPost.v2/
```

Works great:

```
2019-12-10 20:39:42,422 INFO - [SUCESS] 0.2495 MAP verified as expected!
```

But when I run the same command again, I get:

```
2019-12-10 20:48:11,557 INFO - [FAILED] 0.2031 MAP, expected 0.2495 MAP!
```

Seems to be something about the second trial? Perhaps there's some state that's not being cleared?

x389liu:
> Hi @x389liu - Here's the issue I'm encountering: First unpacked tarball - running:
> 
> ```
> python src/main/python/run_solr_regression.py --regression core18 --input /tuna1/collections/newswire/WashingtonPost.v2/
> ```
> 
> Works great:
> 
> ```
> 2019-12-10 20:39:42,422 INFO - [SUCESS] 0.2495 MAP verified as expected!
> ```
> 
> But when I run the same command again, I get:
> 
> ```
> 2019-12-10 20:48:11,557 INFO - [FAILED] 0.2031 MAP, expected 0.2495 MAP!
> ```
> 
> Seems to be something about the second trial? Perhaps there's some state that's not being cleared?

Sounds weird, let me try that.

lintool:
hi @x389liu - I have additional information: the second time around, if I do this:

```
python src/main/python/run_solr_regression.py --delete-index core18
pushd src/main/resources/solr && ./solr.sh ../../../../solrini localhost:9983 && popd
python src/main/python/run_solr_regression.py --create-index core18
python src/main/python/run_solr_regression.py --insert-docs core18 --input /tuna1/collections/newswire/WashingtonPost.v2/
python src/main/python/run_solr_regression.py --evaluate core18
```

Everything works fine... could it be something about re-uploading the configs?

@r-clancy thoughts?

x389liu:
@lintool the config problem should be fixed.

lintool:
@x389liu confirmed this works!

PR #923, opened 2019-12-09
Added docTTTTTquery regressions

lintool:
Added regressions for docTTTTTquery. Tweaked existing regressions and documentation to make word/terminology consistent.


src/main/resources/docgen/templates/msmarco-passage-docTTTTTquery.template
```diff
@@ -0,0 +1,74 @@
+# Anserini: Regressions for [MS MARCO Passage Retrieval](https://github.com/microsoft/MSMARCO-Passage-Ranking)
```
rodrigonogueira4: Just curious, why do we need templates?

lintool: I have an entire regression setup where a document generator takes a template, populates the expected values from the YAML, and builds the final documentation in `docs/`. This way, the YAML serves as the single source of truth.

PR #932, opened 2019-12-19
prefer complete hits instead of top scores, to fix ANN recall and latency

tteofili:
using `TopScoreDocCollector results = TopScoreDocCollector.create(indexArgs.depth, Integer.MAX_VALUE);` instead of default `indexSearcher.search(query, indexArgs.depth);` as it seems `CommonTermsQuery` doesn't properly leverage block-max WAND improvements to speed up query execution.


tteofili:
for the performance regression/bug, see also https://issues.apache.org/jira/browse/LUCENE-9107

PR #937, opened 2020-01-06
Fetch raw document

zeynepakkalyoncu:


src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -332,6 +332,18 @@ public IndexTerm next() {
...
+  /**
+   * Returns the raw document given its collection docid.
+   * @param reader index reader
+   * @param docid collection docid
+   * @return the raw document given its collection docid
+   * @throws IOException if error encountered during document fetching
+   */
+  public static String getRawDocument(IndexReader reader, String docid) throws IOException {
+    Document rawDoc = reader.document(convertDocidToLuceneDocid(reader, docid));
+    return rawDoc.get("raw");
```
lintool: A few things: I think it would be better to _not_ throw an exception - instead trap and return `null`? (and document accordingly.)
Also, check `rawDoc` for `null` before you try and call a method on it.
Finally, please use `FIELD_RAW`: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/generator/LuceneDocumentGenerator.java#L45

PR #938, opened 2020-01-08
Update repro log

JMMackenzie:
As above.


lintool:
Thanks for your contribution!

Issue #889, opened 2019-11-23
Add feature to search other fields in CORE corpus


bazingagin:
We want to index `year` and `topics` in CORE corpus so that we can also search by these fields. Basically we need to modify `LuceneDocumentGenerator.java` or create a new `generator` that outputs more than `content` field for downstream searching.

lintool:
Let's leave `LuceneDocumentGenerator` alone... probably best to create something like a `CoreDocumentGenerator` or something along those lines...

ankit-shukla-dev:
Please add me, so that I can get the code

ankit-shukla-dev:
I will clone and PR then

lintool:
To be clear, this is the CORE dataset: https://core.ac.uk/services/dataset/

edwinzhng:
Commenting so I can assign this issue to myself

edwinzhng:
Hey @bazingagin it seems like the -storeTransformedDocs flag will set the "topics" and "year" fields as a stored document in Lucene, although there is a bug right now that causes "topics" to be an empty string instead of the actual value (this also affects all other JSON fields that are arrays). I can submit a change that for the array values, but did you want me to create the new generator regardless?

lintool:
hey @edwinzhng to provide more context, our goal is to build rich search/browsing/exploration applications to the scientific literature, enabling, for example, analyses like these: https://medium.com/@nlpscholar/state-of-nlp-cbf768492f90

lintool:
Closing with updated feature request in #979

PR #942, opened 2020-01-11
Convert array fields for CORE dataset collection to string

edwinzhng:
Addresses CORE issue https://github.com/castorini/anserini/issues/889.
## Changes
The `CoreCollection` generates a `MultifieldSourceDocument` that is indexed into Lucene for the CORE dataset. Right now we perform `e.getValue().asText()` for all document fields, which for `ArrayNode` JSON types (such as the "topics" field), this [defaults to the empty string](https://fasterxml.github.io/jackson-databind/javadoc/2.10/com/fasterxml/jackson/databind/node/ContainerNode.html#asText--), not the actual text value of the field. This change will convert any array fields into a single string separated by `::`, which can then be indexed by Lucene.

To get the "topics" and "year" fields indexed as "stored" fields in Lucene, we can pass the `-storeTransformedDocs` flag to the indexer, instead of creating a new Generator class. I don't think creating a new generator would be necessary since I'm able to search on the `topic` and `year` fields successfully using the Lucene index. Just FYI I'm using https://github.com/DmitryKey/luke to verify the fields in the index.

### DIFF
--- a/src/main/java/io/anserini/collection/CoreCollection.java
+++ b/src/main/java/io/anserini/collection/CoreCollection.java
@@ -125,9 +125,9 @@ public class CoreCollection extends DocumentCollection<CoreCollection.Document> 
                   "doi:" + json.get("doi").asText();
         } else if ("abstract".equals(e.getKey())) {
           this.contents = json.get("title").asText() + "\n" + json.get("abstract").asText();
-        } else if ("topics".equals(e.getKey())) {
+        } else if (e.getValue() instanceof ArrayNode) {
           ArrayNode arrayField = (ArrayNode) e.getValue();
-          StringJoiner sj = new StringJoiner(" ");
+          StringJoiner sj = new StringJoiner("::");
           arrayField.elements().forEachRemaining( arrayElement -> {
             sj.add(arrayElement.asText());
           });
--- a/src/main/java/io/anserini/collection/CoreCollection.java
+++ b/src/main/java/io/anserini/collection/CoreCollection.java
@@ -19,6 +19,7 @@ package io.anserini.collection;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.MappingIterator;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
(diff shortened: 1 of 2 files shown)


lintool:
@bazingagin has more context than I do, so I'll let her chime in...

> This change will convert the array of topic strings into a single space-separated string, which can then be indexed by Lucene.

Instead of space sep, maybe use `::`, like Python classifiers? That way we can recover individual topics down the road...

src/main/java/io/anserini/collection/CoreCollection.java
```diff
@@ -123,6 +125,13 @@ public Document(JsonNode json) {
                   "doi:" + json.get("doi").asText();
         } else if ("abstract".equals(e.getKey())) {
           this.contents = json.get("title").asText() + "\n" + json.get("abstract").asText();
+        } else if ("topics".equals(e.getKey())) {
```
bazingagin: Follow the logic here, i think it'd be better if we could check whether a value of the field is an array or a string instead of specifying "topics" so that it will enable more array-like value in other fields to be searched.

src/main/java/io/anserini/collection/CoreCollection.java
```diff
@@ -123,6 +125,13 @@ public Document(JsonNode json) {
...
+          ArrayNode arrayField = (ArrayNode) e.getValue();
+          StringJoiner sj = new StringJoiner(" ");
```
bazingagin: I would suggest no use "space" to join the array, since "topics" consist of many phrases.

src/main/java/io/anserini/collection/CoreCollection.java
edwinzhng: Changed the conditional to be for all ArrayNode fields

src/main/java/io/anserini/collection/CoreCollection.java
edwinzhng: Updated to `::` as @lintool suggested, Lucene should ignore the punctuation when indexing so this should not be an issue right?

src/main/java/io/anserini/collection/CoreCollection.java
edwinzhng: Changed the conditional to be for all ArrayNode fields

Issue #943, opened 2020-01-12
Too long classpaths for Windows


guyrosin:
For Windows, appassembler generates invalid batch files.
Since there are many dependencies, the generated classpath is too long for Windows, and an error is shown when executing the generated scripts, such as IndexCollection.bat and SearchCollection.bat ("The input line is too long").

PR #944, opened 2020-01-12
Fixed appassembler for Windows

guyrosin:
Addressing #943:
For Windows, appassembler generated invalid batch files with too long classpaths.
To fix it, we can use wildcards (available from Java 6) along with a flat directory structure.


lintool:
@edwinzhng you're on Windows, right? can you look at this and comment?

edwinzhng:
@lintool @guyrosin I was able to reproduce the error on Windows and I can confirm that this change fixes it 💯

lintool:
Thanks @guyrosin for your contributions!

PR #946, opened 2020-01-13
Spanish Collection

Impavidity:
- Add TREC-3 Spanish test collection.

TrecTopicReader requires improvement for reading multi-lines title topic file and title-missing topic file (for example TREC-4 Spanish topic file only has description)

Results is subject to change after above fix.

Another issue is about the topic file encoding. Currently I did not change the file encoding after I download it from website. But based on the pull request page, some characters are problematic. However, if I save it with utf-8, the retrieval performance drop a lot. @lintool


x65han:
Is this PR still valid? If not, we can close?

PR #951, opened 2020-01-15
Refactoring of SearchCollection

lintool:
+ better logging in SearchCollection
+ made parameter naming more consistent
+ updated fine-tuning experiments


src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -130,12 +132,12 @@
   // the danger of (inadvertently) training on test data. These settings are used in the ATIRE system and also in
   // Lin et al. (ECIR 2016).
 
-  @Option(name = "-k1", handler = StringArrayOptionHandler.class, usage = "BM25 k1 parameter")
-  public String[] k1 = new String[] {"0.9"};
+  @Option(name = "-bm25.k1", handler = StringArrayOptionHandler.class, usage = "BM25: k1 parameter")
+  public String[] bm25_k1 = new String[] {"0.9"};
+
+  @Option(name = "-bm25.b", handler = StringArrayOptionHandler.class, usage = "BM25: b parameter")
+  public String[] bm25_b = new String[] {"0.4"};
```
Peilin-Yang: I found some of the Java parameters are snake_case (e.g. `bm25_b`) while others are camelCase (e.g. `bm25Accurate`).

Shall we normalize them?

docs/regressions-car17v1.5.md
```diff
@@ -40,19 +40,19 @@ nohup target/appassembler/bin/SearchCollection -index lucene-index.car17v1.5.pos
 
 nohup target/appassembler/bin/SearchCollection -index lucene-index.car17v1.5.pos+docvectors+rawdocs \
  -topicreader Car -topics src/main/resources/topics-and-qrels/topics.car17v1.5.benchmarkY1test.txt \
- -bm25 -axiom -rerankCutoff 20 -axiom.deterministic -output run.car17v1.5.bm25+ax.topics.car17v1.5.benchmarkY1test.txt &
+ -bm25 -rerankCutoff 20 -axiom -axiom.deterministic -output run.car17v1.5.bm25+ax.topics.car17v1.5.benchmarkY1test.txt &
```
Peilin-Yang: As a more general question, the parameter `-rerankCutoff` only applies when we specify `isRerank` to true. That means we should first see `args.rm3 || args.axiom || args.bm25prf` before seeing `-rerankCutoff`. So I guess we should put `rm3, axiom, bm25prf` before `-rerankCutoff` (the original way)?

lintool: Should we do `rankCutoff` explicitly for each model? E.g., `rm3.rerankCutoff`?

lintool: My original motivation was to keep all the "groups" together.

lintool: Or I could make the `rankCutoff` parameter to the end?

Peilin-Yang: I don't think `rerankCutoff` for each model is necessary as they will probably not present at the same time? (what is the semantic for that?) 

Putting `rerankCutoff` at the end is fine

src/main/java/io/anserini/search/SearchArgs.java
lintool: I think the convention we adopted is that underscore is used for grouping parameters together, e.g., `bm25_foo`, so `foo` are all the BM25 parameters. Otherwise, everything else is camel case?

Peilin-Yang: okay, I am fine with it -- they just look weird imo.

Peilin-Yang: Oh, I found `bm25Accurate`

PR #953, opened 2020-01-15
Explicit line endings and encoding for MSMARCO passage on Windows

edwinzhng:
Line endings + encoding errors in Python prevented the MSMARCO passage experiments from being replicated on Windows so this should fix it. I also re-ran the experiment again on Linux and it still worked.

Will wait for @LuKuuu to also confirm that it works for him on Windows


lintool:
I'll wait for +1 from @LuKuuu before I ship.

LuKuuu:
@lintool I just tested it and it works on my device.

lintool:
@LuKuuu can you add in your own replication record, per https://github.com/castorini/anserini/pull/939

Thanks!

PR #955, opened 2020-01-17
Close searcher for SearchMsmarco class

edwinzhng:
Current class does not close the searcher after a run is complete which seems to cause some sort of I/O bottleneck/speed degradation. For instance, if I run consecutive experiments (with the exact same parameters), the latency for retrieving queries will increase linearly with each new experiment.

The searcher is closed properly for the other classes so I'd assume we want the same here


lintool:
Good catch! I'll merge when Jenkins turns green.

PR #964, opened 2020-01-22
throw more descriptive error on missing fields in JSONCollection

nikhilro:


lintool:
@nikhilro thanks for your contribution. test case appreciated!

PR #969, opened 2020-01-25
add support for nested json for core collection

nikhilro:


lintool:
@nikhilro Maybe instead of:
```
doc1key1::doc1value1::doc1key2::doc1value2
```
We can use a serialization like
```
doc1 -> key1 :: doc1 -> value1 :: doc1 -> key2 :: doc1 -> value2
```
We need spaces between the keys and values so Lucene can still parse correctly, and a delimiter like `->` will be thrown away during tokenization.

nikhilro:
You mean `key1 -> value1 :: key2 -> value2`, right?

We don't add spaces for array. And when I tried to index small subset of modified CORE as it's currently implemented, it was able to tokenize without the spaces. thoughts?

lintool:
> You mean `key1 -> value1 :: key2 -> value2`, right?

Yes, sorry - You're right.
 
> We don't add spaces for array. And when I tried to index small subset of modified CORE as it's currently implemented, it was able to tokenize without the spaces. thoughts?

I think the spaces is more for easy of tokenization by downstream processes and human readability. That is, we might have, say, a UI that raws the raw string with the delimiters, and render is some human-friendly. way.

nikhilro:
Cool, done.

I'll make another PR adding spaces for arrays.

Issue #927, opened 2019-12-11
Build consistent set of Solr/ES integration test corpora


lintool:
Ref: #811 

Currently, the Solr integration test script works only on Core18. The ES integration test script works on Robust04 and MS MARCO passage. Let's bring these into alignment - make both Solr/ES work on {Core18, Robust04, and MS MARCO passage}.

lintool:
Originally planned to be part of v0.7.0 release, punting.

x389liu:
I've added the test case for Solr on robust04 and it works as expected. But there are some problems for Solr on passage and ES on core18.

1. For Solr retrieval on passage, when running
```
sh target/appassembler/bin/SearchSolr -topicreader TsvString -solr.index msmarco-passage -solr.zkUrl localhost:9983 \
-topics src/main/resources/topics-and-qrels/topics.msmarco-passage.dev-subset.txt \
-output run.solr.msmarco-passage.txt
```
I got an exception `org.apache.solr.client.solrj.impl.HttpSolrClient$RemoteSolrException: Error from server at http://10.36.40.249:8983/solr/msmarco-passage: undefined field define` which is caused by the Solr query`io.anserini.search.SearchSolr.search(SearchSolr.java:211): QueryResponse response = client.query(args.solrIndex, solrq);`.

Just want to confirm if the above command looks okay? I will need to dig into the error if it looks correct.

2. For ES indexing, there is not config file for core18. I created one similar to [this](https://github.com/castorini/anserini/blob/master/src/main/resources/elasticsearch/index-config.robust04.json) but not sure about its k1 and b value. I tried the default value `k1=0.9, b=0.4`, but it gives me `[FAILED] 0.2401 MAP, expected 0.2495 MAP`. Do you have any suggestion on this?

lintool:
@r-clancy can you help @x389liu out here?

r-clancy:
@x389liu 

For the first error, it's because Solr queries are of the form `<field>:<query_terms>` and several of the queries from passage have `define:` in them, with the colon being the culprit - Solr thinks `define` is a field (which isn't in our schema).

```
ryan@thinkpad ~/sync/git/anserini [master] $ grep "define:" src/main/resources/topics-and-qrels/topics.msmarco-passage.dev-subset.txt                                                                                       [09:45:37]
129491  define: dog's life
129517  define: entity's
129641  define: morbid obesity
129684  define: portrait
129792  define: systemic
129837  define: wrongful prosecution
999921  define: shore up
[...]

PR #972, opened 2020-01-26
Add test cases for both Solr/ES with {Core18, Robust04, and MS MARCO passage}

x389liu:
Ref: Issue #927

Note that the current AP values of Solr with MS MARCO passage and ES with Core18 do not match their expected values. All other cases should work well.

Specifically: 
- Solr with MS MARCO passage produces `[FAILED] 0.187 MAP, expected 0.1956 MAP!`
- ES with Core18 produces `[FAILED] 0.2401 MAP, expected 0.2495 MAP!`

### DIFF
--- a/src/main/java/io/anserini/search/SearchSolr.java
+++ b/src/main/java/io/anserini/search/SearchSolr.java
@@ -201,8 +201,8 @@ public final class SearchSolr implements Closeable {
     SolrQuery solrq = new SolrQuery();
     solrq.set("df", "contents");
     solrq.set("fl", "* score");
-    // Remove double quotes in query since they are special syntax in Solr query parser
-    solrq.setQuery(queryString.replace("\"", ""));
+    // Remove some characters in query which are special syntax in Solr query parser
+    solrq.setQuery(queryString.replaceAll("[+=&|<>!(){}~*?:/\"\\^\\-\\[\\]\\\\]", " "));
     solrq.setRows(args.hits);
     solrq.setSort(SortClause.desc("score"));
     solrq.addSort(SortClause.asc(FIELD_ID));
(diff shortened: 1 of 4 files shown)


lintool:
@x389liu Why can't we do this? Abstract the query generator functionality into something like `SolrBagOfWordsQueryGenerator`, which should exhibit the same behavior as `BagOfWordsQueryGenerator`. Then you can easily unit test it.

The topics are already coded here: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/topicreader/Topics.java

You can run through `BagOfWordsQueryGenerator`, print out the query, and make sure they exactly match the Solr version?

lintool:
Alternatively, could we generate the query with `BagOfWordsQueryGenerator`, dump out the query in the proper Lucene format, and then feed into `SolrQuery`?

lintool:
Take a look at this: https://lucene.apache.org/solr/guide/8_3/overview-of-searching-in-solr.html
Offers a pretty good explanation of query parsers, etc. - should be enough to help you debug.

x389liu:
After spending a while on debugging the inconsistent AP values problem, the MS MARCO case is resolved that it works well on Solr now. But Core18 still does not work as expected on Elasticsearch. So we decided to include only working test cases (i.e. Robust04 and MS MARCO on Solr) for now.

src/main/java/io/anserini/search/SearchSolr.java
```diff
@@ -198,11 +198,18 @@ public SearchSolr(Args args) throws IOException {
 
     SolrDocumentList results = null;
 
+    // Replace some characters in query which are special syntax in Solr query parser
```
lintool: Why can't this be replaced with a single regular expression?

x389liu: Thanks for the advice! Updated.

lintool:
This PR title is misnamed, but I confirm that I am able to replicate the MS MARCO results in Solr. So, for Solr, Robust04, Core18, and MS MARCO passage all work. 

Merging.

lintool:
Closed by #1030

PR #973, opened 2020-01-26
robust 4 replication log

nikhilro:


lintool:
This entry should actually be added to this file:
https://github.com/castorini/anserini/blob/master/src/main/resources/docgen/templates/robust04.template

Do a maven build, and and docs/regressions-robust04.md  should be automatically updated.

Commit edits to both template and md.

Thanks!

nikhilro:
done, thanks!

PR #977, opened 2020-01-26
Replication logs for Robost04 and Core18

edwinzhng:
I also added a comment about letting the index "catch up"

Will need to rebase with #976 before merging

### DIFF
--- a/docs/elastirini.md
+++ b/docs/elastirini.md
@@ -76,6 +76,8 @@ sh target/appassembler/bin/IndexCollection -collection JsonCollection -generator
  -es -es.index msmarco-passage -threads 9 -input /path/to/msmarco-passage -storePositions -storeDocvectors -storeRawDocs
 ```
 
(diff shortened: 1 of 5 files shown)


docs/elastirini.md
```diff
@@ -76,6 +76,8 @@ sh target/appassembler/bin/IndexCollection -collection JsonCollection -generator
  -es -es.index msmarco-passage -threads 9 -input /path/to/msmarco-passage -storePositions -storeDocvectors -storeRawDocs
 ```
 
+We may need to wait a few minutes after indexing for the index to catch up before performing retrievals, otherwise wrong evaluation metrics are returned.
```
lintool: retrievals -> retrieval

edwinzhng: Updated, also lumped in some other replication logs in here too

lintool:
Replicate notes here should be added to template: https://github.com/castorini/anserini/blob/master/src/main/resources/docgen/templates/core18.template

Maven build auto-generates .md file. See #974

Issue #979, opened 2020-01-28
Implement custom CoreGenerator to generate custom Lucene document


lintool:
In the current implementation, `CoreCollection` generates a `MultifieldSourceDocument`, and we use the default `LuceneDocumentGenerator` to create a Lucene document. The issue is that `LuceneDocumentGenerator` doesn't differentiate the fields, which we need in this case.

For example, years we probably want to explicitly index as a numeric value so we can do range queries. Journal names and author names we don't want to stem. Etc.

The solution is that we need a custom generator, along the lines of the https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/generator/WapoGenerator.java

The logic of parsing out individual CORE fields in JSON, e.g., https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/collection/CoreCollection.java#L123 should be directly moved into the generator.

lintool:
Addressed by #983

PR #983, opened 2020-02-01
Implement CoreGenerator to customize fields for CORE index

edwinzhng:
Addresses #979

Added a new CoreGenerator class to index fields with different parameters.

Currently:
* "year" is indexed as a numeric to support range queries
* "authors", "doi", "journals", etc (basically any fields with URLs or names) are indexed as StringField to avoid stemming but this also means the text doesn't get tokenized if there are multiple authors or journals etc - not sure if this is the behavior we want? @lintool 
* I also added braces around the nested JSON field strings because occasionally there are multiple layers of nesting and we need to be able to parse them back
  - eg) `{ references -> [] } :: { documentType -> { type -> } :: { confidence -> } }`
* Changed null fields to be the empty string instead of indexing the word "null"

Here is a sample document with the indexed fields:

image


src/test/java/io/anserini/collection/CoreDocumentTest.java
```diff
@@ -46,20 +54,16 @@ public void setUp() throws Exception {
       "  \"abstract\": \"this is the abstract 1\"," +
       "  \"topics\": [\"Topic 1\", \"Other\"]," +
       "  \"year\": 2020," +
-      "  \"field1\": \"doc1 field1 content\"," +
-      "  \"field2\": \"doc1 field2 content\"," +
-      "  \"nested_field\": {\"doc1key1\": \"doc1value1\", \"doc1key2\": \"doc1value2\"}" +
+      "  \"nested_field\": {\"doc1key1\": \"doc1value1\"}" +
       "}\n" +
```
bazingagin: it would be better if we could add test cases for deeper and broader nested fields like `"enrichments": {"references": [], "documentType": {"type": null, "confidence": null}}`

edwinzhng: Oops you're right I forgot to move some of the tests over, will add a new CoreGeneratorTest for JSON node parsing

pom.xml
```diff
@@ -387,5 +387,10 @@
       <artifactId>wdtk-dumpfiles</artifactId>
       <version>0.10.0</version>
     </dependency>
+    <dependency>
```
edwinzhng: Added this for creating mocks in the unit tests, hopefully this should be useful for later development - I'm surprised we didn't use mocks anywhere already

lintool:
The way to think about different field type is as follows: `StringField` are for lookup based on the "entire thing". That's why we index `docid` as a `StringField`. So DOI should be like this.

Authors, journals, etc. should still be tokenized, but _not_ stemmed. We gotta manually set the field type, like here: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/generator/LuceneDocumentGenerator.java#L134

edwinzhng:
> The way to think about different field type is as follows: `StringField` are for lookup based on the "entire thing". That's why we index `docid` as a `StringField`. So DOI should be like this.
> 
> Authors, journals, etc. should still be tokenized, but _not_ stemmed. We gotta manually set the field type, like here: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/generator/LuceneDocumentGenerator.java#L134

Added a new `NonStemmingAnalyzer` class to do this, hopefully reuseable later

pom.xml
lintool: yes, we've not been particularly rigorous in testing... this is my fault in not cracking the whip hard enough :)

edwinzhng:
> It is really great to have `CoreGeneratorTest`! I think adding mock is a good idea for testing. my only small concern is that structure-wise, i'm not sure if it's really necessary to have a `NonStemmingAnalyzer` as a separate class

Yeah wasn't sure about separating it. I was thinking fields that require tokenization but not stemming would show up later so this could be used if we ever build other generators, but I can move it inside the CoreGenerator if that's better?

lintool:
One final comment from me: I think it's fine to have `NonStemmingAnalyzer.java`, but: Rename it `NonStemmingEnglishAnalyzer` but have it match the behavior of the [`EnglishAnalyzer`](https://github.com/apache/lucene-solr/blob/master/lucene/analysis/common/src/java/org/apache/lucene/analysis/en/EnglishAnalyzer.java), which is Anserini's default.

This will cause less confusion because stemming is the minimal pair in terms of differences.

edwinzhng:
Sorry, just realized the EnglishStemmingAnalyzer won't do stemming if you don't specify the stemmer - I removed the NonStemmingAnalyzer

PR #985, opened 2020-02-10
Remove datePublished from CoreGenerator to prevent errors on Solr

edwinzhng:
When we index the CORE hydrology abstracts, Solr reads many of the fields as `Long` values and assigns the the `Long` type to the `datePublished` field. Occasionally there is a `String` that cannot be parsed, breaking the whole indexing pipeline.

Thought it'd be easier to just remove this field entirely since it gives the same value as "year", but occasionally contains month/day. The alternative is to hard-code `datePublished` in the Solr schema.xml file which I'm not a fan of

### DIFF
--- a/src/main/java/io/anserini/index/generator/CoreGenerator.java
+++ b/src/main/java/io/anserini/index/generator/CoreGenerator.java
@@ -84,7 +84,6 @@ public class CoreGenerator extends LuceneDocumentGenerator<CoreCollection.Docume
     CoreField.DOI.name,
     CoreField.OAI.name,
     CoreField.IDENTIFIERS.name,
-    CoreField.DATE_PUBLISHED.name,
     CoreField.DOWNLOAD_URL.name,
     CoreField.RELATIONS.name,
     CoreField.FULL_TEXT_IDENTIFIER.name));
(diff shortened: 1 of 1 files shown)


lintool:
What is the the format of the field usually? What if we converted everything to epoch seconds prior to indexing?

That way we don't lose information prior to indexing?

edwinzhng:
> What is the the format of the field usually? What if we converted everything to epoch seconds prior to indexing?
> 
> That way we don't lose information prior to indexing?

Sure I could do that but the data isn't very consistent, sometimes it has "YYYY-MM-dd", "YYYY", and occasionally "Jan 12th 2018" etc. Not sure if it's worth trying to parse everything

lintool:
Okay, I agree.

> sometimes it has "YYYY-MM-dd", "YYYY", and occasionally "Jan 12th 2018" etc.

Can you add the fact that the format is inconsistent as a reason for dropping. I.e., not worthwhile to try and normalize all variants. Then we'll merge.

arjenpdevries:
(keep it as String?)

lintool:
@arjenpdevries probably not worth it... :)

Issue #986, opened 2020-02-11
Ingest ACL Anthology into Anserini and then Solrini


lintool:
ACL Anthology: https://www.aclweb.org/anthology/

Once we implement indexer of bibtex files ref:  #984 this should be very easy.

Anthology data, in raw XML: https://github.com/acl-org/acl-anthology/tree/master/data/xml
Conversion script to bibtext: https://github.com/acl-org/acl-anthology/blob/master/bin/create_bibtex.py

lintool:
The code for this is done for #994.

We should write up documentation for doing this, since ACL anthology is publicly available. Probably a single page, something like "Ingesting the ACL Anthology into Anserini and Solrini".

The goal when I write documentation is that someone should be able to replicate the results just by copy and pasting the commands.

edwinzhng:
> The code for this is done for #994.
> 
> We should write up documentation for doing this, since ACL anthology is publicly available. Probably a single page, something like "Ingesting the ACL Anthology into Anserini and Solrini".
> 
> The goal when I write documentation is that someone should be able to replicate the results just by copy and pasting the commands.

Will add in the next PR

edwinzhng:
Closing, addressed by #1005

PR #994, opened 2020-02-23
BibtexGenerator and BibtexCollection 

nikhilro:
**Delivers**
* BibtexGenerator and BibtexCollection classes to index `acl-anthology` dataset addressing issue #986

**Open Questions**
* It might be worth combining some Core and Bibtex duplicate code right now or waiting to see how the needs evolve. Examining the steps happening in their Generator and Collection classes, it should be possible to make a more generalized class @lintool @edwinzhng


lintool:
@edwinzhng can you do a PR?

nikhilro:
@lintool wait what PR?

lintool:
@nikhilro sorry I meant CR.

@edwinzhng one big PR or smaller PRs, I don't really care.

src/main/java/io/anserini/collection/BibtexCollection.java
```diff
@@ -0,0 +1,137 @@
...
+    private String contents;
+    private String type;
+    private BibTeXEntry bibtexEntry;
+
+    public Document(Map.Entry<Key, BibTeXEntry> entry) {
+      id = entry.getKey().toString();
+      Map<Key, Value> bibtexFields = entry.getValue().getFields();
+      String doctitle = bibtexFields.getOrDefault(new Key("title"), new StringValue("", StringValue.Style.QUOTED)).toUserString();
+      String docAbstract = bibtexFields.getOrDefault(new Key("abstract"), new StringValue("", StringValue.Style.QUOTED)).toUserString();
+      contents = doctitle + ". " + docAbstract;
```
edwinzhng: Don't need the period since it'll just get indexed away

nikhilro: I added it because when you query. Isn't it useful to be able to read the `contents` fields. If it was just space, how do you tell the title and the abstract from the contents field? Are you saying that it's a field not meant for human consumption but for indexing aid?

nikhilro: resolving per discussion: contents does not need to be readable. remove period in next PR

PR #1002, opened 2020-02-27
More refactoring of collection test cases

lintool:


src/test/java/io/anserini/collection/JsonCollectionTest.java
```diff
@@ -18,7 +18,7 @@
 
 import java.util.Map;
 
-public class JsonCollectionTest extends CollectionTest<JsonCollection.Document> {
+public class JsonCollectionTest extends DocumentCollectionTest<JsonCollection.Document> {
```
edwinzhng: Just to be explicit, should this class be abstract too since the other JSON test classes inherit from this?

lintool: 👍

PR #1003, opened 2020-02-27
Another round of collection test refactoring

lintool:


lintool:
@edwinzhng @nikhilro this is ready for CR, btw. Thanks!

lintool:
Yes, I use IntelliJ and PyCharm - I just ran the default formatter to get the spacing, etc. right.

edwinzhng:
> Yes, I use IntelliJ and PyCharm - I just ran the default formatter to get the spacing, etc. right.

Might be nice to add a Git hook to automatically format things like https://github.com/Cosium/maven-git-code-format

lintool:
Definitely worth thinking about: file an issue?

There's spectrum between prod best practices and what we should be doing in research. If we held ourselves to prod industry standards, our research velocity would decrease... so there's a balance somewhere... but I'd like to know about the right tools around, so thanks for bringing up.

PR #1005, opened 2020-02-29
Add ACL generator and Solr documentation

edwinzhng:
1. Added the `AclGenerator` and `AclDocument` classes along with associated test cases modeled after the new refactor.

2. Added documentation for setting up the ACL anthology Solr index from scratch, and for modifying Solr schemas. This includes commands to modify the Solr `acl` index schema.

3. Cleaned up some unused imports in a few tests and refactored the `BibtexDocumenTest`.

Example Solr ACL document:
image


src/main/java/io/anserini/collection/AclCollection.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+
+/**
+ * A YAML document collection for the ACL anthology.
+ */
+public class AclCollection extends DocumentCollection<AclCollection.Document> {
```
lintool: Why don't we just name `AclAnthology`?

src/main/java/io/anserini/collection/AclCollection.java
```diff
@@ -0,0 +1,201 @@
...
+  private JsonNode volumes;
+  private static final Logger LOG = LogManager.getLogger(AclCollection.class);
+
+  public AclCollection(Path path) {
+    this.path = Paths.get(path.toString(), "/papers"); // Path containing files to iterate
+    this.allowedFileSuffix = new HashSet<>(Arrays.asList(".yaml"));
```
lintool: `Set.of(...)`

docs/acl.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Deploying Solr
+
+Set up Solr locally by following the instructions for setting up a single-node SolrCloud instance in [solrini.md](solrini.md).
+
+## Generating ACL Anthology Data
+
+First, clone the ACL anthology repository containing the raw XML data:
+
+  ```git clone git@github.com:acl-org/acl-anthology.git```
```
lintool: Let's do "```" on it's own line, followed by `\n`, consistently?

docs/acl.md
```diff
@@ -0,0 +1,72 @@
...
+ 
+Next, navigate to the `acl-anthology` folder and install dependencies (with Python 3.7+):
+
+```pip install -r requirements.txt```
```
lintool: Same and below.

docs/acl.md
```diff
@@ -0,0 +1,72 @@
...
+
+First, let's create the index collection in Solr using the default config.
+
+```solrini/bin/solr create -n anserini -c acl```
+
+Modify the `acl` Solr schema to allow multiValued fields for facets and set explicit field types:
+
+```sh
+curl -X POST -H 'Content-type:application/json' --data-binary '{
+  "add-field":{
```
lintool: Make sense to put in a file in `src/resources`?

src/main/java/io/anserini/collection/AclCollection.java
```diff
@@ -0,0 +1,201 @@
...
+        });
+      }
+
+      // Retrieve parent volume metadata
+      String parentVolumeId = paper.get("parent_volume_id").asText();
+      volume = AclCollection.this.volumes.get(parentVolumeId);
+
+      // Process venue facets
+      venues = new ArrayList<>();
+      ArrayNode venuesNode = (ArrayNode) volume.get("venues");
```
lintool: If you're gonna do functional style, might as well go all the way and use `.collect(Collectors.toList());` or something similar? Same below.

src/main/java/io/anserini/collection/AclCollection.java
```diff
@@ -0,0 +1,201 @@
...
+      }
+      return contents;
+    }
+
+    @Override
+    public boolean indexable() {
+      return true;
+    }
+
+    public JsonNode paper() {
```
lintool: I'm a bit torn. Your method naming is cleaner IMO, but Java style guidelines actually dictate `getX()`... thoughts?

src/main/java/io/anserini/index/generator/AclGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+
+import java.io.StringReader;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Converts a {@link AclCollection.Document} into a Lucene {@link Document}, ready to be indexed.
+ */
+public class AclGenerator extends LuceneDocumentGenerator<AclCollection.Document> {
```
lintool: `AclAnthologyGenerator`?

src/main/java/io/anserini/index/generator/AclGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+  public static final String FIELD_ID = "id";
+  public static final String FIELD_BODY = "contents";
+  public static final String FIELD_RAW = "raw";
+
+  public static enum AclField {
```
lintool: `AclAnthologyField`?

src/main/java/io/anserini/index/generator/AclGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+    PAGE_LAST("page_last");
+
+    public final String name;
+
+    AclField(String s) {
+      name = s;
+    }
+  }
+
+  public static final List<String> STRING_FIELD_NAMES = new ArrayList<>(Arrays.asList(
```
lintool: `List.of(...)`

src/main/java/io/anserini/index/generator/AclGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+
+        nonStemmingAnalyzer.close();
+      } else {
+        // default to normal Field with tokenization and stemming
+        doc.add(new Field(key, fieldString, storedFieldType));
+      }
+    });
+
+    // index authors
+    aclDoc.authors().forEach(author -> {
```
lintool: nm, I like `authors()` etc. better.
But here I don't think you need braces?

src/test/java/io/anserini/collection/AclDocumentTest.java
```diff
@@ -0,0 +1,132 @@
...
+package io.anserini.collection;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.Before;
+
+public class AclDocumentTest extends DocumentCollectionTest<AclCollection.Document> {
```
lintool: Rename accordingly.

lintool:
Left a bunch of comments...

src/main/java/io/anserini/collection/AclCollection.java
edwinzhng: Yeah I thought it'd be more consistent since we already have the `id()` and `content()` functions instead of `getId()` and `getContent() baked into the `Document` class, if we changed it to Java guidelines we'd probably want to change those as well

docs/acl.md
edwinzhng: I had that in my first commit, but then I realized it's just a single command they could copy paste so I pulled it into the README. I'll move it back since it's probably better to make a new file in case things need to get added in the future?

src/main/java/io/anserini/collection/AclCollection.java
edwinzhng: `AclAnthologyCollection` or just `AclAnthology` alone? I like `AclAnthologyCollection` for consistency

src/main/java/io/anserini/collection/AclCollection.java
edwinzhng: I think I'd need to convert the JsonNode iterator to a `Stream` first, which might be harder to read:

edwinzhng: I removed the curly braces and put everyone on one line, let me know if that's fine

src/main/java/io/anserini/collection/AclCollection.java
lintool: Anthology is already a collection: https://www.merriam-webster.com/dictionary/anthology
So `AclCollectionCollection` sounds odd?

docs/acl.md
lintool: I agree.

src/main/java/io/anserini/collection/AclCollection.java
lintool: Yea, original looks better. Wish Java had better stream/iterator interop.

src/main/java/io/anserini/collection/AclCollection.java
lintool: Yup, let's leave as is.

src/main/java/io/anserini/collection/AclCollection.java
edwinzhng: Ah right, just renamed it. I guess I forgot all of my GRE vocabulary

lintool:
@edwinzhng Go ahead and merge!

src/main/java/io/anserini/collection/AclCollection.java
```diff
@@ -0,0 +1,201 @@
...
+
+    public List<String> sigs() {
+      return sigs;
+    }
+
+    private String getOrDefault(JsonNode json, String key) {
+      if (json.has(key)) {
+        return json.get(key).asText();
+      }
+      return "";
```
nikhilro: return `key`?

edwinzhng: Doesn't make sense because if the key,value pair is not the map, why would you return the key itself?

PR #1007, opened 2020-03-02
Improve CORE generator for facets and Solr indexing

edwinzhng:
* Added a script to configure the Solr schema for a CORE collection
* I also changed how we index array fields so that instead of a string separated by `::`, we just index a multi-valued field, which allows faceting by individual values similar to what was done in #1005 and also preserves the original field values
* Refactored the CoreCollection tests

Example below for `gwf` documents:
image


src/main/java/io/anserini/collection/CoreCollection.java
```diff
@@ -52,7 +50,7 @@
 
   public CoreCollection(Path path){
     this.path = path;
-    this.allowedFileSuffix = new HashSet<>(Arrays.asList(".json.xz"));
+    this.allowedFileSuffix = new HashSet<>(Arrays.asList(".json.xz", "json"));
```
lintool: `Set.of(...)`

src/main/java/io/anserini/collection/CoreCollection.java
```diff
@@ -112,21 +117,13 @@ public void readNext() throws NoSuchElementException {
...
-        if ("coreId".equals(e.getKey())) {
-          this.id = (json.get("doi").asText().equals("null")) ? json.get("coreId").asText() :
-                  "doi:" + json.get("doi").asText();
-        } else if ("abstract".equals(e.getKey())) {
-          this.contents = json.get("title").asText() + "\n" + json.get("abstract").asText();
-        } else {
-          this.jsonFields.put(e.getKey(), e.getValue());
-        }
-      });
+      id = (json.get("doi").asText().equals("null")) ?
```
lintool: potential NPE here if document doesn't have `doi`?

src/main/java/io/anserini/collection/CoreCollection.java
```diff
@@ -144,8 +141,15 @@ public boolean indexable() {
...
 
-    public Map<String, JsonNode> jsonFields() {
-      return jsonFields;
+    public JsonNode jsonNode() {
+      return jsonNode;
+    }
+
+    public String getJsonString(JsonNode json, String key) {
+      if (!json.has(key) || json.get(key).asText() == "null") {
+        return "";
```
lintool: Is returning `null` better than empty string here?

src/main/resources/solr/setup/core-collection.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env sh
+
+if [ $# -eq 0 ]; then
+  echo "usage: sh core-collection.sh [COLLECTION_NAME]"
+  exit 0
+fi
+
+# Set specific field types and multiValued field properties in Solr schema
+curl -X POST -H 'Content-type:application/json' --data-binary '{
```
lintool: One suggestion might be to put the binary data in a file, and the rest of the curl command directly inline in the docs. The advantage of this is that a script is completely opaque, where you can see, oh, it's a curl command with a bunch of json data... I'll look in the file if I really want to learn about all the data...

Same with ACL anthology docs.

src/main/resources/solr/setup/core-collection.sh
```diff
@@ -0,0 +1,64 @@
...
+    "stored":true,
+    "multiValued":true,
+    "docValues":true
+  },
+  "add-field":{
+    "name":"datePublished",
+    "type":"string",
+    "stored":true
+  }
+}' http://localhost:8983/solr/$1/schema
```
nikhilro: bash variable use should be quoted: `"$1"`  [first link I found on google](https://stackoverflow.com/a/700233)

nikhilro:
Also, I spent some time yesterday and will spend more time today. I wasn't able to the `acl-collection.sh` script because I got the error `fields` already exist even after dropping the collection, creating it again, using `delete-field` & `delete-copy-field` instead of `add-field`... Once I get it work, can probably add a script to "clean up" the solr index for a collection as well

edwinzhng:
> Also, I spent some time yesterday and will spend more time today. I wasn't able to the `acl-collection.sh` script because I got the error `fields` already exist even after dropping the collection, creating it again, using `delete-field` & `delete-copy-field` instead of `add-field`... Once I get it work, can probably add a script to "clean up" the solr index for a collection as well

"replace-field" works pretty well

lintool:
I'm fine for this to be merged once @nikhilro signs off. I don't need to be a blocker.

nikhilro:
> > Also, I spent some time yesterday and will spend more time today. I wasn't able to the `acl-collection.sh` script because I got the error `fields` already exist even after dropping the collection, creating it again, using `delete-field` & `delete-copy-field` instead of `add-field`... Once I get it work, can probably add a script to "clean up" the solr index for a collection as well
> 
> "replace-field" works pretty well

oh, my bad, the error was `"multiValued":trues` (correct: `"multiValued":true`) and solr silently ate the error on running the script. It's fixed in this PR so works now lol

let's add a comment in `docs/acl.md` regarding replacing with `replace-field` if error "field already exists" is returned?

src/main/java/io/anserini/collection/CoreCollection.java
edwinzhng: Left it as empty string since I'd rather not index "null"

src/main/resources/solr/setup/core-collection.sh
edwinzhng: Will clean this up in separate PR

edwinzhng:
> > > Also, I spent some time yesterday and will spend more time today. I wasn't able to the `acl-collection.sh` script because I got the error `fields` already exist even after dropping the collection, creating it again, using `delete-field` & `delete-copy-field` instead of `add-field`... Once I get it work, can probably add a script to "clean up" the solr index for a collection as well
> > 
> > 
> > "replace-field" works pretty well
> 
> oh, my bad, the error was `"multiValued":trues` (correct: `"multiValued":true`) and solr silently ate the error on running the script. It's fixed in this PR so works now lol
> 
> let's add a comment in `docs/acl.md` regarding replacing with `replace-field` if error "field already exists" is returned?

We can add different options to the script later like a flag for replace `--replace` which should be good 👍

nikhilro:
add a `TODO` in the script to keep track of the `--rename` option then

PR #1008, opened 2020-03-02
All Lucene Document field name should be in a single location

lintool:
Previously, Lucene Document field names were defined in `LuceneDocumentGenerator`, such as 
`LuceneDocumentGenerator.FIELD_BODY`. Other generators defined their own field names as static strings. This is a hidden dependency. The field names should be exactly the same across all generator types, and all the static strings should draw from the same source.

This PR fixes this issue. All field names are now defined in `IndexArgs`.

Also, renamed `WapoGenerator` to `WashingtonPostGenerator`.


lintool:
we need an emoticon for "go to market", I suppose, that parallels 🚢 it.

I'll make above changes before merging.

PR #1015, opened 2020-03-06
Add baseline for TREC news BL

### LINKED ISSUE #1012: Create a baseline for TREC background linking with Anserini
@Chriskamphuis will work on this - from what I understand, the baselines can be generated directly by issuing the right commands - we need to document what these are.

I think this is just a markdown guide in `doc/` - might also want to check in the topics directly.

Chriskamphuis:
fix #1012
- Add baselines for both 2018 and 2019
- Add qrels + topics for TREC BL 18+19
- Add link in README.md to the markdown that explains how to generate the baselines.


Chriskamphuis:
Note that the topics of last year are not on the TREC website yet. The link needs to be added after they are published there.

PR #1017, opened 2020-03-06
Analyzer as optional argument in SimpleSearcher

### LINKED ISSUE #1013: Add option to change Analyzers in SimpleSearcher
This can be implemented similarly as setting and unsetting RM3.
This will allow for different analyzers to be used when using the SimpleSearcher through Pyserini, for example when no tokenization is desired.

Chriskamphuis:
fix #1013
Analyzer is an optional argument in the SimpleSearcher constructor. Also added getter and setter for the Analyzer.
Add a TokenizeAnalyzer class to Anserini and a test for this Analyzer.


src/main/java/io/anserini/analysis/TokenizeAnalyzer.java
```diff
@@ -0,0 +1,13 @@
+package io.anserini.analysis;
+
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.analysis.standard.StandardTokenizer;
+
+public class TokenizeAnalyzer extends Analyzer {
```
lintool: Would `TokenizeOnlyAnalyzer` be a better name?

Chriskamphuis: Yes I agree, will update it.

PR #1023, opened 2020-03-06
Added methods to get postings lists based on analyzed and unanalyzed forms in IndexReaderUtils

### LINKED ISSUE #990: getPostingsList method in IndexReaderUtils should take analyzed form also
Currently, the definition of `getPostingsList` in `IndexReaderUtils`:

```
public static List<Posting> getPostingsList(IndexReader reader, String termStr)
```

Takes the raw (unanalyzed) form; we should probably implement a variant that takes the analyzed form as well.

lintool:
Closes #990 

The primary change is that `IndexReaderUtils` now has methods to fetch postings lists given analyzed and unanalyzed forms. However, this necessitated a bunch of changes for proper testing, etc.

### DIFF
--- a/src/main/java/io/anserini/index/IndexReaderUtils.java
+++ b/src/main/java/io/anserini/index/IndexReaderUtils.java
@@ -314,6 +314,20 @@ public class IndexReaderUtils {
     }
   }
 
+  // These are bindings for Pyserini to work, because Pyjnius can't seem to distinguish overloaded methods.
+  // jnius.JavaException: No methods matching your arguments
+  public static List<Posting> getPostingsListForUnanalyzedTerm(IndexReader reader, String term) {
+    return getPostingsList(reader, term, true);
+  }
+
+  public static List<Posting> getPostingsListForAnalyzedTerm(IndexReader reader, String term) {
+    return getPostingsList(reader, term, false);
+  }
+
+  public static List<Posting> getPostingsListWithAnalyzer(IndexReader reader, String term, Analyzer analyzer) {
+    return getPostingsList(reader, term, analyzer);
+  }
+
   /**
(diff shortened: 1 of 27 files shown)


src/test/java/io/anserini/index/IndexReaderUtilsTest.java
```diff
@@ -88,50 +77,48 @@ public void testIterateThroughTerms() throws Exception {
...
 
     assertEquals(false, iter.hasNext());
   }
 
   @Test
-  public void testPostingsLists() throws Exception {
+  public void testPostingsNonExistings() throws Exception {
+    Directory dir = FSDirectory.open(tempDir1);
+    IndexReader reader = DirectoryReader.open(dir);
+    assertEquals(null, IndexReaderUtils.getPostingsList(reader, "asxe"));
```
edwinzhng: assertNull for consistency?

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -280,21 +261,71 @@ public IndexTerm next() {
...
+   * @param term term
+   * @param analyzer analyzer
+   * @return the postings list for an unanalyzed term
+   */
+  public static List<Posting> getPostingsList(IndexReader reader, String term, Analyzer analyzer) {
+    return _getPostingsList(reader, AnalyzerUtils.analyze(analyzer, term).get(0));
+  }
+
+  // Internal helper: takes the analyzed form in all cases.
+  private static List<Posting> _getPostingsList(IndexReader reader, String raw) {
```
edwinzhng: `analyzedTerm` as the parameter name instead of `raw`?

lintool: Done!

PR #1027, opened 2020-03-07
Analyzer refactoring: single analyzer to handle stem/no stem and stopwords

lintool:
@Chriskamphuis instead of a separate class for no-stemming, I think this is much cleaner.

@edwinzhng I think this is a much clear way to construct non-stemming analyzers, rather than the previous way.


lintool:
@Chriskamphuis yup - once this is merged we have to make changes to Pyserini again...

Issue #1001, opened 2020-02-26
ES with Core18 gives incorrect MAP value


x389liu:
Ref: #927 

Elasticsearch with Core18 gives MAP value 0.2401 instead of the expected value 0.2495.

lintool:
Closed by #1030

PR #1030, opened 2020-03-09
Add core18 collection to Elasticsearch

x389liu:
Ref: #1001 #972 #927


x389liu:
Core18 is now working fine with Elasticsearch. I found the inconsistent MAP problem was caused by a silly mistake (I used `JsoupGenerator` instead of `WashingtonPostGenerator` for indexing...)

PR #1031, opened 2020-03-09
Optional analyzer in getTermCounts

Chriskamphuis:
Make it possible to define an analyzer in the IndexReaderUtils.getTermCounts function.

### DIFF
--- a/src/main/java/io/anserini/index/IndexReaderUtils.java
+++ b/src/main/java/io/anserini/index/IndexReaderUtils.java
@@ -17,6 +17,7 @@
 package io.anserini.index;
 
 import io.anserini.analysis.AnalyzerUtils;
+import io.anserini.analysis.DefaultEnglishAnalyzer;
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
 import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
 import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream;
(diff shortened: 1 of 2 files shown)


src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -205,13 +205,17 @@ public static IndexReader getReader(String path) throws IOException {
 
   public static Map<String, Long> getTermCounts(IndexReader reader, String termStr) throws IOException, ParseException {
     EnglishAnalyzer ea = new EnglishAnalyzer(CharArraySet.EMPTY_SET);
```
lintool: hey, while you're at it, can you refactor to our new way of creating analyzer instances?

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -205,13 +205,17 @@ public static IndexReader getReader(String path) throws IOException {
...
-    QueryParser qp = new QueryParser(IndexArgs.CONTENTS, ea);
+    return getTermCounts(reader, termStr, ea);
+  }
+
+  public static Map<String, Long> getTermCounts(IndexReader reader, String termStr, Analyzer analyzer) throws IOException, ParseException {
```
lintool: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/IndexReaderUtils.java#L316

I don't think this will work w/ pyjnius? Just rename method `getTermCountsWithAnalyzer`

lintool:
Also, can you please add test case?

PR #1032, opened 2020-03-09
Segment constructors should be public

lintool:
@edwinzhng @nikhilro I'm going to sneak in this PR before the the next artifact release.

I've made all segment constructors public; there's not reason they shouldn't be (since the segment implementations are static classes, and thus can have existence outside ties to a collection), and in fact some of them are already public. This makes all segment constructors uniformly public.

This also increases testability as a side effect.


lintool:
I'll circle around and add segment specific test cases later.

PR #1033, opened 2020-03-10
Reading multi-line topic titles from TREC Topic files

alxliu:
Hi @lintool, I'm currently taking Prof. Cormack's IR course.

I was looking into the regressions for Disks 1 & 2, and realized that the current TREC topic reader does not read the topic title if it spans multiple lines (see the test case I updated).

The effectiveness results will probably have to be updated for all TREC regressions, but I'm just putting up this PR for the fix in the code for the time being.


lintool:
@alxliu okay thanks - we're in the middle of a release process, so I'm going to hold off merging this until we're done. sg?

alxliu:
Sounds good! No rush.

lintool:
@alxliu any further updates on this?

alxliu:
@lintool sorry, been busy with final projects. I can finish this up after the term is over.

lintool:
@alxliu nudge on this?

eelstretching (commented):
Looks like this change solves the issue with titles that take more than one line. There are a number of such titles in the TREC adhoc topics. Seems like the test should be expanded to include more of these (e.g., 139,  142) as well as a few more of the single-liners.

Since there were no other dependencies, I moved this class into the mainline code and re-ran an index of TREC disks 1 and 2 and then re-ran the adhoc queries, which resulted in changes in the scores for the baselines.

lintool:
hi @eelstretching - thanks for looking into this. Would you mind sending a PR w/ updated replication numbers and improved test cases?

eelstretching:
I'm happy to do it. I presume you would like me to do it on the main branch, rather than over here?

lintool:
Yes please!

lintool:
Closed by #1482

Issue #1009, opened 2020-03-03
Refactor Solr schema scripts for ACL and CORE collections


edwinzhng:
Currently, JSON data for our Solr schemas are put directly into the setup scripts. There are two ways we can refactor this:

1. Put JSON data in a separate file and have different schema files for adding or replacing fields, and allow our bash setup script to take the JSON data as a command line argument. In other words, we have a single setup script but multiple schema data files

2. Keep JSON data in the bash setup scripts as-is, but support different flags such as "--replace" to replace existing fields or "--add" to add new fields. Here, we have multiple setup scripts for each specific collection (ACL, CORE).

We will also need to update documentation after this is done.

Option 1 seems cleaner to me but would like to hear other opinions @nikhilro @lintool

nikhilro:
1. sounds good

lintool:
Issue resolved in #1069 and and #1068.

PR #1035, opened 2020-03-11
Refactoring solr scripts to make config data more explicit

nikhilro:
Addresses #1009 

This PR is a WIP, opening to show different possibilities of handling configs for our Solr collections

Open Question:
* Should we put the entire schema (a) or just modified fields on top of the base `anserini` config (b)?
  * a) [acl-anthology](https://github.com/castorini/anserini/blob/a8c2b52743fc0579c3be44e41acbb2d7ce73407b/src/main/resources/solr/schemas/acl-anthology.json)
   * b) [core-collection](https://github.com/castorini/anserini/blob/a8c2b52743fc0579c3be44e41acbb2d7ce73407b/src/main/resources/solr/schemas/core-collection.json)


lintool:
Just a note that I'm in the middle of cutting a release, so a moratorium on merges until release is complete...

nikhilro:
yea, sorry, don't intend to merge. just wanna show

src/main/resources/solr/schemas/acl-anthology.json
```diff
@@ -0,0 +1,242 @@
+{
```
edwinzhng: Don't think we should include Solr internal fields like `__root__` etc since those are probably not going to change?

src/main/resources/solr/setup/core-collection.sh
```diff
@@ -6,59 +6,4 @@ if [ $# -eq 0 ]; then
...
-    "multiValued":true,
-    "docValues":true
-  },
-  "add-field":{
-    "name":"datePublished",
-    "type":"string",
-    "stored":true
-  }
-}' http://localhost:8983/solr/"$1"/schema
+curl -X POST -H 'Content-type:application/json' --data-binary '@src/main/resources/solr/schemas/core-collection.json' http://localhost:8983/solr/"$1"/schema
```
edwinzhng: Can probably make this script take some more parameters (schema file name, Solr collection name which already exists) to make this generic

src/main/resources/solr/schemas/acl-anthology.json
nikhilro: I agree. What are the internal Solr internal fields? Is it everything with underscores? I'll create a collection with no configuration. Let's see what fields exist

src/main/resources/solr/setup/core-collection.sh
nikhilro: Yea, I was gonna but thought we don't need more generic right now (`yagni` lol). I'll probably add, feels weird to not have generic

src/main/resources/solr/schemas/acl-anthology.json
edwinzhng: Yeah it's just the underlined fields, everything else is from the fields that are actually part of the indexed documents

lintool:
See #1068

> Should we put the entire schema (a) or just modified fields on top of the base anserini config (b)?

I like option (b) better, which is what the instructions in the above PR say. That way, for collections that don't need to do anything "special" wrt schemas, we can save a step.

lintool:
I think this is fine to merge for now. I can take over and work on it more...

edwinzhng:
Closing since this was fixed in #1069

Issue #1034, opened 2020-03-11
Add Resources + Regressions for TREC Million Query Track (2007 - 2009)


mam10eks:
Is there a reason why `src/main/resources/topics-and-qrels` does not contain qrels/prels + topics for the Million Query Tracks?

I can include the million query resources together with regressions (because I would need them).

lintool:
Yup, contributions would be welcome! Let me know if you need any help hooking into our setup.

mam10eks:
The pull request #1036 would add all the required resources. (Although we may discuss whether it is ok to add the "inofficial" qrels.)

mam10eks:
Done with #1036

PR #1036, opened 2020-03-11
WIP: Add million query

mam10eks:
This merge request adds all resources + unit tests for processing the TREC Million Query Tracks 2007, 2008, and 2009.

This will enable the regressions mentioned in #1034.

Attention, the Million Query tracks have only used prels, and I have converted them to "inofficial" qrels for simplified evaluation with trec_eval/ndeval.

The transformation from prels to qrels was performed with:
```
cat prels.mq.1-10000.txt| awk '{print $1 " 0 " $2 " " $3}' > qrels.mq-inofficial.1-10000.txt
cat prels.mq.10001-20000.txt| awk '{print $1 " 0 " $2 " " $3}' > qrels.mq-inofficial.10001-20000.txt
cat prels.mq.20001-60000.txt| awk '{print $1 " 0 " $2 " " $3}' > qrels.mq-inofficial.20001-60000.txt
```

I think this is ok as long as we explicitly mention that. (Still, it is interesting to try algorithms in such scenarios, especially as those "inofficial" qrels match the ground truth that is used in LETOR 4.0).

Should we add regressions on the new million query resources within this issue or within a next issue?
The following command produces a run file for Million Query 2007, and evaluates it (Index is created with the command mentioned in the [gov2 regression](https://github.com/castorini/anserini/blob/master/docs/regressions-gov2.md)):
```
target/appassembler/bin/SearchCollection -index lucene-index.gov2.pos+docvectors \
 -topicreader Web -topics src/main/resources/topics-and-qrels/topics.mq.1-10000.txt \
 -bm25 -output run.gov2.bm25.mq2007.1-10000.txt

./eval/ndeval/ndeval src/main/resources/topics-and-qrels/qrels.mq-inofficial.1-10000.txt run.gov2.bm25.mq2007.1-10000.txt
```


src/main/java/io/anserini/search/topicreader/PrioritisedWebTopicReader.java
```diff
@@ -0,0 +1,33 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class PrioritisedWebTopicReader extends TopicReader<Integer> {
```
lintool: Being American/Canadian, `-ize`?

src/test/java/io/anserini/search/topicreader/PrioritisedWebTopicReaderTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+import java.util.SortedMap;
+
+import org.junit.Test;
+
+public class PrioritisedWebTopicReaderTest {
```
lintool: Same...

lintool:
> Attention, the Million Query tracks have only used prels, and I have converted them to "inofficial" qrels for simplified evaluation with trec_eval/ndeval.

I'm actually not familiar with this...

Is this generally accepted practice? If so, can you add a reference? We can use that as justification...

mam10eks:
I don't think that the conversion from prels to qrels is accepted practice.
In fact, they explicitly recommend against it, since [prels explicitly includes only a sample of judgments, and no complete relevance judgments like typically assumed for qrels (see the Prels section)](http://ciir.cs.umass.edu/million/results07.html).
The only reference that is doing that conversion is LETOR 4.0 (but there the case is slightly different since the task is to rerank only judged documents).

I am also fine if we remove those "inofficial qrels".
In that case, we must use [mtc_eval](http://ir.cis.udel.edu/~carteret/downloads.html) for evaluation.
I often preferred to work with those "inofficial qrels" and trec_eval, because mtc_eval is not easy to use (difficult to compile nowadays, segmentation faults at runtime).

src/main/java/io/anserini/search/topicreader/PrioritisedWebTopicReader.java
mam10eks: Done

src/test/java/io/anserini/search/topicreader/PrioritisedWebTopicReaderTest.java
mam10eks: Done

lintool:
Yes, please remove the `inofficial` qrels and I think this should be ready for merging.

mam10eks:
I have removed the `inofficial` qrels.

lintool:
Thanks for your contribution!

PR #1037, opened 2020-03-12
Release notes for v0.8.0

lintool:
@edwinzhng @nikhilro Quick clarification:

> + Added basic Solrini support for CORE collections, bibtex collections, and the ACL Anthology.

Is this actually true?


lintool:
> ACL and CORE yes basic functionality, bibtex not really tested on anything but _should_ work

Okay, good enough!

PR #1038, opened 2020-03-14
Refactored SearchCollection + added tests

lintool:
Previously, in `SearchCollection`, we had:

```
  private List<TaggedSimilarity> similarities;
  private Map<String, RerankerCascade> cascades;
```
Refactored the second to:

```
  private List<RerankerCascade> cascades;
```

By added the ability to "tag" the cascade. This makes the code cleaner. Added associated tests.


lintool:
hi @edwinzhng @nikhilro can you take a look at this PR?

I spoke to @Peilin-Yang out of band, and he doesn't have the bandwidth right now.

Thanks!

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -206,7 +193,7 @@ public void run() {
         out.close();
         final long durationMillis = TimeUnit.MILLISECONDS.convert(System.nanoTime() - start, TimeUnit.NANOSECONDS);
 
-        LOG.info("[End] " + id);
+        LOG.info("[End  ] " + id);
```
edwinzhng: nitpick: not sure if you meant to include the two extra spaces in `[End  ]`

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -273,73 +260,82 @@ public SearchCollection(SearchArgs args) throws IOException {
...
       }
+    } else if (args.qld) {
+      for (String mu : args.qld_mu) {
+        similarities.add(new TaggedSimilarity(new LMDirichletSimilarity(Float.valueOf(mu)),
+            String.format("qld(mu=%s)", mu)));
+      }
+    } else if (args.qljm) {
+      for (String lambda : args.qljm_lambda) {
+        similarities.add(new TaggedSimilarity(new LMJelinekMercerSimilarity(Float.valueOf(lambda)),
+            String.format("qllm(lambda=%s)", lambda)));
```
edwinzhng: `qljm`, not `qllm`?

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -349,48 +345,49 @@ public void close() throws IOException {
...
+                cascade.add(new ScoreTiesAdjusterReranker());
+                cascades.add(cascade);
+              }
             }
+          }
         }
+      }
     } else {
-      RerankerCascade cascade = new RerankerCascade();
+      RerankerCascade cascade = new RerankerCascade("default");
```
edwinzhng: Since we use the "default" tag a lot, maybe add a constructor with no arguments that automatically sets the tag as "default"? Then we can just use `new RerankerCascade()` here and elsewhere

src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -123,7 +123,7 @@ public SimpleSearcher(String indexDir, Analyzer analyzer) throws IOException {
     this.analyzer = analyzer;
     this.searchtweets = false;
     this.isRerank = false;
-    cascade = new RerankerCascade();
+    cascade = new RerankerCascade("default");
```
edwinzhng: same as above, default constructor?

src/main/java/io/anserini/search/SearchCollection.java
lintool: I wanted to do that to line up with `[Start]`.

src/main/java/io/anserini/search/SearchCollection.java
lintool: Good catch.

src/main/java/io/anserini/search/SearchCollection.java
lintool: 👍

src/main/java/io/anserini/search/SimpleSearcher.java
lintool: 👍

src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -160,7 +160,7 @@ public void setLanguage(String language) {
 
   public void unsetRM3Reranker() {
     this.isRerank = false;
-    cascade = new RerankerCascade();
+    cascade = new RerankerCascade("default");
```
edwinzhng: Sorry, missed this one too

PR #1040, opened 2020-03-20
ExtractDocumentLengths: prints out sum of doclengths, both lossy and lossless

lintool:
Ref: https://github.com/osirrc/ciff/issues/21

Adds check in ExtractDocumentLengths per above issue.


Chriskamphuis:
(I suppose this problem already existed)

lintool:
Fixed the issue you mentioned while I was at it...

PR #1043, opened 2020-03-22
COVID Collection and Generator for Sematic Scholar Corpus

nikhilro:
CovidCollection and CovidGenerator to index [COVID-19 Open Research Dataset](https://pages.semanticscholar.org/coronavirus-research)

### DIFF
--- a/src/main/java/io/anserini/collection/CovidCollection.java
+++ b/src/main/java/io/anserini/collection/CovidCollection.java
@@ -1,0 +1,149 @@
+/*
+ * Anserini: A Lucene toolkit for replicable information retrieval research
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


src/main/java/io/anserini/collection/CovidCollection.java
```diff
@@ -0,0 +1,149 @@
...
+      this.record = record;
+
+      // index full text into raw contents
+      if (record.get("has_full_text").contains("True")) {
+        String[] hashes = record.get("sha").split(";");
+        String fullTextPath = "/" + record.get("full_text_file") + "/" + hashes[hashes.length - 1].strip() + ".json";
+        try {
+          String fullTextJson = new String(Files.readAllBytes(
+            Paths.get(CovidCollection.this.path.toString() + fullTextPath)));
+          contents += "\n " + fullTextJson;
```
edwinzhng: Might be good to replace all instances of "\n" with " " in the CSV record string so that this is the first newline

lintool:
@edwinzhng @nikhilro merge when ready. don't worry about the hit in test coverage, we'll come back and patch it up later...

PR #1047, opened 2020-03-22
SimpleSearcher refactoring, clarifies 'contents' vs. 'raw' field

lintool:
Fixes confusion between `contents` and `raw` fields. The `contents` field holds the contents as indexed; the `raw` field holds the raw docs.

These may be different. For example, in the COVID collection, "raw" comprises the metadata record CSV and the raw JSON of the full text. The "contents" field comprises elements that we might have pulled out of the JSON. We search against "contents", _but_ downstream applications might want to use the "raw".

Fixed `SimpleSearcher` so that this confusion is reduced from the Python end.

Note that this is a Pyserini breaking change. We will need to publish new artifact and make adjustments on the Pyserini end.

### DIFF
--- a/src/main/java/io/anserini/index/IndexReaderUtils.java
+++ b/src/main/java/io/anserini/index/IndexReaderUtils.java
@@ -332,6 +332,7 @@ public class IndexReaderUtils {
 
   /**
    * Returns the document vector for a particular document as a map of terms to term frequencies.
+   *
    * @param reader index reader
    * @param docid collection docid
    * @return the document vector for a particular document as a map of terms to term frequencies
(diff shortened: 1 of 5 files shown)


lintool:
@nikhilro good call.

Actually, `SourceDocument` should just have `raw()` I think:
https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/collection/SourceDocument.java#L37

That's the "raw" original document. We should depend on the `Generator` to transform it into the indexed form.

lintool:
In my original design, a `Transformer`: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/transform/StringTransform.java

Would convert the raw to contents...

This might need to be all rethought now...

Issue #1048, opened 2020-03-22
Discussion point: SourceDocument design


lintool:
Follow up to #1047

We need to rethink the indexing pipeline re: "contents" and "raw", because there's a lot of confusion, mostly because the design has evolved over time...

In the Lucene document, we've agreed (and this is clear) that "contents" field should be the content as indexed for searching, and "raw" should be the raw source document.

What about `SourceDocument` then? Currently, it only has `content()` method, and it's underspecified as to what it should provide. The `Generator` is responsible for calling `content()` and pulling out the necessary parts to populate the Lucene "contents" field and "raw" field.

Does this design make sense? The alternative is that `SourceDocument` has `contents()` and `raw()` methods, i.e., it "knows" its own contents and raw form.

One argument against this is when we have one-to-many `SourceDocument` to Lucene `Document` mappings - like in COVID19 (see #1044), where we take a single `SourceDocument`, which may be full text, and index it at the paragraph level (i.e., multiple Lucene `Documents`)

@nikhilro @edwinzhng thoughts?

If we stick to the original design, then it's a matter of clarifying in documentation. If we adopted the alternative design, it's going to be a painful refactoring job...

nikhilro:
I'm about to create a PR which could serve as an example

nikhilro:
> One argument against this is when we have one-to-many SourceDocument to Lucene Document mappings - like in COVID19 (see #1044), where we take a single SourceDocument, which may be full text, and index it at the paragraph level (i.e., multiple Lucene Documents)

This doesn't hold because the way I implemented there is one-to-one mapping

edwinzhng:
I like the separate fields too for contents and raw, allowing the SourceDocument to build it's own separation makes things more organized imo

nikhilro:
Meat of the resolution in this comment in PR #1049: https://github.com/castorini/anserini/pull/1049#issuecomment-602228384

lintool:
Okay, this will be a fairly painful refactoring to get everything in this format, so we'll punt it for later.

lintool:
Actually, if we take this design to the logical conclusion, should the generator also be folded into the collection?

That is, each `SourceDocument` knows how to generate its own corresponding Lucene document? Or, do we like the separation between Lucene and non-Lucene code?

nikhilro:
I don't even see the need for any refactoring because there is an implicit assumption that `content` == `raw` unless `SourceDocument` for that `Collection` specifies `raw`, no?

[...]

Issue #1045, opened 2020-03-22
COVID-19 documentation tweaks


lintool:
Comments, minor fixes: roll up in your next PR:
https://github.com/castorini/anserini/blob/master/docs/experiments-covid.md

**1.**

`${DATA_DIR}` -> add the date of the collection in the path; I assume we're going to get frequent updates, so let's not get them confused.

**2.**

```
ls ${DATA_DIR}/*.tar.gz | xargs --replace tar -zxvf {} -C ${DATA_DIR}
```

I don't think `--replace` works on Mac.

nikhilro:
1. will do 
2. @lintool can you confirm if `-I` instead works on mac

lintool:
nope:

```
$ ls ${DATA_DIR}/*.tar.gz | xargs -I tar -zxvf {} -C ${DATA_DIR}
xargs: illegal option -- z
usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements]] [-J replstr]
             [-L number] [-n number [-x]] [-P maxprocs] [-s size]
             [utility [argument ...]]
```

It's only four files.... maybe it's easier just to uncompress each explicitly, rather than futz with cross-platform compatibility? Like, what about Windows?

lintool:
Also, next iteration, add link from main `README.md`.

nikhilro:
Resolved by #1049

PR #1049, opened 2020-03-22
Covid Passage Indexing

### LINKED ISSUE #1044: COVID-19 collection: Implement passage indexing
As discussed: an article = ID, TITLE + ABSTRACT (+ P1 + P2 + P3)?...

Decompose into:

+ ID, TITLE + ABSTRACT
+ ID.0001, TITLE + ABSTRACT + P1
+ ID.0002, TITLE + ABSTRACT + P2

nikhilro:
Resolves #1044 and #1045 

Per the issue, a record is decomposed into (id, content, raw):
* ID, TITLE + ABSTRACT, FULL_TEXT_JSON
* ID.0001, TITLE + ABSTRACT + P1, ""
* ID.0002, TITLE + ABSTRACT + P2, ""

Edits to documentation to resolve `xargs` issues and updates to expected number of documents.

In regards to #1048, the CovidCollection's SourceDocument adds `raw` which is ultimately what goes to `raw` for the document. This is the simplest implementation I think

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ For the most part, these runs are based on [_default_ parameter settings](https:
 The experiments described below are not associated with rigorous end-to-end regression testing and thus provide a lower standard of replicability.
 For the most part, manual copying and pasting of commands into a shell is required to replicate our results:
 
++ [Experiments on COVID-19 Open Research Dataset](docs/experiments-covid.md)
 + [Replicating "Neural Hype" Experiments](docs/experiments-forum2018.md)
 + [Guide to running BM25 baselines on the MS MARCO Passage Retrieval Task](docs/experiments-msmarco-passage.md)
 + [Guide to running BM25 baselines on the MS MARCO Document Retrieval Task](docs/experiments-msmarco-doc.md)
--- a/docs/experiments-covid.md
+++ b/docs/experiments-covid.md
@@ -25,18 +25,52 @@ ls "${DATA_DIR}"/*.tar.gz | xargs -I {} tar -zxvf {} -C "${DATA_DIR}"
 # tar -zxvf "${DATA_DIR}"/biorxiv_medrxiv.tar.gz -C "${DATA_DIR}"
 ```
 
-We can now index these docs as a `CovidCollection` using Anserini:
+We can now index these docs as using Anserini; we have three versions:
 
-```bash
-sh target/appassembler/bin/IndexCollection \
-  -collection CovidCollection -generator CovidGenerator \
-  -threads 8 -input "${DATA_DIR}" \
-  -index "${DATA_DIR}"/lucene-index-covid \
-  -storePositions -storeDocvectors -storeRawDocs -storeTransformedDocs
-```
+* `CovidCollection` which adds `title` + `abstract` to Lucene Document's `content`
 
-The output message should be something like this:
+  ```bash
+  sh target/appassembler/bin/IndexCollection \
+    -collection CovidCollection -generator CovidGenerator \
+    -threads 8 -input "${DATA_DIR}" \
+    -index "${DATA_DIR}"/lucene-index-covid \
+    -storePositions -storeDocvectors -storeRawDocs -storeTransformedDocs
+  ```
 
-```bash
-2020-03-22 15:24:49,305 INFO  [main] index.IndexCollection (IndexCollection.java:845) - Total 1,096,241 documents indexed in 00:11:35
-```
+  The output message should be something like this:
+
+  ```bash
+  2020-03-22 16:55:00,711 INFO  [main] index.IndexCollection (IndexCollection.java:845) - Total 44,145 documents indexed in 00:01:07
+  ```
+
+* `CovidFullTextCollection` which adds `title` + `abstract` + `full json text` to Lucene Document's `content`
+
+  ```bash
+  sh target/appassembler/bin/IndexCollection \
+    -collection CovidFullTextCollection -generator CovidGenerator \
+    -threads 8 -input "${DATA_DIR}" \
+    -index "${DATA_DIR}"/lucene-index-covid-full-text \
+    -storePositions -storeDocvectors -storeRawDocs -storeTransformedDocs
+  ```
+
+  The output message should be something like this:
+
+  ```bash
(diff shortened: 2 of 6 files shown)


lintool:
hey @nikhilro I'm thinking (but up for discussion) that the paragraph chopping should be performed in the `Generator`?

That is each `SourceDocument` still presents an independent article, but the `Generator` does the chopping into multiple paragraphs and a Lucene document for each.

nikhilro:
My mental model for `Collection` vs `Generator` is that the collection deals with actual files on the system whereas the generator never has to worry about that. To generator, there is a guarantee that the document data exists and it's job is to convert it to `LuceneDocument`. So, I think it makes more sense to keep it in `CovidCollection`

lintool:
In this design, each `Collection` has arbitrary flexibility in presenting what it thinks its "content" should be. 

In this case, it'd be `CSVRecord(\nJSON_FULL_TEXT)?`. The Generator knows what to do with this in generating Lucene `Documents`.

nikhilro:
> In this design, each Collection has arbitrary flexibility in presenting what it thinks its "content" should be.

confused, the current implementation?

lintool:
> My mental model for `Collection` vs `Generator` is that the collection deals with actual files on the system whereas the generator never has to worry about that. To generator, there is a guarantee that the document data exists and it's job is to convert it to `LuceneDocument`. So, I think it makes more sense to keep it in `CovidCollection`

But what about the case if we want to change the indexing scheme? Let's say later on we find out that paragraph by paragraph isn't that good, and we want to change to, say sliding window of 2 paragraphs?

In your design, you'd need to modify the collection.

And if you want to keep both, you'd need two collections.

lintool:
> > In this design, each Collection has arbitrary flexibility in presenting what it thinks its "content" should be.
> 
> confused, the current implementation?

Yes, that's the implicit contract of the current design, I think.

nikhilro:
> In your design, you'd need to modify the collection.
And if you want to keep both, you'd need two collections.

Yes, that seems reasonable to me. Maybe Collection isn't the best word for it but `Generator` to seems has a simple job. Convert 1 `SourceDocument` to 1 `LuceneDocument`

lintool:
> > In your design, you'd need to modify the collection.
> > And if you want to keep both, you'd need two collections.
> 
> Yes, that seems reasonable to me. Maybe Collection isn't the best word for it but `Generator` to seems has a simple job. Convert 1 `SourceDocument` to 1 `LuceneDocument`

I think I like the contract of Generator as Convert 1 `SourceDocument` to _N_ `LuceneDocument`.

So in my design, you'd have 1 Collection, but multiple generators. E.g.,

+ `CovidFullDocumentGenerator`
+ `CovidParagraphGenerator`
+ `CovidSlidingParagraphWindowGenerator`

nikhilro:
To implement `1 SourceDocument to N LuceneDocument`, there'd be two complications:

* We need to change pieces of our core design. Right now, all a generator does is `createDocument(SourceDocument)` which is expected to return 1 LuceneDocument. Each generator will have to implement some complex iterator logic to support the aforementioned 1:N SourceDocument:LuceneDocument.

* Even if we could change the above, there is an additional complexity on creating those N `LuceneDocument`s in the `Generator`. This will involve JsonNode, CSVParser, BufferedReader etc. having to move to `Generator` in addition to staying in `Collection`.

edwinzhng:
> To implement `1 SourceDocument to N LuceneDocument`, there'd be two complications:
> 
> * We need to change pieces of our core design. Right now, all a generator does is `createDocument(SourceDocument)` which is expected to return 1 LuceneDocument. Each generator will have to implement some complex iterator logic to support the aforementioned 1:N SourceDocument:LuceneDocument.
> * Even if we could change the above, there is an additional complexity on creating those N `LuceneDocument`s in the `Generator`. This will involve JsonNode, CSVParser, BufferedReader etc. having to move to `Generator` in addition to staying in `Collection`.

In that case, we would just need to modify the IndexCollection to support a list of returned documents instead of a single one (and probably modify all existing generators to return a list of size == 1)

https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/IndexCollection.java#L195

lintool:
Okay, I think you've sold me. We'll go with your design.

But let's keep both `CovidCollection` (the old one) and also `CovidParagraphCollection` (new one)? Maybe refactor to inherit from same super class?

nikhilro:
Sounds good!

lintool:
Actually, @nikhilro can we have three versions? (1) title/abstract only, (2) full text, and (3) title/abstract/each paragraph.

Down the road, we probably want to evaluate them...

nikhilro:
and to clarify those three versions

* (1) title/abstract only: `CovidCollection`
  * (id, content, raw)
  * (ROW_NUMBER, TITLE + ABSTRACT, FULL_JSON_TEXT)

* (2) full text: `CovidFullTextCollection`
  * (id, content, raw)
  * (ROW_NUMBER, TITLE + ABSTRACT + FULL_JSON_TEXT, FULL_JSON_TEXT)

* and (3) title/abstract/each paragraph: ``CovidParagraphCollection`
  * (id, content, raw)
  * (ROW_NUMBER, TITLE + ABSTRACT, FULL_JSON_TEXT)
  * (ROW_NUMBER.001, TITLE + ABSTRACT + P1, "")
  * (ROW_NUMBER.002, TITLE + ABSTRACT + P2, "")
  * ....

lintool:
> and to clarify those three versions
> 
> * (1) title/abstract only: `CovidCollection`
>   
>   * (id, content, raw)
>   * (ROW_NUMBER, TITLE + ABSTRACT, FULL_JSON_TEXT)

First bullet is input, and remaining our outputs Lucene docs, right? If so, yes.

nikhilro:
yea, I was mentioning the format in each for the upcoming bullets (id, content, raw) :sweat_smile:. Sounds good, on it.

nikhilro:
Gonna get @edwinzhng 's help to get the inheritance working in Java

nikhilro:
```
sh target/appassembler/bin/IndexCollection \
    -collection CovidCollection -generator CovidGenerator \
    -threads 8 -input "${DATA_DIR}" \
    -index "${DATA_DIR}"/lucene-index-covid \
    -storePositions -storeDocvectors -storeRawDocs -storeTransformedDocs
``` 
works but same with `CovidFullTextCollection` because of inheritance issues that @edwinzhng will fix. Thanks!

PR #1052, opened 2020-03-22
Update experiments-covid.md

lintool:
Oops, sorry, broke previous instructions.


lintool:
Let me run through all instructions again before I merge...

nikhilro:
The second and third version of the Collection won't work right now

PR #1053, opened 2020-03-22
Refactor CovidCollectionDocument base class for CovidGenerator

edwinzhng:
Added a CovidCollectionDocument base class so that we can parameterize `CovidGenerator` properly for all 3 `CovidCollection` classes.

### DIFF
--- a/src/main/java/io/anserini/collection/CovidCollectionDocument.java
+++ b/src/main/java/io/anserini/collection/CovidCollectionDocument.java
@@ -1,0 +1,75 @@
+/*
+ * Anserini: A Lucene toolkit for replicable information retrieval research
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.anserini.collection;
+
+import org.apache.commons.csv.CSVRecord;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.IOException;
+import java.nio.file.Files;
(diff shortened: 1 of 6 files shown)


src/main/java/io/anserini/collection/CovidCollectionDocument.java
```diff
@@ -0,0 +1,75 @@
...
+
+import org.apache.commons.csv.CSVRecord;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+public abstract class CovidCollectionDocument implements SourceDocument {
```
lintool: Shouldn't this be an inner class to be consistent with all other collections?

src/main/java/io/anserini/collection/CovidCollectionDocument.java
```diff
@@ -0,0 +1,75 @@
...
+    String fullTextPath = "/" + record.get("full_text_file") + "/" + hashes[hashes.length - 1].strip() + ".json";
+    try {
+      String fullTextJson = new String(Files.readAllBytes(
+        Paths.get(basePath + fullTextPath)));
+      return fullTextJson;
+    } catch (IOException e) {
+      LOG.error("Error parsing file at " + fullTextPath);
+    }
+
+    return "";
```
lintool: I think a `null` here is more informative.

src/main/java/io/anserini/collection/CovidCollectionDocument.java
edwinzhng: Yes, the `CovidCollection`, `CovidFullTextCollection` and `CovidParagraphCollection` all have the inner class still, this is just in it's own file so we can do

```
public class CovidGenerator extends LuceneDocumentGenerator<CovidCollectionDocument>
```

as a type parameter for the `CovidGenerator`

Don't think it would make sense to make this an inner class of one of the specific `Collection` classes

lintool: Why not inner class of the base class?

edwinzhng: I don't think we need to make the entire `Collection` abstract since both the inner Segment and  Document classes would be completely different for all three, and there isn't much overlap.

lintool: okay, I don't think I have the proper context, so let's go with what you have for now.

PR #1054, opened 2020-03-22
SourceDocument refactor for separate raw field

edwinzhng:
Addresses #1048

Made the `SourceDocument` an abstract class so that we can have a separate `raw()` function that returns `content()` by default. I also added the respective tests and updated relevant collections.


lintool:
Ref: #1048

PR #1059, opened 2020-03-23
SourceDocument refactoring

lintool:
Major changes:

---
Created `DefaultLuceneDocumentGenerator`, which is the new default generator for converting a `SourceDocument` into a Lucene `Document`. 

Previously, `LuceneDocumentGenerator` had folded inside it a `StringTransform`; a concrete implementation like `JsoupStringTransform`, for example, stripped HTML tags from a span of text. Based on #1048 we decided that each collection should know it's own "contents" and "raw" fields, so the `StringTransform` has been folded inside each collection.

That is, the `StringTransform` *is* the code that converts "raw" to "contents", and this is now encapsulated inside each collection.

---

`SourceDocument` has been changed (back) to an interface. The `content()` method has been renamed to `contents()` to be consistent with Lucene `Document` fields and everything else.

---

Generators used to take in the counter from the indexer and manipulate it directly. This introduced a coupling that didn't really make sense. The generator now signals abnormal processing behavior (errors, skipped documents, empty documents) via exceptions that are caught in the indexer, which increments counters as appropriate.

---

`-storeContents` and `-storeRaw` are the two indexing flags now to store "contents" and "raw" fields, respectively.

---

Refactored Solr and ES bindings to ensure everything still works.


lintool:
hi @edwinzhng @nikhilro - this massive code dump is ready for review now. Moves us some ways along the SourceDocument 'contents` vs. `raw` discussion we had per #1048

src/main/java/io/anserini/index/generator/GeneratorExpection.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.anserini.index.generator;
+
+/**
+ * The root class of all exceptions that indicate abnormal behavior by {@link LuceneDocumentGenerator}.
+ */
+public class GeneratorExpection extends Exception {
```
edwinzhng, line 22: GeneratorException*? (Here and everywhere else we have "GeneratorExpection")

docs/regressions-core17.md
```diff
@@ -76,12 +76,12 @@ With the above commands, you should be able to replicate the following results:
 
 MAP                                     | BM25      | +RM3      | +Ax       | QL        | +RM3      | +Ax       |
 :---------------------------------------|-----------|-----------|-----------|-----------|-----------|-----------|
-[TREC 2017 Common Core Track Topics](../src/main/resources/topics-and-qrels/topics.core17.txt)| 0.2087    | 0.2823    | 0.2788    | 0.2032    | 0.2606    | 0.2613    |
+[TREC 2017 Common Core Track Topics](../src/main/resources/topics-and-qrels/topics.core17.txt)| 0.2087    | 0.2823    | 0.2739    | 0.2032    | 0.2606    | 0.2579    |
```
edwinzhng, line 79: Is there a reason these numbers dropped slightly?

src/main/java/io/anserini/index/generator/GeneratorExpection.java
lintool: oops, will fix.

docs/regressions-core17.md
lintool: I'll explain in the thread what happened here.

lintool:
Before merging, my explanation of why NYT regression numbers went down. I'll document in more detail in a follow-up PR, but just a quick sketch. Previously, NYT used `-generator JsoupGenerator`, which performs XML tag removal, unnecessarily - since the collection impl already cleaned up tags internally.

This means that documents with < and > were being interpreted as tags, and had content removed. So the number of terms in the collection went up; the number of empty documents decrease by one also.

With this bug fix, effectiveness went down slightly.

PR #1062, opened 2020-03-25
Improved integration testing

lintool:
Clearly separates different phases of the testing: building the index, checking the built index, and performing retrieval runs. Provides abstract methods for subclasses to override and supply ground truth and custom parameters.

Indexing calls `main` of `IndexCollection`. Since we're doing end-to-end testing, might as well test command-line parsing.

Added initial implementation of `BibtexEndToEndTest`.


lintool:
Figured out a better organization.

Issue #1064, opened 2020-03-26
SimpleSearcher flaky test case


lintool:
From time to time, `SimpleSearcher` test case fails with an exception along the lines of:

```
java.lang.AssertionError: The top-reader used to create Weight is not the same as the current reader's top-reader (org.apache.lucene.index.CompositeReaderContext@ca72a3
	at __randomizedtesting.SeedInfo.seed([699D6CD7153A8C02]:0)
	at org.apache.lucene.search.TermQuery$TermWeight.scorer(TermQuery.java:108)
	at org.apache.lucene.search.Weight.scorerSupplier(Weight.java:147)
	at org.apache.lucene.search.TopFieldCollector.populateScores(TopFieldCollector.java:479)
	at org.apache.lucene.search.IndexSearcher.searchAfter(IndexSearcher.java:552)
	at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:461)
	at io.anserini.search.SimpleSearcher.search(SimpleSearcher.java:321)
	at io.anserini.search.SimpleSearcher.search(SimpleSearcher.java:287)
	at io.anserini.search.SimpleSearcher.lambda$batchSearchFields$0(SimpleSearcher.java:238)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

PR #1065, opened 2020-03-26
Made methods of accessing 'contents' and 'raw' more consistent

lintool:
+ Both `IndexReaderUtil` and `SimpleSearcher` have similarly-named methods that do similar things now. In fact, `SimpleSearcher` dispatches to `IndexReaderUtil` for consistent implementations.
+ Better test cases
+ Also proposed fix for #1064 
+ Broke overly-long times

Note this will break Pyserini when we next push an artifact, but test cases there will catch.


src/main/java/io/anserini/index/IndexUtils.java
```diff
@@ -427,40 +425,6 @@ public void run() {
     LOG.info(String.format("Raw documents are output to: %s", outputDir));
   }
 
-  public String getTransformedDocument(String docid) throws IOException, NotStoredException {
```
edwinzhng: Why are we removing these again? Is it because you got rid of the LuceneDocId stuff in the last PR?

lintool: The transformer is the thing that converts "raw" to "contents", it's not relevant anymore based on our refactoring plans, so getting rid of it while I'm at it.

In general `IndexUtils` is legacy, we've been adding new functionality to `IndexReaderUtils`.

PR #1068, opened 2020-03-27
Instructions for indexing COVID collection into Solr

lintool:
@tteofili can you try out these instructions?


lintool:
I'm going to wait for @tteofili to test drive these instructions prior to merging.

lintool:
Since there's a new data drop of CORD from AI2, I'm going to merge this in so we have the latest instructions.

tteofili:
sorry for the late reply, anyway all is working on good on my side too, so 👍

lintool:
Thanks for trying the instructions out!

PR #1069, opened 2020-03-28
Update COVID for new datasets and standardize Solr schemas

edwinzhng:
* Updated the COVID generator and collection to support new fields in the March 27th dataset release
* Moved all Solr schemas into JSON files, updated documentation for setting schema


lintool:
This renders #1035 defunct right?

PR #1081, opened 2020-04-05
SimpleNearestNeighborSearcher should use correct cutoff factor

tteofili:
It was wrongly set to 0 in my initial patch yesterday, it should be instead set to 0.999f.


lintool:
It's really odd... I went over to Travis CI and it shows green for this patch, but the information isn't being relayed back here for some reason...

Overriding and merging.

PR #1092, opened 2020-04-11
Refactoring of WashingtonPost collection for Core18

lintool:
Refactoring of WashingtonPost collection based on the clarified definition of `contents()` and `raw()` in `SourceDocument`, per #1048.

Before, both `contents()` and `raw()` returned the raw JSON, and `WashingtonPostGenerator` extracted the article text for indexing.

Now, `raw()` returns the raw JSON, and `contents()` returns the extracted document contents. In other words, the logic for parsing the JSON has been moved from `WashingtonPostGenerator` into the collection itself. As a result, the generator has been simplified. It inherits from the `DefaultLuceneDocumentGenerator`, which creates the default fields, and then adds additional specific ones.

Regression values went down slightly for `Ax`, which seems very sensitive to tiny differences. The difference here is that, before, the "empty document" check was performed on the JSON, so it never triggered (since the JSON was never empty). With the new processing logic, the "empty document" check is performed on `contents()`, and so the number of empty documents is now accurate (there are 6). Apparently, this is enough to change regression numbers for `Ax`.


src/main/java/io/anserini/collection/WashingtonPostCollection.java
```diff
@@ -103,45 +107,85 @@ private void parseRecord(String record) {
...
     protected Optional<String> articleUrl;
     protected Optional<String> author;
     protected long publishDate;
     protected Optional<String> title;
-    protected String content;
+    protected String raw;
     protected WashingtonPostObject obj;
 
+    protected String fullCaption = null;
+    protected String kicker = null;
```
edwinzhng, line 130: What's a kicker?

lintool: It's one of the fields in WashingtonPost corpus: http://trec-news.org/guidelines-2020.pdf

> ... we decree that articles from the "Opinion", "Letters to the Editor", or "The Post's View" sections, as labeled in the "kicker" field, are not relevant

PR #1097, opened 2020-04-16
COVID-19 query generator

lintool:
Query generator specifically for COVID-19. If it finds a variant of COVID-19 in the query, replaces it with all the variants of COVID-19, including 2019-nCOV and SARS-COV-2.


src/test/java/io/anserini/search/query/BagOfWordsQueryGeneratorTest.java
```diff
@@ -29,7 +30,7 @@
 public class BagOfWordsQueryGeneratorTest {
   @Test
   public void test1() {
-    Analyzer analyzer = new EnglishAnalyzer();
+    Analyzer analyzer = IndexCollection.DEFAULT_ANALYZER;
```
rodrigonogueira4, line 33: What will be the default analyzer? Just curious, why don't we use the English analyzer?

lintool: This is a recent change to standardize the default analyzer... basically, we _are_ using the `EnglishAnalyzer()`

src/main/java/io/anserini/search/query/Covid19QueryGenerator.java
```diff
@@ -0,0 +1,74 @@
...
+    BooleanQuery.Builder builder = new BooleanQuery.Builder();
+    for (String t : tokens) {
+      builder.add(new TermQuery(new Term(field, t)), BooleanClause.Occur.SHOULD);
+    }
+
+    QueryParser parser = new QueryParser(IndexArgs.CONTENTS, analyzer);
+
+    try {
+      List<Query> disjuncts = new ArrayList<>();
+      disjuncts.add(parser.parse("\"COVID-19\""));
```
rodrigonogueira4, line 63: Is there a way to verify that these new terms will improve (and not hurt) effectiveness without having any annotation?

lintool: Nothing is sure without judgments... ;)

But: https://lucene.apache.org/core/8_0_0/core/org/apache/lucene/search/DisjunctionMaxQuery.html

src/test/java/io/anserini/search/query/BagOfWordsQueryGeneratorTest.java
rodrigonogueira4: Oh ok, thanks!

src/main/java/io/anserini/search/query/Covid19QueryGenerator.java
rodrigonogueira4: Ok!

Issue #1091, opened 2020-04-11
Better tokenization for COVID-19 and related terms


lintool:
Anserini currently uses [`DefaultEnglishAnalyzer`](https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/analysis/DefaultEnglishAnalyzer.java), but tokenizes "COVID-19" into "COVID" and "19". It'd be nice to fix this.

Other related terms - 2019-nCoV and SARS-CoV-2.

Perhaps the solution is to implement a `DefaultCovidEnglishAnalyzer` that subclasses `DefaultEnglishAnalyzer`? Takes a static list of tokens that's explicitly is _not_ broken up?

tteofili:
I've added a first version of such an Analyzer which uses Lucene's ClassicAnalyzer instead of StandardAnalyzer.
Another possible improvement would be not to lowercase tokens related to e.g. proteins, like "A protein" becomes "a" and "protein" and so a very relevant term would finish in the bucket of highly unrelevant (or even stopwords) terms.

PR #1098, opened 2020-04-17
Add a Covid-19 analyzer

tteofili:
First basic implementation of an analyzer for covid-19 data, see #1091.
A first requirement would be not to split at hyphens (covid-19, sars-cov-2, etc.).
It can be improved iteratively.


lintool:
hi @tteofili - just a follow up on this: I think the best course of action is to wait for TREC-COVID round 1 results so that we have relevance judgments to measure the effects of this.

PR #1099, opened 2020-04-17
Aligning methods for fetching doc in IndexReaderUtils and SimpleSearcher

lintool:
`IndexReaderUtils` and `SimpleSearcher` should have the same methods for access documents. Implements `documentByField`, which was previous missing in `SimpleSearcher`. Also aligns documentation so they say the same thing.

### DIFF
--- a/src/main/java/io/anserini/index/IndexReaderUtils.java
+++ b/src/main/java/io/anserini/index/IndexReaderUtils.java
@@ -394,7 +394,7 @@ public class IndexReaderUtils {
       ScoreDoc[] hits = rs.scoreDocs;
 
       if (hits == null || hits.length == 0) {
-        // Silently eat the error and return -1
+        // Either the id doesn't exist or there are multiple documents with the same id. In both cases, return null.
         return null;
       }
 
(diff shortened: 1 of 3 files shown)


tteofili:
+1

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -375,6 +375,36 @@ public static Document document(IndexReader reader, String docid) {
...
+   */
+  public static Document documentByField(IndexReader reader, String field, String id) {
+    try {
+      IndexSearcher searcher = new IndexSearcher(reader);
+      Query q = new TermQuery(new Term(field, id));
+      TopDocs rs = searcher.search(q, 1);
+      ScoreDoc[] hits = rs.scoreDocs;
+
+      if (hits == null || hits.length == 0) {
+        // Silently eat the error and return -1
```
rodrigonogueira4, line 397: maybe remove/change this comment as this is not exactly an error?

lintool: Yup!

PR #1106, opened 2020-04-19
Improved test cases for Covid collections

lintool:
+ Added article without full text as sample
+ Added test case for `CovidFullTextCollection`
+ Previously, `CovidFullTextCollection` just lumps JSON to `contents()` - now properly extracts `body_text`


lintool:
> LGTM, perhaps we should also add a test for when the PMC parse is not available and we fall back to the PDF parse?

This command brings up such and example

```
grep 'True,False,' metadata.csv  | head -1
```

lintool:
@edwardhdlu addressed CR. Take another quick look?

PR #1108, opened 2020-04-20
Improved repo layout with runs/ and logs/ directories

lintool:
Instead of various scripts dumping output in top level directory, created `runs/` and `logs/` for output, checked into GitHub with a dummy placeholder.


logs/.placeholder
```diff
@@ -0,0 +1 @@
+# This is the default directory for logs. Placeholder so that directory is kept in git.
```
edwinzhng, line 1: Usually pretty common to use `.gitkeep` instead of `.placeholder` so we don't need to write a line in the file, but it's fine if you keep it like this too. Feel free to merge or change then merge

lintool: Oh, I didn't know about this convention. Will rename files and stick to it.

PR #1109, opened 2020-04-20
New indexes CORD-19 04/17 data drop

lintool:
@edwinzhng  @rodrigonogueira4 

Minor concern is that for paragraph index, size went up from 3.4G to 4.7G. The only relevant commit I can see that would impact this is: https://github.com/castorini/anserini/commit/cc3b0a3bd2eb83d8f4802fe2e57e37187586a5be where we us pmc json is available.

Thoughts?

### DIFF
--- a/docs/experiments-covid.md
+++ b/docs/experiments-covid.md
@@ -13,9 +13,9 @@ If you don't want to build the index yourself, you can download the latest pre-b
 
 | Type | Version | Size | Link| Checksum |
 |:-----|:--------|:-----|:----|:---------|
-| Title + Abstract | 2020-04-10 | 1.2G | [[Dropbox]](https://www.dropbox.com/s/j55t617yhvmegy8/lucene-index-covid-2020-04-10.tar.gz) | `ec239d56498c0e7b74e3b41e1ce5d42a`
-| Full-Text | 2020-04-10 | 3.3G | [[Dropbox]](https://www.dropbox.com/s/gtq2c3xq81mjowk/lucene-index-covid-full-text-2020-04-10.tar.gz) | `401a6f5583b0f05340c73fbbeb3279c8`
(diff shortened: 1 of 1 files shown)


lintool:
BTW, the full-text index (which we don't use anyway) is down from 2.2G to 3.3G; this is easily explained by https://github.com/castorini/anserini/commit/1a030d83ba6d5482c2eb2c9f8880c3345bcdd629 - we now properly parse out `contents()`, whereas before we just glommed on the entire JSON.

rodrigonogueira4:
> @edwinzhng @rodrigonogueira4
> 
> Minor concern is that for paragraph index, size went up from 3.4G to 4.7G. The only relevant commit I can see that would impact this is: [cc3b0a3](https://github.com/castorini/anserini/commit/cc3b0a3bd2eb83d8f4802fe2e57e37187586a5be) where we us pmc json is available.
> 
> Thoughts?

Yeah, that seems right, but @edwinzhng might note something else...

lintool:
Okay, I've taken a detailed look at this issue - it's basically this article: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4896250/

Run this and you'll see ;)

```
$ grep 'ij3ncdb6' metadata.csv 
```

Basically, the "abstract" is really really long:

```
$ grep 'ij3ncdb6' metadata.csv | wc
       1   26358  180437
```

So what happens is that the "abstract" gets appended to every paragraph... which results in an explosion of documents. This doesn't happen _as much_ with the pdf_json because the JSON is shorter.

Compare:

```
$ ls -l comm_use_subset/pmc_json/PMC4896250.xml.json 
-rw-r--r--  1 jimmylin  staff  3932453 Apr 10 16:08 comm_use_subset/pmc_json/PMC4896250.xml.json
$ ls -l comm_use_subset/pdf_json/a5293bb4f17ad25a72133cdd9eee8748dd6a4b8d.json 
-rw-r--r--  1 jimmylin  staff  1628393 Apr 10 16:06 comm_use_subset/pdf_json/a5293bb4f17ad25a72133cdd9eee8748dd6a4b8d.json
```

Okay, I'm satisfied that this isn't a bug now...

lintool:
I'll wait for @edwinzhng for a quick sanity check before merging.

PR #1114, opened 2020-04-24
Store CSV and full text in raw field for CORD-19

### LINKED ISSUE #1105: Store CSV record in JSON in raw()
Hi @edwinzhng - as I understand, in the current design of the Covid collections, if a document does not have full text, the CSV record is converted into a JSON and stored as the `raw()`. If there's full text, the full text itself is stored in the `raw()`. Right?

However, in the case of full text, there seems to be information in the CSV record (e.g., `cord_uid`, `pmcid`, etc.) that gets lost? Shouldn't we serialize the CSV record and add into the full-text JSON, e.g., under attribute `csv_metadata` (since `metadata` is already taken).

edwinzhng:
Fixes #1105

The `raw` field now contains the following for all documents:

```
{
   "cord_uid": ID,
   "has_full_text": true/false,
   "csv_metadata": { ... }
   ...(rest of fields from full text inline)
}
```

For tests, the reason the `raw` field length decreased is because I load the original full-text JSON string, add the new fields ("cord_uid", "has_full_text", "csv_metadata") and then re-serialize to string which removes a lot of whitespace characters.

### DIFF
--- a/src/main/java/io/anserini/collection/CovidCollection.java
+++ b/src/main/java/io/anserini/collection/CovidCollection.java
@@ -103,18 +103,14 @@ public class CovidCollection extends DocumentCollection<CovidCollection.Document
    */
   public class Document extends CovidCollectionDocument {
     public Document(CSVRecord record) {
+      this.record = record;
+
       id = record.get("cord_uid");
       content = record.get("title").replace("\n", " ");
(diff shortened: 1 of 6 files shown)


lintool:
lgtm, sorry I didn't get this until now...

lintool:
@edwinzhng go ahead and merge when you're ready.

Issue #1116, opened 2020-04-25
Rename classes from Covid* to Cord19*


lintool:
With the publication of the paper [CORD-19: The Covid-19 Open Research Dataset](https://arxiv.org/abs/2004.10706), I think it's important to accurately capture the _actual_ name of the dataset.

lintool:
The Solr stuff needs to be renamed from `covid` to `cord19` also:
https://github.com/castorini/anserini/blob/master/docs/experiments-covid.md

PR #1117, opened 2020-04-25
Updating CORD-19 to 4/24 data drop

lintool:
Per #1116 - note that the indexes have been renamed.


lintool:
@edwinzhng yes, we should rename collection classes also, but let's merge #1114 first.

PR #1136, opened 2020-05-02
Add support for stopwords from external files

niazarak:


lintool:
It'd be awesome to have a simple test case for this? Something that parallels:
https://github.com/castorini/anserini/blob/master/src/test/java/io/anserini/integration/TrecEndToEndWhitelistTest.java

PR #1137, opened 2020-05-03
PhraseQueryGenerator + Phrase df

Chriskamphuis:
Created a PhraseQueryGenerator to build phrase queries, and added a support function in the SimpleSearcher to find the document frequency of a phrase. Also added tests for this.

### DIFF
--- a/src/main/java/io/anserini/search/query/PhraseQueryGenerator.java
+++ b/src/main/java/io/anserini/search/query/PhraseQueryGenerator.java
@@ -1,0 +1,41 @@
+/*
+ * Anserini: A Lucene toolkit for replicable information retrieval research
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


src/main/java/io/anserini/search/query/PhraseQueryGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+
+import io.anserini.analysis.AnalyzerUtils;
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.index.Term;
+import org.apache.lucene.search.*;
+
+import java.util.List;
+
+/*
+ * Bag of Terms query builder
```
lintool, line 27: Outdate Javadoc?

src/main/java/io/anserini/search/query/PhraseQueryGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.anserini.search.query;
+
+import io.anserini.analysis.AnalyzerUtils;
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.index.Term;
+import org.apache.lucene.search.*;
```
lintool, line 22: We try to avoid import *?

PR #1149, opened 2020-05-05
Exposes TREC-COVID topics from UDel

lintool:
To provide convenient access. Will expose in Pyserini in a separate RP.


lukuang:
Tests past. LGTM!

PR #1159, opened 2020-05-07
Weighted Query Generator

Chriskamphuis:


lintool:
hey @Chriskamphuis I think this is worth a discussion. I see that you envision the class to be used like this, based on your test:

```
    Analyzer analyzer = IndexCollection.DEFAULT_ANALYZER;
    WeightedQueryGenerator weightedQueryGenerator = new WeightedQueryGenerator();
    weightedQueryGenerator.addTermsWithWeight("information", 1F);
    weightedQueryGenerator.addTermsWithWeight("retrieval",2F);
    Query query = weightedQueryGenerator.buildQuery("contents", analyzer, "test");
```

Previously, the contract for `QueryGenerator` was that `buildQuery` took some text (e.g., TREC topic title). So, it's stateless. But here, you're introducing internal state with `addFoo`, which might be confusing?

From what I understand of your goals, I think you basically want to expose a bunch of query building blocks that you can manipulate in Python? So maybe the right design is to create a util class with a much of static methods that correspond to your `addFoo` methods, and then do the rest of the query assembly in Python?

Chriskamphuis:
The problem that I faced was that buildQuery inherents its signature from the QueryGenerator superclass, so I had to work with it such that it is still usable in Pyserini, how its implemented right now (as you can give a QueryGenerator as parameter in the search). 

I understand it is confusing, but I didn't see another elegant way to provide the weights ..

Right now I don't really need other building blocks. I am not sure how you envision the static utils class. Something else what would be possible is to incorporate the weights in the query string somehow and make the object stateless that way. What do you think?

lintool:
How about we add the ability for `SimpleSearcher` in Pyserini to take a Lucene query directly? Then you can just expose a number of hooks for building pieces of queries (adding weights, etc.) and then "assemble" the final query all on the Python side?

Chriskamphuis:
Ok!

Chriskamphuis:
Then I just close this PR as this is already possible on the Anserini side.

PR #1160, opened 2020-05-07
Adds script to measure percentage of judged documents in a run

rodrigonogueira4:


src/main/python/measure_judged.py
```diff
@@ -0,0 +1,69 @@
...
+import argparse
+import collections
+
+from typing import Dict
+from typing import List
+from typing import Set
+
+
+def load_qrels(path: str) -> Dict[str, Set[str]]:
+    """Loads qrels into a dict of key: query_id, value: set of relevant doc
```
lintool, line 12: I don't think these line breaks are necessary? i.e., these lines aren't too long?

src/main/python/measure_judged.py
```diff
@@ -0,0 +1,69 @@
+"""Script to measure the percentage of judged documents at various
```
lintool, line 1: Can you add the standard boilerplate with Apache 2 headers etc.?

src/main/python/measure_judged.py
```diff
@@ -0,0 +1,69 @@
...
+                    'cutoffs.')
+    parser.add_argument('--qrels', type=str, required=True, help='qrels file')
+    parser.add_argument('--run', type=str, required=True, help='run file')
+
+    args = parser.parse_args()
+
+    qrels = load_qrels(args.qrels)
+    run = load_run(args.run)
+
+    for max_rank in [5, 10, 20, 30]:
```
lintool, line 58: maybe take this in as a command-line arg, e.g., `5,10,20,30` with this being a default?

lintool:
@x65han take a look also please.

src/main/python/measure_judged.py
rodrigonogueira4: I'm following the 80-char limit from PEP8. BTW, should we add linting to the repo?

src/main/python/measure_judged.py
rodrigonogueira4: Sure

src/main/python/measure_judged.py
rodrigonogueira4: Good idea!

rodrigonogueira4:
> comments for you...

Thanks for the review, PTAL

src/main/python/measure_judged.py
```diff
@@ -0,0 +1,84 @@
...
+    """Loads qrels into a dict of key: query_id, value: set of relevant doc
+    ids."""
+    qrels = collections.defaultdict(set)
+    with open(path) as f:
+        for i, line in enumerate(f):
+            line = ' '.join(line.split())
+            query_id, _, doc_id, relevance = line.rstrip().split(' ')
+            qrels[query_id].add(doc_id)
+            if i % 1000000 == 0:
+                print('Loading qrels {}'.format(i))
```
x65han, line 32: Tested this script on my local machine. My qrels file is short. Hence this script printed:
```
Loading qrels 0
judged@5: 0.2
judged@10: 0.15333333333333335
judged@20: 0.15666666666666662
judged@30: 0.14
Done
```
It threw me off a little bit when I see `Loading qrels 0` alone.

PR #1163, opened 2020-05-07
Adds run interpolation script

rodrigonogueira4:
Interpolate two runs by first normalizing their query-level scores (between 0 and 1) and then doing:
alpha * run1_score + (1 - alpha) * run2_score.

### DIFF
--- a/src/main/python/interpolate.py
+++ b/src/main/python/interpolate.py
@@ -1,0 +1,98 @@
+# -*- coding: utf-8 -*-
+'''
+Anserini: A Lucene toolkit for replicable information retrieval research
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+'''
+import argparse
+import collections
+import pytrec_eval
+
+
+def load_qrels(path):
+    qrels = collections.defaultdict(dict)
+    for line in open(path):
+        line = ' '.join(line.split())
+        query_id, _, doc_id, relevance = line.split(' ')
+        qrels[query_id][doc_id] = int(relevance)
+    return qrels
+
+
+def load_run(path):
+    run = collections.defaultdict(dict)
+    for line in open(path):
+        query_id, _, doc_id, _, score, _ = line.split(' ')
+        run[query_id][doc_id] = float(score)
+
+    return run
+
+
+parser = argparse.ArgumentParser(
+    description='Interpolate 2 runs by first normalizing their scores (between'
+                '0 and 1) and then doing: alpha * run1_score + (1 - alpha) * '
+                'run2_score.')
+parser.add_argument('--qrels', required=True, type=str, help='qrels file.')
+parser.add_argument('--run1', required=True, type=str, help='run 1 file.')
+parser.add_argument('--run2', required=True, type=str, help='run 2 file.')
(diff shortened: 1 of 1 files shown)


src/main/python/interpolate.py
```diff
@@ -0,0 +1,98 @@
+# -*- coding: utf-8 -*-
+'''
+Anserini: A Lucene toolkit for replicable information retrieval research
```
lintool, line 3: mangled boilerplate... use https://github.com/castorini/anserini/blob/master/src/main/python/run_es_regression.py#L1

src/main/python/interpolate.py
```diff
@@ -0,0 +1,98 @@
...
+def load_run(path):
+    run = collections.defaultdict(dict)
+    for line in open(path):
+        query_id, _, doc_id, _, score, _ = line.split(' ')
+        run[query_id][doc_id] = float(score)
+
+    return run
+
+
+parser = argparse.ArgumentParser(
```
lintool, line 37: If we don't have qrels, add an ability to just output a set alpha?

src/main/python/interpolate.py
```diff
@@ -0,0 +1,98 @@
...
+    description='Interpolate 2 runs by first normalizing their scores (between'
+                '0 and 1) and then doing: alpha * run1_score + (1 - alpha) * '
+                'run2_score.')
```
lintool, line 40: Are you setting width=100? These lines seem really short to need linebreaks...

src/main/python/interpolate.py
```diff
@@ -0,0 +1,98 @@
...
+    print('alpha:', alpha, 'score:', score)
+
+    if score > best_score:
+        best_run = run_fusion
+        best_score = score
+        best_alpha = alpha
+
+print('best_alpha:', best_alpha, 'best_score:', best_score)
+
+with open(args.output, 'w') as fout:
```
lintool, line 92: It appears that you don't require the doc to be in _both_ runs, right? If so, the union may yield more hits?

If so, maybe add a `-k` parameter to trim down to a set `k`, e.g., 1000?

src/main/python/interpolate.py
```diff
@@ -0,0 +1,98 @@
...
+    for query_id, doc_id_scores in best_run.items():
+        for rank, (doc_id, score) in enumerate(doc_id_scores.items()):
+            rank += 1
+            fout.write(f'{query_id} Q0 {doc_id} {rank} {score} Fusion\n')
```
lintool, line 96: Can we make `Fusion` into a `-runtag` param? We need this to generate final TREC runs...

src/main/python/interpolate.py
rodrigonogueira4: I'm using PEP8, which is 80 chars

lintool: > Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters), provided that comments and docstrings are still wrapped at 72 characters.

https://pep8.org/#maximum-line-length

haha, can we agree? ;)

src/main/python/interpolate.py
rodrigonogueira4: Sure

src/main/python/interpolate.py
rodrigonogueira4: Interesting... if we just copy and paste, the line breaks disappear. Anyway, fixed!

src/main/python/interpolate.py
rodrigonogueira4: haha, ok, I do prefer longer lines

src/main/python/interpolate.py
rodrigonogueira4: Sure

src/main/python/interpolate.py
rodrigonogueira4: Sure

rodrigonogueira4:
@lintool, PTAL

PR #1169, opened 2020-05-08
Expose query building blocks

Chriskamphuis:


lintool:
Great! I assume these are necessary because Pyjnius has trouble with statics? If so, a quick note to document would be great!

Also, please merge in latest master?

PR #1174, opened 2020-05-09
Adds script to convert TREC run to MS MARCO run

rodrigonogueira4:


src/main/python/msmarco/convert_trec_to_msmarco_run.py
```diff
@@ -0,0 +1,36 @@
...
+limitations under the License.
+"""
+
+import argparse
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Converts a TREC run file to a MS MARCO-formatted run file.')
+    parser.add_argument('--input_run', required=True, default='', help='TREC-formatted run file')
```
lintool, line 24: how about just `input` and `output` to be consistent with args from other scripts?

src/main/python/msmarco/convert_trec_to_msmarco_run.py
```diff
@@ -0,0 +1,36 @@
...
+    parser.add_argument('--output_run', required=True, default='',
+                        help='output MS MARCO-formatted run file')
```
lintool, line 26: Maybe add arguments to change the `-runtag` and `-k`?

src/main/python/msmarco/convert_trec_to_msmarco_run.py
```diff
@@ -0,0 +1,36 @@
...
+
+    args = parser.parse_args()
+
+    with open(args.output_run, 'w') as fout:
+        for line in open(args.input_run):
+            query_id, _, doc_id, rank, _, _ = line.strip().split(' ')
+            fout.write('{}\t{}\t{}\n'.format(
+                query_id, doc_id, rank))
+
+    print('Done!')
```
lintool, line 36: Maybe something more informative, like "Wrote output run to ...".

lintool:
Might as well make it more complete while we're at it?

src/main/python/msmarco/convert_trec_to_msmarco_run.py
ronakice: Also is there any reason to assume that the run file is sorted by rank? Because I feel like in most of our eval scripts we sort it again by rank regardless.

lintool: Yes, a check to make sure the ranks are strictly increase and scores are strictly decreasing would be good, and throw a warning.

Score ties are a hidden gotcha, and this will expose.

src/main/python/msmarco/convert_trec_to_msmarco_run.py
rodrigonogueira4: Sure

src/main/python/msmarco/convert_trec_to_msmarco_run.py
rodrigonogueira4: There is no runtag in MS MARCO run files, but I will add k

src/main/python/msmarco/convert_trec_to_msmarco_run.py
rodrigonogueira4: Sure

rodrigonogueira4:
@lintool @ronakice, thanks for your reviews, PTAL

ronakice:
LGTM! Thanks for adding this.

lintool:
lg, 🚢 it

PR #1175, opened 2020-05-09
add methods to expose n to pySearcher

stephaniewhoo:
I think leaving extracting the number here would be easier since we don't need to import IndexReader and IndexSearcher to the pyserini.


src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -206,6 +206,17 @@ public void setBM25Similarity(float k1, float b) {
     searcher.setSimilarity(similarity);
   }
 
+  /**
+   * Returns the number of documents in the provided index file. Used to expose the total number of documents in Pyserini
+   *
+   * @return the number of documents in total
+   */
+
+   public Int getTotalNumDocuments(){
```
lintool, line 208: primitive int instead if object Int.
I think there's already an `IndexSearcher` for you to read from? see line 96.

stephaniewhoo: I found line 96 was initialized to be a null and was assigned by some values in `setLMDirichletSimilarity`. Not sure if there are other assignments for this searcher

lintool: IIRC, the code initializes an `IndexSearcher` on demand, as in: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SimpleSearcher.java#L308

You method should probably do the same also?

If there are any cases that would yield an NPE, can you please add a test case to catch, and then fix in code while you're at it?

stephaniewhoo: added test cases, still trying to catch possible NPE. For now, haven't encountered that.

src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -199,6 +199,19 @@ public void setBM25Similarity(float k1, float b) {
...
+  /**
+   * Returns the number of documents in the provided index file. Used to expose the total number of documents in Pyserini
+   *
+   * @return the number of documents in total
+   */
+
+   public int getTotalNumDocuments(){
+     if (searcher == null) {
+      searcher = new IndexSearcher(reader);
+    }
```
lintool, line 211: fix indentation please.

src/test/java/io/anserini/search/SimpleSearcherTest.java
```diff
@@ -249,4 +249,11 @@ public void testFieldedBatchSearch() throws Exception {
 
     searcher.close();
   }
+
+  @Test
+  public void testTotalNumDocuments() throws Exception {
+    SimpleSearcher searcher = new SimpleSearcher(super.tempDir1.toString());
+    assertEquals(3,searcher.getTotalNumDocuments());
```
lintool, line 256: space after 3, 
remove empty newline.

lintool:
@stephaniewhoo can you please address those minor changes?

stephaniewhoo:
> @stephaniewhoo can you please address those minor changes?

Done formatting. Will be careful about the format next time! Sorry

PR #1179, opened 2020-05-10
Cleanup of regression script

lintool:
The regression documentation has drifted slightly away from the actual working regression commands - for example, the index paths. General cleanup to bring back into alignment.


src/main/python/run_regression.py
```diff
@@ -53,15 +53,12 @@ def check_output(command):
...
-        if not index_path or not os.path.exists(index_path):
-            for input_root in yaml_data['input_roots']:
-                if os.path.exists(os.path.join(input_root, yaml_data['index_path'])):
-                    index_path = os.path.join(input_root, yaml_data['index_path'])
-                    break
+    index_path = yaml_data['index_path']
+    if not index_path or not os.path.exists(index_path):
+        for input_root in yaml_data['input_roots']:
+            if os.path.exists(os.path.join(input_root, yaml_data['index_path'])):
+                index_path = os.path.join(input_root, yaml_data['index_path'])
```
lukuang, line 60: How about move this out of `if` and do checking `index_path`?

lintool: yup, good call!

PR #1183, opened 2020-05-11
remove_judged_docids script for TREC-COVID: added runtag option

lintool:
Lets user specify runtag from command line.


src/main/python/trec-covid/remove_judged_docids.py
```diff
@@ -54,6 +52,9 @@ def main():
                 score = float(cols[4])
                 tag = cols[5]
 
+                if args.runtag is not None:
```
rodrigonogueira4, line 55: The code below is preferred as it covers both `runtag == None` and `runtag == ''`
```suggestion
                if args.runtag:
```

x65han: Oh wow, good to know.

lintool:
I've confirmed!

```
>>> bool('')
False
>>> bool(None)
False
>>> bool('foo')
True
```

PR #1184, opened 2020-05-11
Adds script to visually inspect a run

rodrigonogueira4:


src/main/python/trec-covid/print_run.py
```diff
@@ -0,0 +1,42 @@
...
+import csv
+import utils
+from tqdm import tqdm
+
+
+def load_metadata(path):
+    metadata = {}
+    with open(path) as f:
+        for row in tqdm(csv.DictReader(f)):
+            metadata[row['cord_uid']] = row['title'], row['abstract']
```
x65han, line 12: In the future, we might need the value field to be an object with more info:

metadata[row['cord_uid']] = {
    "title": row['title'],
    ...
}

rodrigonogueira4: Yeah, makes sense!

PR #1189, opened 2020-05-13
Untangle SimpleSearcher form SimpleTweetSearcher + verification script

lintool:
Previously, `SimpleSearcher` had different code paths for searching "normal" documents and searching tweets (which has additional temporal logic). I've untangled into `SimpleSearcher` and `SimpleTweetSearcher` for better maintainability.

Also, I added a `verify_simplesearcher.py` script to confirm that the output of `SimpleSearcher` and `SimpleTweetSearcher` matches _exactly_ the output of `SearchCollection` - i.e., that the different code paths give identical results.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -133,6 +133,10 @@
               <id>SimpleSearcher</id>
             </program>
             <program>
+              <mainClass>io.anserini.search.SimpleTweetSearcher</mainClass>
+              <id>SimpleTweetSearcher</id>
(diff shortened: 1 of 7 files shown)


lintool:
Yes, we take a hit in terms of test coverage, but the additional new code is tested externally end-to-end by the Python verification script, so I think overall it's okay for now... can circle back and add test coverage on the Java side later.

PR #1191, opened 2020-05-14
Update collections for 05/12 CORD-19 dataset

edwinzhng:
* Updates collections to use new folder format and a few different field names
* Changed the test document without a full text parse since the other one we were using now actually has a full text parse


lintool:
For the record, changes are discussed here: https://discourse.cord-19.semanticscholar.org/t/faqs-about-cord-19-dataset/94

PR #1194, opened 2020-05-14
Computes query-doc score for a specific document

lintool:
See: https://github.com/castorini/pyserini/issues/100

cc/ @cmacdonald


cmacdonald:
+1
looks good to me. I mentioned to Jimmy the extension to multiple documents.

PR #1210, opened 2020-05-19
Replication of robust04, core17 & core18

x65han:
- macOS Mojave Version 10.14.6
- Python 3.7.2
- Java OpenJDK 11.0.2 2019-01-15

### DIFF
--- a/src/main/resources/docgen/templates/core17.template
+++ b/src/main/resources/docgen/templates/core17.template
@@ -44,6 +44,7 @@ ${effectiveness}
 
 ## Replication Log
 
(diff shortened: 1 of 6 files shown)


src/main/resources/docgen/templates/core17.template
```diff
@@ -44,6 +44,7 @@ ${effectiveness}
 
 ## Replication Log
 
-* Results replicated by [@tteofili](https://github.com/tteofili) on 2019-01-27 (commit [`951090`](https://github.com/castorini/Anserini/commit/951090b66230040f037dde46534d896416467337))
```
x65han, line 47: changing the bullets to `+` to be consistent with other templates.

PR #1217, opened 2020-05-22
Updates pyserini usage for Open Research retrieval

rodrigonogueira4:


rodrigonogueira4:
> hey @rodrigonogueira4 should we start checking things like this into pyserini repo?
> 
> E.g., `pyserini/scripts/openresearch/`? As pyserini becomes more and more feature complete, things that are not directly connect to Java should be in that repo?

As discussed on Slack, we should keep old python scripts in Anserini and add new ones to pyserini. Is that right?

PR #1218, opened 2020-05-23
Add index statistics function

PepijnBoers:
Adds function that returns selection of index statistics in map format (as previously printed in `IndexUtils`)

### DIFF
--- a/src/main/java/io/anserini/index/IndexReaderUtils.java
+++ b/src/main/java/io/anserini/index/IndexReaderUtils.java
@@ -30,6 +30,8 @@ import org.apache.logging.log4j.Logger;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.index.DirectoryReader;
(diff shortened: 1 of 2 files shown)


src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -768,4 +770,33 @@ public static String convertLuceneDocidToDocid(IndexReader reader, int docid) {
...
+   * @return Index statistics as a map of statistic's name to statistic.
+   * @throws IOException
+   */
+  public static Map<String, Object> getIndexStats(IndexReader reader) throws IOException {
+    Map<String, Object> indexStats = new HashMap<String, Object>();
+    try {
+      Terms terms = MultiTerms.getTerms(reader, IndexArgs.CONTENTS);
+
+      indexStats.put("documents", reader.numDocs());
+      indexStats.put("documents (non-empty)", reader.getDocCount(IndexArgs.CONTENTS));
```
lintool, line 787: shall we make these into pythonic strings like "non_empty_documents"?

Issue #1198, opened 2020-05-15
Test failures (Java 13, encoding issues)


cmacdonald:
```
(base) [craigm@trhead anserini]$export JAVA_HOME=/local/trmaster/opt/jdk-13.0.2
(base) [craigm@trhead anserini]$which mvn
alias mvn='/local/trmaster/opt/apache-maven-3.3.9/bin/mvn'
	/local/trmaster/opt/apache-maven-3.3.9/bin/mvn
(base) [craigm@trhead anserini]$mvn clean package appassembler:assemble install
[INFO] Scanning for projects...
[INFO]                                                                         
[INFO] ------------------------------------------------------------------------
[INFO] Building Anserini 0.9.3-SNAPSHOT
[INFO] ------------------------------------------------------------------------
Downloading: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-gpg-plugin/1.6/maven-gpg-plugin-1.6.pom
Downloaded: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-gpg-plugin/1.6/maven-gpg-plugin-1.6.pom (6 KB at 5.2 KB/sec)
Downloading: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-gpg-plugin/1.6/maven-gpg-plugin-1.6.jar
Downloaded: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-gpg-plugin/1.6/maven-gpg-plugin-1.6.jar (47 KB at 337.4 KB/sec)
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ anserini ---
[INFO] Deleting /users/tr.craigm/src/anserini/target
[INFO] 
[INFO] --- jacoco-maven-plugin:0.8.2:prepare-agent (default) @ anserini ---
[INFO] argLine set to -javaagent:/users/craigm/.m2/repository/org/jacoco/org.jacoco.agent/0.8.2/org.jacoco.agent-0.8.2-runtime.jar=destfile=/users/tr.craigm/src/anserini/target/jacoco.exec
[INFO] 
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ anserini ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] Copying 205 resources
[INFO] 
[INFO] --- maven-compiler-plugin:3.8.1:compile (default-compile) @ anserini ---
[...]

PR #1226, opened 2020-05-26
Replication for elastrini and solrini

shaneding:
- General: Had to use ./bin/qbuild.sh rather than the mvn clean package appassembler:assemble given in the docs (not sure if its just me or a general issue)

Elastrini and Solrini:
- MS MARCO Passage: No issues, but retrieval takes really long, maybe create shorter file just for replication, like in the original MS MARCO passage replication?


lintool:
> * General: Had to use ./bin/qbuild.sh rather than the mvn clean package appassembler:assemble given in the docs (not sure if its just me or a general issue)

What's your env setup? OS, Java version, etc. What errors are you getting?

shaneding:
Environment
OS: macOS 10.15 (Catalina)
Java: 14.0.1
Python: 3.7.6

Error:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test (default-test) on project anserini: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test failed: The forked VM terminated without saying properly goodbye. VM crash or System.exit called ? -> [Help 1]

lintool:
@shaneding 

> Java: 14.0.1

That's the issue... I think this is a known issue - https://github.com/castorini/anserini/issues/1198

shaneding:
Okay, so should ./qbuild.sh be used in place of "mvn clean package appassembler:assemble" for now?

lintool:
@shaneding I think the cleaner solution would be for you to downgrade to Java 11 so that you can stay in sync with the project... so you won't run into other JVM compat issues down the road...

Issue #1222, opened 2020-05-25
AttributeError in retrieving passages for MS MARCO passage ranking


kelvin-jiang:
In the replication of the MS MARCO passage ranking experiment, we retrieve passages by running
```
python ./src/main/python/msmarco/retrieve.py --hits 1000 --threads 1 \
 --index indexes/msmarco-passage/lucene-index-msmarco --qid_queries collections/msmarco-passage/queries.dev.small.tsv \
 --output runs/run.msmarco-passage.dev.small.tsv
```

This leads to the following error:
```
Traceback (most recent call last):
  File "./src/main/python/msmarco/retrieve.py", line 58, in <module>
    hits = searcher.search(query.encode('utf8'), int(args.hits))
  File "/home/kelvin/.local/lib/python3.6/site-packages/pyserini/search/pysearch.py", line 99, in search
    return self.object.search(JString(q.encode('utf8')), k)
AttributeError: 'bytes' object has no attribute 'encode'
```

This error is also encountered in https://github.com/castorini/anserini/pull/1220.

lintool:
@kelvin-jiang thanks for reporting. yup, we're aware of this issue... the repo is in rapid flux right now due to EMNLP deadlines and TREC COVID runs. We'll return to this once things quiet down.

lintool:
I believe this has been closed by #1238 

@kelvin-jiang can you please confirm and then close issue?

kelvin-jiang:
Confirmed!

PR #1229, opened 2020-05-26
Replicated MS MARCO passage ranking, doc ranking, and doc2query (MS MARCO + CAR) experiments

kelvin-jiang:
Environment
OS: Ubuntu 18.04.1 LTS
Java: openjdk 11.0.7 2020-04-14
Python: 3.6.9

MS MARCO passage ranking experiment:
* Fixed a command where one of the file paths was missing `collections/` prefix

doc2query experiment:
* Updated all of the file paths to follow the new `collections/` and `indexes/` repo structure
* Got a different word count for `wc collections/msmarco-passage/pred-test_topk10.txt`: line count and character count were still the same, the MD5 checksum was the same, and the final eval results were the same, so I updated the `wc` output

I also updated the username (it's me) in a replication entry for robust04.


lintool:
Did you rename your GitHub account names? I think the release notes need to be updated as well? https://github.com/castorini/anserini/blob/master/docs/release-notes/release-notes-v0.9.3.md

Can you grep through and fix those instances?

kelvin-jiang:
Yeah, I did. I could fix those instances in the release notes, but it's a bit problematic cause my old commits are still attributed to my old username in Github (not sure why Github does that), so the tool that collects the repo's contributors will still pick up my old username in the future.

lintool:
That's fine, I can do the mapping by hand in the future... but whatever you prefer...

lintool:
Please recheck with my branch: https://github.com/castorini/anserini/pull/1238

kelvin-jiang:
Reran the MS MARCO passage ranking experiment on updated master and #1222 seems to be resolved now.

kelvin-jiang:
Both files should be checked in already?

lintool:
oops, didn't notice all the way at the end...

PR #1233, opened 2020-05-27
Cord19 solr blacklight replication doc

shaneding:
An extension of experiments-cord19.md. 

I encountered this error during indexing through solr, but I think this is a result of the format of data rather than the code:
Request to collection [cord19] failed due to (400) org.apache.solr.client.solrj.impl.HttpSolrClient$RemoteSolrException: Error from server at http://192.168.1.20:8983/solr/cord19_shard1_replica_n1: ERROR: [doc=y6o5e2k5] multiple values encountered for non multiValued field source_x: [Elsevier, Medline, PMC], retry=0 commError=false errorCode=400


lintool:
@edwinzhng can you take a look? that multi-value issue sounds familiar.

@shaneding we can remove the Solr section from `experiments-cord19.md`, right?

docs/experiments-cord19-extra.md
```diff
@@ -0,0 +1,94 @@
+## Title: Ingesting CORD-19 into Solr and Blacklight
```
lintool, line 1: Remove "Title: "

shaneding:
Yes, the solr section in `experiments-cord19-extra.md` is basically a copy and paste from `experiments-cord19.md`, also just a note the error didn't actually stop the indexing, indexing just continued past it.

edwinzhng:
Also, would you be able to rename `covid.json` to `cord19.json` for consistency (and update the docs as well)?  😄

shaneding:
The error was resolved after updating the json file :+1:

PR #1238, opened 2020-05-28
Fixed MS MARCO docs to latest version of pyserini==0.9.3.0

lintool:
General cleanup as well.


src/main/python/msmarco/retrieve.py
```diff
@@ -43,10 +39,10 @@
     total_start_time = time.time()
 
     searcher = pysearch.SimpleSearcher(args.index)
-    searcher.set_bm25_similarity(float(args.k1), float(args.b))
+    searcher.set_bm25(float(args.k1), float(args.b))
```
rodrigonogueira4, line 42: Since we are here, add type=float to these arguments and remove the unnecessary `float`s here?

src/main/python/msmarco/retrieve.py
```diff
@@ -55,7 +51,7 @@
             start_time = time.time()
             for line_number, line in enumerate(open(args.qid_queries, 'r', encoding='utf8')):
                 qid, query = line.strip().split('\t')
-                hits = searcher.search(query.encode('utf8'), int(args.hits))
+                hits = searcher.search(query, int(args.hits))
```
rodrigonogueira4, line 54: same here, remove unnecessary `int`?

Issue #1166, opened 2020-05-07
Improvements to measure_judged.py


lintool:
Let's make the behavior closer to `trec_eval`?

Output should be 0.XXXX - four sig digits, and something like:

```
judged_cut_5   all  0.XXXX
judged_cut_10  all  0.XXXX
```
(tab separated)

With `-q` option, prints per topic scores (instead of `all`), just like `trec_eval`.

lintool:
We should make options take single dash instead of double dash, e.g., `-run` instead of `--run` to be consistent with `trec_eval` and other tools.

PR #1246, opened 2020-05-31
[#1166] eval/measure_judged.py improvments

adamyy:
## Summary 

As per described in https://github.com/castorini/anserini/issues/1166
* Print percentage as 0.XXXX
* Added `-q` option to print per topic measurement

## Example

```
$ python eval/measure_judged.py --qrels src/main/resources/topics-and-qrels/qrels.covid-round12.txt --cutoffs 10 --run runs/anserini.covid-r3.abstract.qq.bm25.txt -q

judged_cut_10   1       0.1000
judged_cut_10   2       0.4000
...
judged_cut_10   40      0.0000
judged_cut_10   all     0.3300

Done
```

### DIFF
--- a/eval/measure_judged.py
+++ b/eval/measure_judged.py
@@ -67,6 +67,9 @@ if __name__ == "__main__":
                         default=[5, 10, 20, 30],
                         help='Space-separate list of cutoffs. '
                              'E.g.: --cutoffs 5 10 20')
+    parser.add_argument('-q', action='store_true', dest='print_topic',
+                        help='In addition to summary evaluation,'
+                             'give evaluation for each query/topic')
 
     args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


eval/measure_judged.py
```diff
@@ -67,6 +67,9 @@ def load_run(path: str) -> Dict[str, List[str]]:
                         default=[5, 10, 20, 30],
                         help='Space-separate list of cutoffs. '
                              'E.g.: --cutoffs 5 10 20')
+    parser.add_argument('-q', action='store_true', dest='print_topic',
```
lintool, line 70: we use 120 char line width... so I don't think you need to break up the string?

adamyy: Out of curiosity, do we have linter set up for Python?

lintool: nope... we decided it (in a separate issue) that it was too heavyweight for a research project...

lintool: but we do stick to PEP8 as much as possible, we're just not so absolute about it.

PR #1250, opened 2020-06-02
Updated TREC-COVID Round 3 final submissions

lintool:


docs/experiments-covid.md
```diff
@@ -220,6 +235,16 @@ cut -d ' ' -f 1 runs/anserini.covid-r3.paragraph.qq.bm25.txt | sort | uniq | wc
 cut -d ' ' -f 1 runs/anserini.covid-r3.paragraph.qdel.bm25.txt | sort | uniq | wc
 ```
 
+To prepare the final runs for submission (removing judgments from round 1 and round 2):
+
+```bash
+python src/main/python/trec-covid/remove_judged_docids.py --qrels src/main/resources/topics-and-qrels/qrels.covid-round12.txt \
```
lukuang, line 241: I think we want this round12 qrels to be in the repo as well?

lintool: Around line 152, further up, I have:

```
cat src/main/resources/topics-and-qrels/qrels.covid-round1.txt src/main/resources/topics-and-qrels/qrels.covid-round2.txt > src/main/resources/topics-and-qrels/qrels.covid-round12.txt
```

So I don't think it's necessary? (It's part of the sequence of replication commands...)

lukuang: Yes, you are right.

PR #1265, opened 2020-06-07
Prune unused code paths for TREC News Track background linking task

### LINKED ISSUE #1258: Removed unused code in TREC News Tracks background linking baseline
For background linking, we have regressions for BM25, BM25+RM3, and BM25+RM3+DF: https://github.com/castorini/anserini/blob/master/docs/regressions-backgroundlinking18.md

I would like to remove code that's not involved in running one of those three conditions:

+ The other configurations and code paths are not tested.
+ According to TREC 2018 results, the other configurations aren't as effective anyway.

@Peilin-Yang is this okay with you? This is your code...
@PepijnBoers @Chriskamphuis I assume this is okay with you also?

lintool:
Closes #1258 

@PepijnBoers @Chriskamphuis please take a look?
@Peilin-Yang you're welcome take a look if you've got time!

I've got regressions queued up. Getting CR first, but won't merge until all the regressions pass.


src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -242,6 +242,22 @@ public static IndexReader getReader(String path) throws IOException {
...
 
+  /**
+   * Returns the document frequency of a term. Simply dispatches to <code>docFreq</code> but wraps the exception so
+   * that the caller doesn't need to deal with it; this is potentially dangerous but makes code less verbose.
+   *
+   * @param reader index reader
+   * @param term term
+   * @return tthe document frequency of a term
+   */
+  public static long getDF(IndexReader reader, String term) {
```
Chriskamphuis, line 253: The function above already provides this functionality right.

lintool: Correct, but this is a more efficient implementation., doesn't generate intermediate data... which makes a difference if `getDF` is used in an inner loop of some algorithm.

PR #1276, opened 2020-06-14
Experiment on different CORD-19 paragraph indexing conditions

lintool:
hey @rodrigonogueira4 remember when we had a discussion about exactly how to index paragraphs?

The old (existing) implement was:

+ docid: title + abstract
+ docid.00001: title + abstract + 1st paragraph
+ docid.00002: title + abstract + 2nd paragraph
+ docid.00003: title + abstract + 3rd paragraph
+ ...

But an equally reasonable alternative would be to *not* repeat the abstract:

+ docid: title + abstract
+ docid.00001: title + 1st paragraph
+ docid.00002: title + 2nd paragraph
+ docid.00003: title + 3rd paragraph
+ ...

I ran an experiment comparing the old and the new approaches, and the tl;dr is that the old approach is more effective:

|    | index | nDCG@10 | Judged@10 | MAP    | Recall@1k | Judged@1k |
|---:|:--------------|--------:|----------:|-------:|----------:|----------:|
|  1 | old qry+ques           | 0.4738  | 0.9457    | 0.2330 | 0.6332    | 0.2459    |
|  2 | old UDel     | 0.5508  | 0.9686    | 0.2572 | 0.6466    | 0.2411    |
|  3 | new qry+ques           | 0.4161  | 0.8800    | 0.2045 | 0.6190    | 0.2353    |
|  4 | new UDel   | 0.4865  | 0.8943    | 0.2273 | 0.6264    | 0.2307    |

I first ran retrieval to 50k hits on 5/19 index and then filtered down to only `cord_uid`s in the 5/1 corpus, and then evaluated on the union of round 1 and round 2 qrels. Each topic has 1k hits, so the comparison is fair.

For NDCG@10, new index is a lot lower, but judged is lower also. At depth 1k, judged is comparable, but MAP and recall for new is lower.

So I think we can conclude that the old indexing scheme is still better?

### DIFF
--- a/src/main/python/trec-covid/filter_run.py
+++ b/src/main/python/trec-covid/filter_run.py
@@ -1,0 +1,75 @@
+#
+# Pyserini: Python interface to the Anserini IR toolkit built on Lucene
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import argparse
+from collections import defaultdict
+
+
+def read_file(file):
+    docids = set()
+    with open(file) as f:
+        for line in f:
+            cols = line.split()
+            if len(cols) > 0:
+                docids.add(cols[0])
+    return docids
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Filter a TREC run.")
+    parser.add_argument('--whitelist', type=str, metavar='docids', help='qrels', required=True)
+    parser.add_argument('--input', type=str, metavar='run', help='input run', required=True)
+    parser.add_argument('--output', type=str, metavar='run', help='output run', required=True)
+    parser.add_argument('--runtag', type=str, default=None, metavar='runtag', help='run tag')
+    parser.add_argument('--k', type=int, default=1000, help='the number of results to keep per topic')
+    args = parser.parse_args()
+
+    docids = read_file(args.whitelist)
+    print(f'Read {len(docids)} docids from {args.whitelist}')
+    counts = defaultdict(int)
+
+    prev_score = None
+    check_score = True
+    with open(args.output, 'w') as output_f:
+        with open(args.input) as input_f:
+            for line in input_f:
+                cols = line.split()
+                qid = cols[0]
+                docid = cols[2]
+                score = float(cols[4])
+                tag = cols[5]
+
+                if args.runtag:
+                    tag = args.runtag
+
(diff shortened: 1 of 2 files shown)


src/main/python/trec-covid/filter_run.py
```diff
@@ -0,0 +1,75 @@
+#
```
lintool, line 1: We should probably merge with the script that removes judged docs, but let's keep separate for now.

lintool:
For the record, I used the following to filter out the "valid" ids:

```bash
cut -f 1 -d ',' collections/cord19-2020-05-01/metadata.csv > cord_uids-2020-05-01.txt
```

lintool:
Woah, this is interesting:

|    | index                          | nDCG@10 | Judged@10 | MAP    | Recall@1k | Judged@1k |
|---:|:--------------------|--------:|----------:|-------:|----------:|----------:|
|  1 | old qry+ques           | 0.4921  | 1.0000    | 0.4088 | 0.8582    | 1.0000    |
|  2 | old UDel query      | 0.5616  | 1.0000    | 0.4355 | 0.8571    | 1.0000    |
|  3 | new qry+ques           | 0.4547  | 1.0000    | 0.4148 | 0.9080    | 1.0000    |
|  4 | new UDel query     | 0.5120  | 1.0000    | 0.4499 | 0.9082    | 1.0000    |

These are the results if we remove _all_ unjudged documents.

It seems that the old approach is good at early precision (high ranks), but the new approach gives better recall. Once again, this is a fair comparison because we dedup in both cases all topics have 1k hits.

rodrigonogueira4:
> hey @rodrigonogueira4 remember when we had a discussion about exactly how to index paragraphs?
> 
> The old (existing) implement was:
> 
> * docid: title + abstract
> * docid.00001: title + abstract + 1st paragraph
> * docid.00002: title + abstract + 2nd paragraph
> * docid.00003: title + abstract + 3rd paragraph
> * ...
> 
> But an equally reasonable alternative would be to _not_ repeat the abstract:
> 
> * docid: title + abstract
> * docid.00001: title + 1st paragraph
> * docid.00002: title + 2nd paragraph
> * docid.00003: title + 3rd paragraph
> * ...
> 
> I ran an experiment comparing the old and the new approaches, and the tl;dr is that the old approach is more effective:
> 
> index	nDCG@10	Judged@10	MAP	Recall@1k	Judged@1k
> 1	old qry+ques	0.4738	0.9457	0.2330	0.6332	0.2459
> 2	old UDel	0.5508	0.9686	0.2572	0.6466	0.2411
> 3	new qry+ques	0.4161	0.8800	0.2045	0.6190	0.2353
> 4	new UDel	0.4865	0.8943	0.2273	0.6264	0.2307
> I first ran retrieval to 50k hits on 5/19 index and then filtered down to only `cord_uid`s in the 5/1 corpus, and then evaluated on the union of round 1 and round 2 qrels. Each topic has 1k hits, so the comparison is fair.
> 
> For NDCG@10, new index is a lot lower, but judged is lower also. At depth 1k, judged is comparable, but MAP and recall for new is lower.
> 
> So I think we can conclude that the old indexing scheme is still better?

yes, I think so! Great to have these ablation experiments, btw!

PR #1277, opened 2020-06-14
Refactor and organize filter_run.py and filter_run_with_qrels.py

lintool:
`filter_run.py` - keep docids that are on a "whitelist"

`filter_run_with_qrels.py` - either keep or throw away docs in qrels. Replaces `remove_judged_docids.py`

There's a bit of duplication in these scripts. The tradeoff is a single script with more complex logic (hence more potential for bugs) or separate scripts. Going with separate scripts for now.


src/main/python/trec-covid/filter_run_with_qrels.py
```diff
@@ -0,0 +1,95 @@
...
+                docid = cols[2]
+                score = float(cols[4])
+                tag = cols[5]
+
+                if args.runtag:
+                    tag = args.runtag
+
+                if counts[qid] >= args.k:
+                    if check_score:
+                        if score == prev_score:
```
rodrigonogueira4, line 73: replace `==` with `<=`?

src/main/python/trec-covid/filter_run_with_qrels.py
```diff
@@ -0,0 +1,95 @@
...
+    parser.add_argument('--output', type=str, metavar='run', help='Output run.', required=True)
+    parser.add_argument('--runtag', type=str, default=None, metavar='runtag', help='Runtag.')
+    parser.add_argument('--k', type=int, metavar='hits', default=1000, help='Number of hits to retain per topic.')
+    parser.add_argument('--retain',  action='store_true',
+                        help="Retain judged docids, i.e., throw away all unjudged documents.")
+    parser.add_argument('--discard',  action='store_true',
+                        help="Discard judged docids, i.e., keep only documents that are unjudged.")
+    args = parser.parse_args()
+
+    if (args.retain and args.discard) or (not args.retain and not args.discard):
```
rodrigonogueira4, line 50: not a big thing, but `assert` is preferred as it does not need `sys.exit()`

src/main/python/trec-covid/filter_run_with_qrels.py
lintool: Got it!

src/main/python/trec-covid/filter_run_with_qrels.py
lintool: Since this is an actual command-line program, `assert` will throw stack trace which I think is actually more confusing.

lintool: Will change to just `return`,

PR #1282, opened 2020-06-16
Bind anserini-tools submodule to tools/

lintool:
Also:

+ Removed `src/main/python/msmarco/`, since the scripts are in `tools` already.
+ Updated MS MARCO docs.


src/main/java/io/anserini/search/SearchMsmarco.java
```diff
@@ -35,13 +35,12 @@
...
 /**
- * Class that performs retrieval for the MS MARCO passage ranking task. This is the Java version of the Python script
- * <code>src/main/python/msmarco/retrieve.py</code>.
+ * Class that performs retrieval for the MS MARCO passage ranking task.
  */
 public class SearchMsmarco {
   public static class Args {
     // required arguments
-    @Option(name = "-qid_queries", metaVar = "[file]", required = true, usage="query id - query mapping file")
+    @Option(name = "-queries", metaVar = "[file]", required = true, usage="Queries file.")
```
ronakice, line 43:
```suggestion
    @Option(name = "-queries", metaVar = "[file]", required = true, usage="queries file.")
```
Maybe?

lintool: https://github.com/lintool/guide/blob/master/coding-style.md#describing-arguments-in-command-line-applications

I'm going by `docstring` conventions in Python, which want initial caps... seems reasonable to be consistent?

ronakice: I was just determining by the other arguments in the same file, in which case those need to be changed?

lintool: Yes, good point. Will make it consistent with the Python version also:
https://github.com/castorini/anserini-tools/blob/master/scripts/msmarco/retrieve.py

PR #1283, opened 2020-06-16
add option to output external document ids in ExtractDocumentLengths

nsndimt:
The current implementation only outputs the internal document ids. If users want to get the external document ids, they need to use Pyserini to look up each id one-by-one. Adding the option will save users the trouble.


lintool:
I think it's fine if we output the external docids by default also... so no need for an option.

Can you improve the test case accordingly also? Thanks!

src/main/java/io/anserini/util/ExtractDocumentLengths.java
```diff
@@ -90,7 +92,8 @@ public static void main(String[] args) throws Exception {
       // See https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/similarities/BM25Similarity.java
       int lossyDoclength = SmallFloat.byte4ToInt(SmallFloat.intToByte4((int) exactDoclength));
       int lossyTermCount = SmallFloat.byte4ToInt(SmallFloat.intToByte4((int) exactTermCount));
-      out.println(String.format("%d\t%d\t%d\t%d\t%d", i, exactDoclength, exactTermCount, lossyDoclength, lossyTermCount));
+      out.println(String.format("%s\t%d\t%d\t%d\t%d", IndexReaderUtils.convertLuceneDocidToDocid(reader, i),
```
lintool, line 95: So sorry about the delayed response... can we keep _both_ the internal id and the collection docid?

nsndimt: finished

PR #1285, opened 2020-06-17
fix WET index contents missing last character

MXueguang:
The warc header in WET format doesn't count the empty line that follows the header to Content-Length like what WARC format does. Overwritten `readNextRecord` to fix this.


src/main/java/io/anserini/collection/CommonCrawlWetCollection.java
```diff
@@ -106,6 +108,96 @@ public static Document readNextWarcRecord(DataInputStream in)
...
+      if (in == null || headerBuffer == null) {
+        throw new NoSuchElementException();
+      }
+
+      String line = null;
+      boolean foundMark = false;
+      boolean inHeader = true;
+      byte[] retContent = null;
+
+      // cannot be using a buffered reader here!!!!
```
lintool, line 129: Can you clarify these comments and below?

MXueguang:
> Should we improve the test case to catch this issue?

This issue is covered in test case. old test document is written manually, so it missed this issue. now it fixed.

src/main/java/io/anserini/collection/CommonCrawlWetCollection.java
MXueguang: I'll remove the comment about buffered reader here. It should belong to `readLineFromInputStream` and was well described in that function.

PR #1290, opened 2020-06-19
Add capabilities to parse "CC-News-En" corpus

JMMackenzie:
We have a forthcoming corpus named "CC-News-En" which is a static capture of a chunk of CC-News articles. It adds a `WARC-TREC-ID` to each document.

This PR allows Anserini to index this corpus. Some documentation is currently missing and will be updated once the corpus is ready for public use.


lintool:
Thanks for the contribution!

Issue #1291, opened 2020-06-19
Exception when Building WaPo v3 Index


PepijnBoers:
I'm building the index for the 3th version of the Washington Post collection, which adds 154,418 new documents from 2018 and 2019. However, I encountered an error while doing so:

```
ERROR [pool-2-thread-1] index.IndexCollection$LocalIndexerThread (IndexCollection.java:241) - pool-2-thread-1: Unexpected Exception:
java.lang.RuntimeException: com.fasterxml.jackson.databind.exc.MismatchedInputException: Missing required creator property ‘published_date’ (index 3)
```

Indicating that the specific article **did not possess a publication date**. The only date available is in the `contents` section as: `{"content": 1514865008000, "mime": "text/plain", "type": "date"}`, the  usual `"published_date": 1325376562000` outside `contents` is missing.

This is the case for at least 2 articles, but I suspect there to be more.

I used the following command line option to build the Index:
```
./target/appassembler/bin/IndexCollection -collection WashingtonPostCollection \
 -input /Volumes/Samsung_T5/WashingtonPost.v3/data/ -generator WashingtonPostGenerator \
 -index lucene-index.core18.pos+docvectors+rawdocs_all+contents \
 -threads 1 -storePositions -storeDocvectors -storeRaw -storeContents -optimize 
```

ChanikanWitt:
Hi, I am wondering whether this issue is solved yet. I try building index in WashingtonPost v3 and still had the same issue.

PepijnBoers:
Hi @ChanikanWitt this issue has been resolved on June 21 and was included in the v0.9.4 release. Is the build you're using up to date?

ChanikanWitt:
Oh okay. It works now. Thank you!

PR #1295, opened 2020-06-21
Fix WaPo v3 indexing

PepijnBoers:
This PR resolves https://github.com/castorini/anserini/issues/1291 by making the `published_date` attribute optional for Washington Post articles during indexing. We do, however, attempt to obtain the publication date from the `contents` section when no dedicated `published_date` is available.


lintool:
hey @PepijnBoers does this work for v2 also? We should probably maintain backwards compatibility?

src/main/java/io/anserini/collection/WashingtonPostCollection.java
```diff
@@ -48,8 +48,8 @@
 
 /**
  * An instance of the <a href="https://trec.nist.gov/data/wapost/">TREC Washington Post Corpus</a>.
- * The collection contains 608,180 news articles and blog posts from January 2012 through August 2017,
```
lintool, line 51: Should we reference stats for v2 and v3 separately? I believe v1 is defunct, right?

PepijnBoers: We could mention that version 3 adds 154,418 new documents from 2018 and 2019 and removes (near) duplicates.

v2 was used in TREC News 2018 & 2019 and is basically v1 without exact duplicates.

PepijnBoers:
This still works for v2 because all articles there have an explicit `published_date` attribute. The script always uses the attribute whenever it is available and only uses the date in `contents` if that is the only one present.

I have also built the v2 index and checked if I obtained the same scores for the `BM25+RM3` run for TREC 2019 bg-linking as were mentioned [here](https://github.com/castorini/anserini/blob/master/docs/regressions-backgroundlinking19.md).

lintool:
great, thanks! I'll confirm regressions on my end and then we'll get this merged!

lintool:
These regressions confirmed:

```
python src/main/python/run_regression.py --collection core18 --index
python src/main/python/run_regression.py --collection backgroundlinking18
python src/main/python/run_regression.py --collection backgroundlinking19
```

src/main/java/io/anserini/collection/WashingtonPostCollection.java
lintool: Yes, I think we should explicitly describe the date ranges and num of documents of v2 and v3.

PR #1296, opened 2020-06-21
Upgrade to Lucene 8.5.2

### LINKED ISSUE #1292: Possible Geospatial indexing issue in Lucene 8.3.0
In the updated version of Lucene, a Triangle ShapeField `org.apache.lucene.document.ShapeField.Triangle` is the field used for indexing polygons, however this appears to be unavailable in anserini's current version of lucene. This may also be due to the fact that although anserini uses lucene-core 8.3, the lucene-sandbox version is still 8.0 (geo-indexing feature is always contained in lucene-sandbox from 8.0-8.3 due to it being experimental). The issue may be resolved by simply updating lucene-sandbox to be 8.3 or higher.

lintool:
Would close #1292

I'm in the process or running all regressions to make sure nothing has changed. Also, index format has changed, so I think we need to rebuild indexes.

@ronakice @rodrigonogueira4 We should update what we have here, right?
https://git.uwaterloo.ca/jimmylin/anserini-indexes

Also, I'll build a doc index while we're at it?


ronakice:
SGTM!

lintool:
Outdated... closing.

Issue #1289, opened 2020-06-18
Add documentation for FEVER baseline


lintool:
In the FEVER paper: https://www.aclweb.org/anthology/N18-1074/

Table 2:
<img width="305" alt="Screen Shot 2020-06-17 at 7 33 50 AM" src="https://user-images.githubusercontent.com/313837/85028237-d060fc00-b148-11ea-8f5d-c8ff44d1fdc5.png">

We should add documentation that compares Anserini baselines directly to results in this table.

cc/ @ronakice @kelvin-jiang

ronakice:
Yup, @kelvin-jiang is on it!

ronakice:
Close?

PR #1297, opened 2020-06-23
Add documentation for FEVER baseline

kelvin-jiang:
Addresses #1289.

This PR adds the documentation and scripts needed to replicate and evaluate FEVER document retrieval using anserini, which after some BM25 tuning, outperforms the TF-IDF baseline in the paper at every value of k 🎉! Results are listed in the docs.


docs/experiments-fever.md
```diff
@@ -0,0 +1,196 @@
...
+wget https://s3-eu-west-1.amazonaws.com/fever.public/paper_dev.jsonl -P collections/fever
+```
+
+To confirm, `wiki-pages.zip` should have MD5 checksum of `ed8bfd894a2c47045dca61f0c8dc4c07`.
+
+## Building Lucene Indexes
+
+Next, we want to index the Wikipedia dump (`wiki-pages.zip`) using Anserini. Note that this Wikipedia dump consists of Wikipedia articles' introductions only, which we will refer to as "paragraphs" from this point onward.
+
+We currently have implemented two variants: (1) indexing paragraphs and (2) indexing sentences within each paragraph separately.
```
ronakice, line 28:
```suggestion
We will consider two variants: (1) Paragraph Indexing and (2) Sentence Indexing.
```

docs/experiments-fever.md
```diff
@@ -0,0 +1,196 @@
...
+ -collection FeverSentenceCollection -generator DefaultLuceneDocumentGenerator \
+ -threads 9 -input collections/fever/wiki-pages \
+ -index indexes/fever/lucene-index-fever-sentence -storePositions -storeDocvectors -storeRaw 
+```
+
+Upon completion, we should have an index with 25,247,887 documents (sentences).
+
+## Performing Retrieval on the Dev Queries
+
+(We will use the paragraph index for this section, though the steps could be easily modified for the sentence index.)
```
ronakice, line 58:
```suggestion
Note that while we use paragraph indexing for this section, these steps can easily be modified for sentence indexing.
```

docs/experiments-fever.md
```diff
@@ -0,0 +1,196 @@
...
+| k   | Fully Supported | Oracle Accuracy |
+|:----|----------------:|----------------:|
+| 1	  | 0.3856885689    | 0.5904590459    |
+| 5	  | 0.6366636664    | 0.7577757776    |
+| 10  | 0.7193219322    | 0.8128812881    |
+| 25  | 0.8003300330    | 0.8668866887    |
+| 50  | 0.8472847285    | 0.8981898190    |
+| 100 | 0.8804380438    | 0.9202920292    |
+
+which outperforms the TF-IDF baseline in the FEVER paper at almost every value of k.
```
ronakice, line 196:
```suggestion
which outperforms the TF-IDF baseline in the FEVER paper at every tested value of k.
```

docs/experiments-fever.md
```diff
@@ -0,0 +1,196 @@
...
+```bash
+python src/main/python/fever/evaluate_doc_retrieval.py \
+ --truth_file collections/fever/paper_dev.jsonl \
+ --run_file runs/run.fever-paragraph-0.6-0.5.dev.tsv
+```
+
+then we can achieve the following results:
+
+| k   | Fully Supported | Oracle Accuracy |
+|:----|----------------:|----------------:|
```
ronakice, line 188: Can we fix this at the same sig figs as the table in the paper (modify the code to do that too?). Table will look a little neater.

src/main/python/fever/tune_bm25.py
```diff
@@ -0,0 +1,86 @@
...
+        # maximize R@1000
+        if recall > max_recall:
+            max_recall = recall
+            max_file = file
+
+    print(f'Best parameters: {max_file} (R@1000 = {max_recall})')
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Tunes BM25 parameters for FEVER document retrieval.')
+    parser.add_argument('--runs_folder', required=True, help='Directory for storing runs. Do not use runs/.')
```
ronakice, line 74: Why this warning?

src/main/python/fever/generate_subset.py
```diff
@@ -0,0 +1,58 @@
+#
```
ronakice, line 1: Sorry, what's the use of this file?

docs/experiments-fever.md
kelvin-jiang: Unfortunately we don't seem to outperform TF-IDF at k=50?

kelvin-jiang: Oh nvm, looks like their reported oracle accuracy is wrong (assuming their fully supported % is right)

docs/experiments-fever.md
kelvin-jiang: My concern for modifying the code to do that is that the tabbed printing will get messed up, but I guess that's not too important lol

docs/experiments-fever.md
ronakice: yeah, I just looked at the first column since the second is determined exclusively by it.

src/main/python/fever/tune_bm25.py
kelvin-jiang: The script iterates through all run files in `runs_folder`, if you pass it `runs` then it will evaluate all run files found in the directory (even non-FEVER ones).

lintool:
> Thanks for adding this! I was wondering if we should move these python scripts to anserini-tools? @lintool, I'm not sure what we want to still keep here.

I think we can get it checked in first and then move around.

Related - I was wondering if we can switch away from the MS MARCO custom `retrieve.py` script? It really is supposed to a one-off. I'd rather use `SearchCollection` and use the standard TREC run format. There's a TSV topic reader, so it shouldn't be too difficult a switch.

I'm fine with an early merge, and then other PRs to fix.

src/main/python/fever/generate_subset.py
kelvin-jiang: It generates the training data subset that I used to tune BM25 params on. Intent was to create a subset with balanced labels.

src/main/python/fever/tune_bm25.py
ronakice: Cool

PR #1300, opened 2020-06-24
Update CORD-19 auto index script to use TREC-COVID Round 3 data

lintool:


rodrigonogueira4:
A side note: do you think we should start using qrels 1+2+3? They will likely give more reliable results than only individual qrels

lintool:
hey @rodrigonogueira4 we should be using https://github.com/castorini/anserini/blob/master/src/main/resources/topics-and-qrels/qrels.covid-round3-cumulative.txt for rounds 1, 2, 3 I think!

rodrigonogueira4:
> hey @rodrigonogueira4 we should be using https://github.com/castorini/anserini/blob/master/src/main/resources/topics-and-qrels/qrels.covid-round3-cumulative.txt for rounds 1, 2, 3 I think!

Yes, thanks!

PR #1316, opened 2020-07-11
Iso19115 indexer, indexed more fields

shaneding:


src/main/java/io/anserini/collection/Iso19115Collection.java
```diff
@@ -113,6 +120,20 @@ public Document(JsonNode json) {
...
+      this.recommendedCitation = json.get("gmd:MD_Metadata").get("gmd:identificationInfo").get("gmd:MD_DataIdentification").get("gmd:citation")
+                                 .get("gmd:CI_Citation").get("gmd:otherCitationDetails").get("gco:CharacterString").asText();
+      this.thesaurusName = json.get("gmd:MD_Metadata").get("gmd:identificationInfo").get("gmd:MD_DataIdentification").get("gmd:descriptiveKeywords")
+                           .get(0).get("gmd:MD_Keywords").get("gmd:thesaurusName").get("gmd:CI_Citation").get("gmd:title").get("gco:CharacterString").asText()
+                           + " : " +
+                           json.get("gmd:MD_Metadata").get("gmd:identificationInfo").get("gmd:MD_DataIdentification").get("gmd:descriptiveKeywords")
+                           .get(0).get("gmd:MD_Keywords").get("gmd:thesaurusName").get("gmd:CI_Citation").get("gmd:otherCitationDetails").get("gco:CharacterString")
+                           .asText();
 
       // extracting all the authors of the paper
```
lintool, line 138: Comment needs to be fixed?

src/main/java/io/anserini/collection/Iso19115Collection.java
```diff
@@ -113,6 +120,20 @@ public Document(JsonNode json) {
               .get("gco:CharacterString").asText();
       this.publish_time = json.get("gmd:MD_Metadata").get("gmd:dateStamp").get("gco:Date").asText();
       this.url = json.get("gmd:MD_Metadata").get("gmd:dataSetURI").get("gco:CharacterString").asText();
+      this.purpose = json.get("gmd:MD_Metadata").get("gmd:identificationInfo").get("gmd:MD_DataIdentification").get("gmd:purpose")
+                     .get("gco:CharacterString").asText();
+      this.supplInfo = json.get("gmd:MD_Metadata").get("gmd:identificationInfo").get("gmd:MD_DataIdentification").get("gmd:supplementalInformation")
```
lintool, line 125: Would it make sense to change `gmd:MD_Metadata` and other keywords into `static final String` so we don't get hard-to-debug typos?

lintool: At least for the "keys" that are used multiple times.

src/main/java/io/anserini/collection/Iso19115Collection.java
```diff
@@ -132,8 +164,8 @@ public Document(JsonNode json) {
               .get("gmd:geographicElement").get("gmd:EX_GeographicBoundingBox").get("gmd:southBoundLatitude").get("gco:Decimal").asDouble();
       // ensuring that a single coordinate location will be drawn as a small rectangle
       if (latitude[0] == latitude[2]) {
-        latitude[0] -= 0.00001;
-        latitude[2] += 0.00001;
+        latitude[0] -= 0.01;
```
lintool, line 167: I think this should be done in the UI - I don't think we should be altering the records?

src/main/java/io/anserini/collection/Iso19115Collection.java
```diff
@@ -204,9 +258,9 @@ private String getCoordinateString() {
       return coordinates.toString();
     }
 
-    public double[] getLatitude() { return latitude; }
+    // public double[] getLatitude() { return latitude; }
 
-    public double[] getLongitude() { return longitude; }
+    // public double[] getLongitude() { return longitude; }
```
lintool, line 263: Remove dead code.

src/main/java/io/anserini/collection/Iso19115Collection.java
shaneding: because we are putting longitude/latitude in a string, it would be a lot more painful to do this on the frontend, I don't think we are altering the records in this case, just converting a single point to a small rectangle.

src/main/java/io/anserini/collection/Iso19115Collection.java
shaneding: Idts, because there are a lot of these possible keys and it can get really messy even if its just for keys that appear multiple times.

shaneding: In terms of hard-to-debug typos, java threw null pointer exceptions with the line number whenever a key is incorrect so debugging shouldnt be too difficult

lintool: ok

src/main/java/io/anserini/collection/Iso19115Collection.java
lintool: oh, storing in a string a bit janky, but given that, yes, I agree.

PR #1320, opened 2020-07-17
Replication results with Solr/ElasticSearch

LizzyZhang-tutu:
Replicated results for msmarco-doc and msmarco-passage using Elastic Search.
image
image

image
image

**Environment**
OS:
 macOS Catalina Version 10.15.5 
Java: 
openjdk version "11.0.7" 2020-04-14
 OpenJDK Runtime Environment Adopt
OpenJDK (build 11.0.7+10) OpenJDK 64-Bit Server VM Adopt
OpenJDK (build 11.0.7+10, mixed mode) 
Python: 
Python 3.7.4

**Issue Encountered**
I was not able to replicate result for robust04, core18 since it seems we need to submit organizational/individual agreement forms to download  TREC Disks 4 and 5 or TREC Washington Post Corpus. But I think it’s worth mentioning that the Solr indexing command for robust04 on solrini.md still has JsoupGenerator, which does not exist in the repo, should be DefaultLuceneDocumentGenerator I think?

Also, I was not able to run the end-to-end integration test with run_es_regression.py.
I think it's because the command line for evaluation is
https://github.com/castorini/anserini/blob/master/src/main/python/run_solr_regression.py#L151, but in my case the eval directory is inside tools directory.


lintool:
Closing, superseded by #1322

PR #1323, opened 2020-07-20
Fixes parameter name

rodrigonogueira4:


rodrigonogueira4:
> I actually like `-queries` better... can we change the code?

Despite also liking `-queries` better, I prefer to change parameter names only when necessary as it breaks other people's scripts, including mine. We were using `-queries` in the past, and for some reason, we changed it `qid_queries`.

lintool:
okay, reasonable decision. go ahead and merge yourself?

PR #1332, opened 2020-07-23
Add TREC COVID round 5 baselines

lintool:
Also added Java bindings.

Can someone take a look and actually run the commands? Thanks!


lintool:
Hi @lukuang - thanks!

Correct, those tied score warnings have been there for a while now... I think it's okay to ignore for now.

PR #1337, opened 2020-07-27
expose document level term positions

nsndimt:
Currently, [get_document_vector](https://github.com/nsndimt/anserini/blob/6b2bd2cd963a476df600f3bf37f3ba4f7b281d25/src/main/java/io/anserini/index/IndexReaderUtils.java#L364) only return terms and corresponding frequencies. This pull request implements a new function called getPositionList in which returns terms and, for each term,  every position this term occurs in the document.

### DIFF
--- a/src/main/java/io/anserini/index/IndexReaderUtils.java
+++ b/src/main/java/io/anserini/index/IndexReaderUtils.java
@@ -394,7 +394,7 @@ public class IndexReaderUtils {
    * @throws IOException if error encountered during query
    * @throws NotStoredException if the term vector is not stored
    */
-  public static Map<String, List<Long>> getTermPositions(IndexReader reader, String docid) throws IOException, NotStoredException {
+  public static Map<String, List<Integer>> getTermPositions(IndexReader reader, String docid) throws IOException, NotStoredException {
     int ldocid = convertDocidToLuceneDocid(reader, docid);
     if (ldocid == -1) {
       return null;
@@ -408,16 +408,16 @@ public class IndexReaderUtils {
       throw new NotStoredException("Document vector not stored!");
     }
 
-    Map<String, List<Long>> termPosition = new HashMap<>();
+    Map<String, List<Integer>> termPosition = new HashMap<>();
     PostingsEnum positionIter = null;
 
     while ((termIter.next()) != null) {
-      List<Long> positions = new ArrayList<>();
+      List<Integer> positions = new ArrayList<>();
       Long termFreq = termIter.totalTermFreq();
       positionIter = termIter.postings(positionIter, PostingsEnum.POSITIONS);
       positionIter.nextDoc();
       for ( int i = 0; i < termFreq; i++ ) {
-        positions.add(Long.valueOf(positionIter.nextPosition()));
+        positions.add(positionIter.nextPosition());
       }
       termPosition.put(termIter.term().utf8ToString(), positions);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -383,6 +383,48 @@ public IndexTerm next() {
...
+   * {@code null} if the document does not exist (as opposed to an empty map), so that the caller is explicitly forced
+   * to handle this case.
+   *
+   * @param reader index reader
+   * @param docid collection docid
+   * @return term position mapping for a particular document or {@code null} if document does not exist.
+   * @throws IOException if error encountered during query
+   * @throws NotStoredException if the term vector is not stored
+   */
+  public static Map<String, List<Long>> getPositionList(IndexReader reader, String docid) throws IOException, NotStoredException {
```
lintool, line 397: How about `getTermPositions`?

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -383,6 +383,48 @@ public IndexTerm next() {
...
+    if (terms == null) {
+      throw new NotStoredException("Document vector not stored!");
+    }
+    TermsEnum te = terms.iterator();
+    if (te == null) {
+      throw new NotStoredException("Document vector not stored!");
+    }
+
+    Map<String, List<Long>> term_position = new HashMap<>();
+    PostingsEnum position_iter = null;
```
lintool, line 412: variables should be camel case?

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -383,6 +383,48 @@ public IndexTerm next() {
...
+
+    while ((te.next()) != null) {
+      List<Long> positions = new ArrayList<>();
+      Long freq = te.totalTermFreq();
+      position_iter = te.postings(position_iter, PostingsEnum.POSITIONS);
+      position_iter.nextDoc();
+      for ( int i = 0; i < freq; i++ ) {
+        positions.add(Long.valueOf(position_iter.nextPosition()));
+      }
+      term_position.put(te.term().utf8ToString(), positions);
```
lintool, line 422: camel case.

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -383,6 +383,48 @@ public IndexTerm next() {
...
+    Terms terms = reader.getTermVector(ldocid, IndexArgs.CONTENTS);
+    if (terms == null) {
+      throw new NotStoredException("Document vector not stored!");
+    }
+    TermsEnum termIter = terms.iterator();
+    if (termIter == null) {
+      throw new NotStoredException("Document vector not stored!");
+    }
+
+    Map<String, List<Long>> termPosition = new HashMap<>();
```
lintool, line 411: Just noticed this - do we need to use long? I don't think documents are going to be that looong? How about just `List<Integer>`?

lintool:
Another comment for you...

src/main/java/io/anserini/index/IndexReaderUtils.java
nsndimt: Yes. `position_iter.nextPosition()` return int. Using long is unnecessary.  I have changed the declaration of `termPostion` accordingly.
I also discovered that `Long termFreq = termIter.totalTermFreq();` introduce autoboxing. I have changed `Long` to `long` to avoid this.

lintool: yes, autoboxing is inefficient, but since you have `List<Integer>`, you'll have to convert it back into an object anyway...

PR #1338, opened 2020-07-28
doc2query baselines for round4/round5

ronakice:
Run links need to be updated but the rest should be fine for now!


docs/experiments-covid-doc2query.md
```diff
@@ -0,0 +1,87 @@
+# TREC-COVID Doc2Query Baselines
```
lintool, line 1: I think @rodrigonogueira4 and I have been consistent with doc2query?

lintool:
One more thing - add link in README also?

PR #1343, opened 2020-07-30
Estella

estella98:
Environment

OS:
macOS Mojave Version 10.14.6
Java:
openjdk 11.0.7 2020-04-14
OpenJDK Runtime Environment (build 11.0.7+10)
OpenJDK 64-Bit Server VM (build 11.0.7+10, mixed mode)
Python:
Python 3.7.7
No issues encountered, everything works as expected.


estella98:
Hi Prof Jimmy:

Thank you for approving my PR and sorry for the delay. I want to mention that for the first exercise, the branch that I made the PR out of is not where I completed the exercise. I completed the exercise in my local directory of the original Anserini project repo, but made the pr later from my own forked version of Anserini. But other than that, I was able to reproduce the results and got the exact same answer listed on the page.
However, I encountered the same problem that Lizzy got when running nohup sh target/appassembler/bin/IndexCollection -collection CleanTrecCollection
-generator DefaultLuceneDocumentGenerator -threads 1 -input collections/msmarco-doc
-index indexes/msmarco-doc/lucene-index.msmarco-doc.pos+docvectors+rawdocs
-storePositions -storeDocvectors -storeRaw >& logs/log.msmarco-doc.pos+docvectors+rawdocs &.   in the second exercise.

My Terminal shows [1] 86349, and if I ignored it and continued to do the trec_eval evaluation,?I would get "
map  all   0.0136, recall_1000  all   0.0189"  which does not match the result listed.
Should I still add my name to the replication log in this case?
Thank you!

Yours,
Estella Liu

________________________________

lintool:
> map  all   0.0136, recall_1000  all   0.0189

This doesn't sound right... is this for MS MARCO?

estella98:
yes it is. I wonder is it because I failed to run the previous indexing step? Because according to the documentation that process should take around 40 mins but I got  [1] 86349 immediately, so I don't think the documents are indexed properly.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

lintool:
`[1] 86349` is the process id when you put a job in the background. You'll need to wait for the job to finish... check by `tail logs/log.msmarco-doc.pos+docvectors+rawdocs`

PR #1345, opened 2020-08-02
Issue1113: Simplify Base Solr Config

LizzyZhang-tutu:
Simplify Solr schema config by removing non-English text field types and corresponding dynamic fields in managed-schema. 
Also removing corresponding stopwords files in lang.zip.


LizzyZhang-tutu:
Note that **stopwords.txt**(https://github.com/castorini/anserini/blob/master/src/main/resources/solr/anserini/conf/stopwords.txt) is currently empty and is used for fieldTypes: text_gen_sort, text_general, text_genera_rev.
There’s also a **stopwords_en.txt** in lang.zip which is used for text_en with some contents. I think maybe we can get rid of the **stopwords.txt** and the three corresponding fieldTypes also?

lintool:
@LizzyZhang-tutu Agreed.

Is there a reason conf is checked in as a zip? Can we replace with plain-text files so we can keep track of diffs?

LizzyZhang-tutu:
I kept lang.zip as it was, but yes I think since we are only having one language now, we can replace it with a plain-text file - stopwords_en.txt. I also removed stopwords.txt and corresponding fieldTypes and dynamic fields.

src/main/resources/solr/anserini/conf/stopwords_en.txt
```diff
@@ -12,3 +12,43 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# a couple of test stopwords to test that the words are really being
```
lintool, line 16: should we remove these couple of lines?

add a comment at the beginning saying this is Lucene's default?

lintool:
Can you also confirm that the Solrini tests still work okay?

src/main/resources/solr/anserini/conf/stopwords_en.txt
LizzyZhang-tutu: sure I'll remove these lines.
It thought the line after is already stating that this is Lucene's default? -
https://github.com/castorini/anserini/pull/1345/files/5fff9e78959a7a578f12d8d816c1f7052ca4228c#diff-22892be18d725ce8edba42553448193fR21

LizzyZhang-tutu:
We'll need to update solr.sh https://github.com/castorini/anserini/blob/4fb0d3f4f6df0e7a57a5e04a7be8b9b5bfe08cb3/src/main/resources/solr/solr.sh#L21-L23 since we are removing lang.zip.

LizzyZhang-tutu:
Other than that, I was able to replicate results here https://github.com/castorini/anserini/blob/master/docs/experiments-cord19-extras.md

lintool:
Yes, please update `anserini/src/main/resources/solr/solr.sh` as well?

Please make sure these work also: https://github.com/castorini/anserini/blob/master/docs/solrini.md

LizzyZhang-tutu:
Sure - I'll update and confirm the integration tests work. 
Also, if these work, do we want to make the same updates for anserini-twitter/conf? Now we are just modifying anserini/conf.

LizzyZhang-tutu:
I can confirm that the integration tests (https://github.com/castorini/anserini/blob/master/docs/solrini.md) work for robust04, core18, msmarco-passage.

One issue I encountered 
image,
so basically, I needed to change the command lines for evaluation https://github.com/castorini/anserini/blob/master/src/main/python/run_solr_regression.py#L141-L152 to add dir tools/:
image

@lintool Should I update src/main/python/run_solr_regression.py as well?

lintool:
> Sure - I'll update and confirm the integration tests work.
> Also, if these work, do we want to make the same updates for anserini-twitter/conf? Now we are just modifying anserini/conf.

Yes please.

lintool:
> I can confirm that the integration tests (https://github.com/castorini/anserini/blob/master/docs/solrini.md) work for robust04, core18, msmarco-passage.
> 
> One issue I encountered
> image,
> so basically, I needed to change the command lines for evaluation https://github.com/castorini/anserini/blob/master/src/main/python/run_solr_regression.py#L141-L152 to add dir tools/:
> image
> 
> @lintool Should I update src/main/python/run_solr_regression.py as well?

I don't think this is necessary. Building `trec_eval` is already in the instructions for "Getting Started":
https://github.com/castorini/anserini#getting-started

LizzyZhang-tutu:
> I don't think this is necessary. Building `trec_eval` is already in the instructions for "Getting Started":
> https://github.com/castorini/anserini#getting-started

Yeah I followed the steps to build trec_eval as well:
`cd tools/eval && tar xvfz trec_eval.9.0.4.tar.gz && cd trec_eval.9.0.4 && make && cd ../../..`
`cd tools/eval/ndeval && make && cd ../../..`

That's why I had the problem I think, so the path is actually 
`tools/eval/trec_eval.9.0.4/trec_eval`, not 
`eval/trec_eval.9.0.4/trec_eval`(in https://github.com/castorini/anserini/blob/master/src/main/python/run_solr_regression.py#L141-L152) right?

LizzyZhang-tutu:
Updated the PR by simplifying Solr config for anserini-twitter as well.

lintool:
> That's why I had the problem I think, so the path is actually
> `tools/eval/trec_eval.9.0.4/trec_eval`, not
> `eval/trec_eval.9.0.4/trec_eval`(in https://github.com/castorini/anserini/blob/master/src/main/python/run_solr_regression.py#L141-L152) right?

Yes, you're right. Can you fix `run_solr_regression.py` also? Thanks!

@yxzhu16 do you mind taking a look at this and seeing if you can replicate on this branch? Thanks!

LizzyZhang-tutu:
> Yes, you're right. Can you fix `run_solr_regression.py` also? Thanks!

Sure. Updated the pr with the path fix.

yxzhu16:
@lintool 
Can confirm that Solrini tests work. 
One concern is, in many of the .md files `tools/eval/trec_eval.9.0.4/trec_eval` is actually written as `eval/trec_eval.9.0.4/trec_eval`, like in https://github.com/castorini/anserini/blob/master/docs/solrini.md.

lintool:
Merging, but opened #1352

PR #1348, opened 2020-08-05
Refactor scripts for generating TREC-COVID baselines (rounds 3, 4, 5)

lintool:
Runs remain exactly the same - just refactoring to reduce code duplication.

### DIFF
--- a/src/main/python/trec-covid/covid_baseline_tools.py
+++ b/src/main/python/trec-covid/covid_baseline_tools.py
@@ -24,6 +24,102 @@ sys.path.insert(0, '../pyserini/')
 import pyserini.util
 
 
+def perform_runs(round_number, indexes):
+    base_topics = f'src/main/resources/topics-and-qrels/topics.covid-round{round_number}.xml'
+    udel_topics = f'src/main/resources/topics-and-qrels/topics.covid-round{round_number}-udel.xml'
+
+    # Use cumulative qrels from previous round for relevance feedback runs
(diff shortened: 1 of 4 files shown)


src/main/python/trec-covid/covid_baseline_tools.py
```diff
@@ -24,6 +24,146 @@
 import pyserini.util
 
 
+def perform_runs(round_number, indexes):
+    base_topics = f'src/main/resources/topics-and-qrels/topics.covid-round{round_number}.xml'
+    udel_topics = f'src/main/resources/topics-and-qrels/topics.covid-round{round_number}-udel.xml'
+
+    # Use cumulative qrels from previous round for relevance feedback runs
+    cumulative_qrels = f'src/main/resources/topics-and-qrels/qrels.covid-round{round_number - 1}-cumulative.txt'
```
rodrigonogueira4, line 32: will this break if round_number=1?

lintool: Correct. This will only work for rounds 3, 4, 5 - where we've established a consistent set of runs. It'll most definitely break for rounds 1 and 2. I'll need to go back and retro fit those runs to make consistent.

rodrigonogueira4: All right, I will approve it now so it will be up to you if you want to do this in this PR or in another one

PR #1357, opened 2020-08-18
adding url by correcting typo

estella98:


lintool:
hi @estella98 - sorry I didn't carefully check, but your commit doesn't seem to refer to a valid commit id? Should point to the commit point at which you ran the code...

Issue #1359, opened 2020-08-20
Inconsistent argument names for `qid_queries` between doc and code


tangsaidi:
The doc was updated recently in #1323, where the parameter was changed to `qid_queries`

https://github.com/castorini/anserini/blob/aba846aa07d6f319fb3dc9cb591c20b4ae69f9ef/docs/experiments-msmarco-passage.md#L63

However, the argument expected by the MSMarco program is `queries`

https://github.com/castorini/anserini/blob/aba846aa07d6f319fb3dc9cb591c20b4ae69f9ef/src/main/java/io/anserini/search/SearchMsmarco.java#L43

I'm not sure if we should revert the documentation change to conform with the code or simply change the argument name in the code.

resolves #1359

lintool:
Please fix the documentation and send PR?

PR #1360, opened 2020-08-20
Appends on MSMarco Replication Log by @tangsaidi

tangsaidi:
Environment
OS:
Ubuntu 20.04 LTS

Java:
openjdk 11.0.8 2020-07-14
OpenJDK Runtime Environment (build 11.0.8+10-post-Ubuntu-0ubuntu120.04)
OpenJDK 64-Bit Server VM (build 11.0.8+10-post-Ubuntu-0ubuntu120.04, mixed mode, sharing)

Python:
Python 3.7.6

I encountered a minor issue while running the experimentation. I described it in #1359 

Other than that, everything was smooth. I was able to produce identical results as described in the document :100: 

msmarco-passage

image

msmarco-doc

Default parameter
image

Tuned parameter
image

I'm interested in the data engineering/system building aspect of your research. I'll continue to work on the replication of Solrini and Elastirini before signing up for a task.


lintool:
hrm, your commit doesn't seem to refer to a valid commit id? Should point to the commit point at which you ran the code...

Issue #1367, opened 2020-09-05
Failed Solr integration tests caused by SecureRandom algorithm "NativePRNG"


adamyy:
Environment:

- openjdk 11.0.7 2020-04-14
- OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.7+10)
- OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.7+10, mixed mode)

I have observed the following test failures in my local build of master:

```
java.lang.AssertionError: SecureRandom algorithm 'NativePRNG' is in use by your JVM, which is a potentially blocking algorithm on some environments. Please report the details of this failure (and your JVM vendor/version) to solr-user@lucene.apache.org. Y
ou can try to run your tests with -Djava.security.egd=file:/dev/./urandom or bypass this check using -Dtest.solr.allowed.securerandom=NativePRNG as a JVM option when running tests.
```

**TL;DR** if you do experience this issue, please try again using `-Djava.security.egd=file:/dev/./urandom`.

It seems like the Solr test framework is forbidding the use of certain SecureRandom algorithm, namely `NativePRNG` and `NativePRNGBlocking`. The related Solr [thread](https://issues.apache.org/jira/browse/SOLR-10338) suggested that `SHA1PRNG` should be preferred in tests for its nonblocking nature. 

I suspect that this issue is JDK dependent, for instance the TravisCI [build](https://travis-ci.org/github/castorini/anserini/jobs/720180310) wasn't impacted. 

There is a number of ways to alleviate this issue if you do experience it. We can set the system property `java.security.egd` to `file:/dev/./urandom` for all tests, which change the default SecureRandom algorithm to `SHA1PRNG`. Ideally I think the configuration should be set only for Solr tests, still trying to figure out how to do that.

To set the system property for all tests, either supply the option in the command line, or set that in the Maven surefire plugin:

```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-surefire-plugin</artifactId>
  <version>2.12.4</version>
  <configuration>
    <systemPropertyVariables>
      <java.security.egd>file:/dev/./urandom</java.security.egd>
    </systemPropertyVariables>
  </configuration>
</plugin>
```

Full logs:

```
Running io.anserini.integration.solr.TrecEndToEndTest
[...]

PR #1368, opened 2020-09-06
Use standalone EmbeddedSolrServer for Solr integration tests

adamyy:
As described in https://github.com/castorini/anserini/issues/1367, `EmbeddedSolrServerTestBase` enforces specific PRNG algorithm to be used (or not used) by Java `SecureRandom`. This PR uses `EmbeddedSolrServer` by itself in tests to bypass this check.


lintool (approved):

lintool:
This is great, thanks!

PR #1355, opened 2020-08-18
JsoupGenerator -> DefaultLuceneDocumentGenerator in doc

adamyy:
Seems like [`JsoupGenerator`](https://github.com/castorini/anserini/blob/anserini-0.8.1/src/main/java/io/anserini/index/generator/JsoupGenerator.java) was removed in commit [`9a28a0`](https://github.com/castorini/anserini/commit/9a28a098dfd85366be29a6feb385c9e2493f988c) (04/07/2020), but there are some lingering appearances of it:

```
docs/document-matchzoo.md:13: -generator JsoupGenerator -threads 16 -input /path/to/robust04 \
docs/regressions-log.md:25:Previously, Core17 used the `NewYorkTimesCollection` and was indexed with `JsoupGenerator` as the generator, which assumes that the input is HTML (or XML) and removes tags.
docs/runbook-ecir2019-ccrf.md:68: -generator JsoupGenerator -threads 16 -input /path/to/robust04 \
docs/runbook-ecir2019-ccrf.md:73: -generator JsoupGenerator -threads 16 -input /path/to/robust05 \
docs/runbook-ecir2019-ccrf.md:78: -generator JsoupGenerator -threads 16 -input /path/to/core17 \
docs/runbook-trec2018-anserini.md:117: -generator JsoupGenerator -threads 44 -input /path/to/cw12 -index lucene-index.cw12.pos+docvectors+rawdocs \
docs/runbook-trec2018-anserini.md:135: -generator JsoupGenerator -threads 44 -storePositions -storeDocvectors -storeRawDocs -optimize \
docs/runbook-trec2018-anserini.md:145: -generator JsoupGenerator -threads 44 -input /path/to/cw12 -index \
docs/runbook-trec2018-anserini.md:164: -generator JsoupGenerator -threads 44 -storePositions -storeDocvectors -storeRawDocs \
docs/runbook-trec2018-anserini.md:174: -generator JsoupGenerator -threads 44 -input /path/to/cw12 -index \
docs/runbook-trec2018-anserini.md:185: -generator JsoupGenerator -threads 8 -uniqueDocid -storePositions -storeDocvectors \
docs/runbook-trec2018-anserini.md:195:    -generator JsoupGenerator -threads 8 -uniqueDocid -storePositions -storeDocvectors \
docs/runbook-trec2018-anserini.md:209:-generator JsoupGenerator -threads 16 -input enwiki-20180620-pages-articles.xml.bz2 -index \
[...]

PR #1371, opened 2020-09-08
Purge JsoupGenerator in docs

adamyy:
This is a continuation of https://github.com/castorini/anserini/pull/1355 and it simply replaces `JsoupGenerator` with `DefaultLuceneDocumentGenerator` in docs.

Collections involved:
- `TrecCollection`
- `NewYorkTimesCollection`
- `ClueWeb12Collection`
- `ClueWeb09Collection`
- `HtmlCollection`
- `JsonCollection`
- `WikipediaCollection`

I was able to find examples of `JsoupGenerator` being replaced with `DefaultLuceneDocumentGenerator` for most of these collections. With the exception of `WikipediaCollection`, which only appeared once, in https://github.com/castorini/anserini/blob/master/docs/runbook-trec2018-anserini.md.

### DIFF
--- a/docs/document-matchzoo.md
+++ b/docs/document-matchzoo.md
@@ -10,7 +10,7 @@ This is the document for the intergration between Anserini and MatchZoo. Current
 
 ```
 target/appassembler/bin/IndexCollection -collection TrecCollection \
- -generator DefaultLuceneDocumentGenerator -threads 16 -input /path/to/robust04 \
+ -generator JsoupGenerator -threads 16 -input /path/to/robust04 \
  -index lucene-index.robust04.pos+docvectors+rawdocs \
  -storePositions -storeDocvectors -storeRawDocs >& log.robust04.pos+docvectors+rawdocs
 ```
--- a/docs/document-matchzoo.md
+++ b/docs/document-matchzoo.md
@@ -10,7 +10,7 @@ This is the document for the intergration between Anserini and MatchZoo. Current
 
 ```
 target/appassembler/bin/IndexCollection -collection TrecCollection \
- -generator JsoupGenerator -threads 16 -input /path/to/robust04 \
+ -generator DefaultLuceneDocumentGenerator -threads 16 -input /path/to/robust04 \
  -index lucene-index.robust04.pos+docvectors+rawdocs \
  -storePositions -storeDocvectors -storeRawDocs >& log.robust04.pos+docvectors+rawdocs
 ```
--- a/docs/runbook-ecir2019-ccrf.md
+++ b/docs/runbook-ecir2019-ccrf.md
@@ -65,17 +65,17 @@ Run the following commands to index the `Robust04`, `Robust05`, and `Core17` col
 
 ```bash
 nohup sh target/appassembler/bin/IndexCollection -collection TrecCollection \
- -generator JsoupGenerator -threads 16 -input /path/to/robust04 \
+ -generator DefaultLuceneDocumentGenerator -threads 16 -input /path/to/robust04 \
  -index lucene-index.robust04.pos+docvectors+rawdocs \
  -storePositions -storeDocvectors -storeRawDocs >& log.robust04.pos+docvectors+rawdocs &
 
 nohup sh target/appassembler/bin/IndexCollection -collection TrecCollection \
- -generator JsoupGenerator -threads 16 -input /path/to/robust05 \
+ -generator DefaultLuceneDocumentGenerator -threads 16 -input /path/to/robust05 \
  -index lucene-index.robust05.pos+docvectors+rawdocs \
  -storePositions -storeDocvectors -storeRawDocs >& log.robust05.pos+docvectors+rawdocs &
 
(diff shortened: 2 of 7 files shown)


docs/document-matchzoo.md
```diff
@@ -10,7 +10,7 @@ This is the document for the intergration between Anserini and MatchZoo. Current
 
 ```
 target/appassembler/bin/IndexCollection -collection TrecCollection \
- -generator JsoupGenerator -threads 16 -input /path/to/robust04 \
```
lintool, line 13: We can remove this page outright.

docs/runbook-ecir2019-ccrf.md
```diff
@@ -65,17 +65,17 @@ Run the following commands to index the `Robust04`, `Robust05`, and `Core17` col
 
 ```bash
 nohup sh target/appassembler/bin/IndexCollection -collection TrecCollection \
- -generator JsoupGenerator -threads 16 -input /path/to/robust04 \
```
lintool, line 68: hrm, I'm torn about this.

The `runbook-*` files are really runbooks, so it's implicit that they only work with the commit at the time they were written... 

So if we roll back to an old commit, the runbooks should still work.

I'm worried that if we partially fix the runbook, then we'll be in a weird broken state...

thoughts?

src/main/python/paragraph_indexing/README.md
```diff
@@ -45,7 +45,7 @@ Suppose you're under Anserini directory. First indexing
 
 ```bash
 nohup sh target/appassembler/bin/IndexCollection -collection TrecCollection \
- -input /path/to/disk45/ -generator JsoupGenerator \
+ -input /path/to/disk45/ -generator DefaultLuceneDocumentGenerator \
```
lintool, line 48: This is really old code, I think we can delete, period?

Seems like `passage_retrieval/example/robust04.md` is the updated version?

lintool (commented):
Looking for your comment on the runbook issue?

For the files we can delete outright, can you start a separate PR to remove? This will keep history much cleaner. Thanks!

docs/runbook-ecir2019-ccrf.md
adamyy: I was also a bit concerning about the changes to runbooks. IMHO it would be nice to note in the runbook which commit specifically is the runbook validated against. May be this will be the commit before `JsoupGenerator` is removed?

Perhaps I should leave out the changes to runbooks for now? And maybe we can create a ticket for validating and stamping the runbook with a given *working* commit, then changes to the runbook can be restricted.

src/main/resources/fine_tuning/collections.yaml
```diff
@@ -13,7 +13,7 @@ collections:
         - -storeDocvectors
         - -storeRaw
       collection: TrecCollection
-      generator: JsoupGenerator
+      generator: DefaultLuceneDocumentGenerator
```
adamyy, line 16: Actually I was a bit hesitant about the changes here as well. As you mentioned here https://github.com/castorini/anserini/pull/1355#issuecomment-675485158 @lintool, updating non-running code would be a nightmare to verify. In this case, do we want to keep `fine_tuning`'s doc updated?

adamyy:
So action items:

1. Remove `docs/document-matchzoo.md` (in separate PR)
2. Undo changes to `runbooks-*`
3. Remove `src/main/python/paragraph_indexing` module? (looking at https://github.com/castorini/anserini/pull/430 I am not quite certain that it is now deprecated)

@lintool ?

lintool:
> So action items:
> 
> 1. Remove `docs/document-matchzoo.md` (in separate PR)

yup

> 2. Undo changes to `runbooks-*`

yup

> 3. Remove `src/main/python/paragraph_indexing` module? (looking at #430 I am not quite certain that it is now deprecated)

yes, but that code is pretty old...

iirc, I asked @emmileaf to rewrite, and that's the other verison.

@emmileaf are you still around? 

> 
> @lintool ?

emmileaf:
Hi! Jimmy is correct that `src/main/python/paragraph_indexing` is older code by Kyle a while back, and `src/main/python/passage_retrieval` is what I rewrote to experiment with paragraph/passage-based retrieval last year.

Even that was written during the earlier stages of Pyserini development though (before it moved out to a separate repo), so it’s a bit out of date as well. The idea was that Pyserini would make it easy to integrate document parsing with passing content through NLTK tokenizers, so I'd imagine any tokenizing-based work would be able to do it that way now.

Let me know if this helps or if you have any other questions!

lintool:
@emmileaf thanks for the quick response!

so, @adamyy let's keep `src/main/python/paragraph_indexing`, remove `src/main/python/paragraph_indexing`.

Thanks all!

arjenpdevries:
Keep _passage retrieval_ I suppose?

lintool:
> Keep _passage retrieval_ I suppose?

haha, good call: delete `src/main/python/paragraph_indexing`, keep `src/main/python/passage_retrieval`!

lintool:
I think this PR can be closed? superseded by #1374 and #1375

PR #1378, opened 2020-09-10
MS MARCO Passage and Doc Retrieval Duplication

wiltan-uw:
Environment: 
OS: Ubuntu 18.04.4 LTS (VM)
Python 3.8.5
Apache Maven 3.6.0

Results: The results are identical to the those on the experiments page.

Passage with default params
passage_default
Passage with tuned params
passage

Doc with default params
doc
Doc with tuned params
doc_tuned

Issues/concerns:
Retrieval runs for Doc took 10~12 hours, in comparison to the 12 minutes in the guideline, which is unexpected.
Might be due to the limitation of VirtualBox VM.
One unit test failed for Pyserini, but did not seem to affect the results of this duplication.


lintool:
> Issues/concerns:
Retrieval runs for Doc took 10~12 hours, in comparison to the 12 minutes in the guideline, which is unexpected.
Might be due to the limitation of VirtualBox VM.
One unit test failed for Pyserini, but did not seem to affect the results of this duplication.

Might be because of VM...
Please file an issue with failed test - we should at least know about it.

lintool (approved):

Issue #1010, opened 2020-03-03
"Direct installation" of Lucene indexes into Solrini


lintool:
Solr is basically just a webapp around Lucene. It is possible to bypass the Solr indexing REST APIs and just directly copy the Lucene indexes into the right path in the Solr directory structure. We should provide documentation on how to do this, to save users an extra indexing step.

edwinzhng:
Looked into it a bit, according to the steps [here](https://stackoverflow.com/a/35927597) from 2016, we should:

1. Explicitly create the Solr schema via the JSON API for **all** fields (will need to add more on top of the existing schema scripts). This is also good since we can control types in one place instead of letting Solr infer the field type like it does right now. Probably makes sense to do #1009 first

2. Change the "dataDir" in solrconfig.xml to point to the Lucene index

lintool:
hey @r-clancy you did this before, right? can you comment on whether @edwinzhng 's on the right track?

@edwinzhng for step 2 above we have pre-built indexes here: https://git.uwaterloo.ca/jimmylin/anserini-indexes

We can get just `wget` it and uncompress into the location that Lucene expects it... so we won't even need to change the `dataDir`?

r-clancy:
@lintool Yeah, I had done this before. @edwinzhng is on the right track - the process is just making sure the Solr schema aligns with the Lucene schema and copying the Lucene index (or changing the Solr data dir to point to the Lucene index) into the correct place.

PR #1379, opened 2020-09-11
[#1010] Add Solr instructions for prebuilt Lucene index

adamyy:
https://github.com/castorini/anserini/issues/1010

* Provides instructions about how to configure Solr to read prebuilt Lucene indexes

### DIFF
--- a/docs/solrini.md
+++ b/docs/solrini.md
@@ -48,6 +48,7 @@ curl -X POST -H 'Content-type:application/json' --data-binary @src/main/resource
 We can use Anserini as a common "frontend" for indexing into SolrCloud, thus supporting the same range of test collections that's already included in Anserini (when directly building local Lucene indexes).
 Indexing into Solr is similar indexing to disk with Lucene, with a few added parameters.
 Most notably, we replace the `-index` parameter (which specifies the Lucene index path on disk) with Solr parameters.
+Alternatively, Solr can also be configured to [read prebuilt Lucene index](#markdown-header-solr-with-prebuilt-lucene-index), since Solr uses Lucene index under the hood.
 
 We'll index [robust04](regressions-robust04.md) as an example.
 First, create the `robust04` collection in Solr:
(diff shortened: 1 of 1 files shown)


docs/solrini.md
```diff
@@ -98,6 +99,54 @@ Make sure `core18` collection is created and `/path/to/WashingtonPost` is update
 Solrini has also been verified to work with the [MS MARCO Passage Retrieval Corpus](experiments-msmarco-passage.md).
 There should be no major issues with other collections that are supported by Anserini, but we have not tested them.
 
+## Solr with Prebuilt Lucene Index
+
+Solr can be considered a front-end for Lucene, and it is entirely possible for Solr to read prebuilt Lucene indexes. 
+To achieve this, some housekeeping are required. The following uses [robust04](regressions-robust04.md) as an example.
```
lintool, line 105: We adopt the convention of each sentence on a line; here and below.
robust04 -> Roboust04

docs/solrini.md
```diff
@@ -98,6 +99,54 @@ Make sure `core18` collection is created and `/path/to/WashingtonPost` is update
...
+
+Second, make proper Solr schema adjustments if required.
+Here `robust04` is a Trec collection whose schema is already taken care of by [managed-schema](https://github.com/castorini/anserini/blob/master/src/main/resources/solr/anserini/conf/managed-schema) in the Solr configset.
+However, if you are dealing with a collection such as `cord19`, remember to make proper adjustments to the Solr schema, as [previously described](#setting-up-a-single-node-solrcloud-instance).
+
+```
+curl -X POST -H 'Content-type:application/json' --data-binary @src/main/resources/solr/schemas/SCHEMA_NAME.json http://localhost:8983/solr/COLLECTION_NAME/schema
+```
+
+Then, either *copy the index files to where Solr expected*, or *tell Solr where to look for*.
```
lintool, line 124: Let's be opinionated - pick the better option (my vote is for the first) and do that - don't give the user the choice.

docs/solrini.md
```diff
@@ -98,6 +99,54 @@ Make sure `core18` collection is created and `/path/to/WashingtonPost` is update
...
+
+```
+solrini/bin/solr create -n anserini -c robust04
+```
+
+Along with the collection, Solr will create a core instance, whose name can be found in the Solr UI under collection overview.
+Which might look something like `<collection_name>_shard<id>_replica_<id>` (e.g., `robust04_shard1_replica_n1`). Solr stores configurations and data for the core instances under Solr home, which for us is `solrini/server/solr/` by default.
+
+Second, make proper Solr schema adjustments if required.
+Here `robust04` is a Trec collection whose schema is already taken care of by [managed-schema](https://github.com/castorini/anserini/blob/master/src/main/resources/solr/anserini/conf/managed-schema) in the Solr configset.
```
lintool, line 117: Trec -> TREC

lintool (changes requested):
Minor comments, thanks for the PR!

adamyy:
Actually, as I was testing the prebuilt index with `run_solr_integration.py`, I noticed that there are discrepancy between expected and actual trec_eval benchmark. Currently looking into the cause and this PR might not be ready to merge just yet!

adamyy:
Looked a bit more into this, turns out my index was not built with the `contents` field (only `id` and `raw`) and `SearchSolr` uses `contents` as the default field of search. So it either errors out / produce nonsense output.

I guess it is worth noting that the prebuilt index should have the `contents` field for this to work properly? When we are indexing into Solr using `IndexCollection`, it seems like the `contents` field are created regardless of whether we specify `-storeContents` or not. However, if we are just producing a Lucene index then `contents` field only present if we specify `-storeContents`.

lintool:
yup, this should be documented somewhere!

lintool:
Also, check out: https://github.com/castorini/anserini/blob/master/.travis.yml#L20

It _might_ be reasonable to do a complete e2e integration test based on this? I mean, travis is running it all the time... just a matter of moving it inside JUnit?

adamyy:
Actually, I have misunderstood the issue again. 😅 

I read more about what `stored` means for a Lucene Document Field, and found this [explanation](https://stackoverflow.com/a/4185226). Turns out `stored=false` still makes the field searchable, it is just that the field is not stored as is, and will not be returned as part of the query result.

Then I was wondering, how come `SearchCollection` on non-`storeContents` MSMARCO index can produce the output that matches the regression, but `SearchSolr` cannot?

I noticed [solrconfig.xml#L785](https://github.com/castorini/anserini/blob/master/src/main/resources/solr/anserini/conf/solrconfig.xml#L785) set the default field of search to a field called `_text_` (this is the default Solr config). However, in our case we want it to be `contents`, and it is what we were trying to enforce at [SearchSolr.java#L225](https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SearchSolr.java#L225). My hunch is that Solr internally picks the default field of search based on both query parameters and solrconfig.xml, and it might have picked the wrong one in my previous run.

I modified this field from `_text_` to `contents`, recreated the Solr core with the updated configset and the non-`storeContents` index, then performed `SearchSolr` again. The good news is that the result now matches the regression. I will perform a couple more tests to ensure that this is what is causing the issue.

adamyy:
> Also, check out: https://github.com/castorini/anserini/blob/master/.travis.yml#L20
> 
> It _might_ be reasonable to do a complete e2e integration test based on this? I mean, travis is running it all the time... just a matter of moving it inside JUnit?

Do we want to do `run_solr_regression.py` in travis? @lintool

adamyy:
> I modified this field from `_text_` to `contents`, recreated the Solr core with the updated configset and the non-`storeContents` index, then performed `SearchSolr` again. The good news is that the result now matches the regression. I will perform a couple more tests to ensure that this is what is causing the issue.

Big oops, I tested my assumption again with a newly built index (without `-storeContents` flag) everything worked just fine. TL;DR I think my original index was broken. Sorry for causing the confusion! This PR should be ready now! @lintool

lintool:
> Do we want to do `run_solr_regression.py` in travis? @lintool

Maybe `run_solr_regression.py` with the small CACM collection in travis? Shouldn't be too difficult to do? If you think it's a good idea, create another issue to keep track of progress?

lintool (approved):

Issue #1259, opened 2020-06-05
Write ingester for TREC 2020 Health Misinformation Track


lintool:
The collection for the TREC 2020 Health Misinformation Track appears to be common crawl WARCS.

ClueWeb12 is distributed as WARCs also:
https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/collection/ClueWeb12Collection.java

See if we can adapt code there? Or maybe common crawl has their own APIs?

Call this `CommonCrawlCollection`.

MXueguang:
There are some difference between ClueWeb12's WARCs and CommonCrawl's WARCs.

1. The fields of WARC header is different
2. CommonCrawl contains WARC request while ClueWeb not. 
    - a content given by CW is: <WARC_RESPONSE> <HTTP_RESPONSE> <CONTENT_BODY>
    - a content given by CC is: <WARC_REQUEST> <HTTP_REQUEST> <WARC_RESPONSE> <HTTP_RESPONSE> <CONTENT_BODY>

ClueWeb's WARC example:
```
WARC/1.0
WARC-Type: response
WARC-Date: 2012-02-10T21:51:20Z
WARC-TREC-ID: clueweb12-0000tw-00-00013
WARC-Target-URI: http://cheapcosthealthinsurance.com/2012/01/25/what-is-hiv-aids/
WARC-Payload-Digest: sha1:YZUOJNSUMFG3JVUKM6LBHMRMMHWLVNQ4
WARC-IP-Address: 100.42.59.15
WARC-Record-ID: <urn:uuid:74edc71e-a881-4942-81fc-a40db4bf1fb9>
Content-Type: application/http; msgtype=response
Content-Length: 71726

HTTP/1.1 200 OK
Date: Fri, 10 Feb 2012 21:51:22 GMT
Server: Apache/2.2.21 (Unix) mod_ssl/2.2.21 OpenSSL/0.9.8e-fips-rhel5 mod_auth_passthrough/2.1 mod_bwlimited/1.4 FrontPage/5.0.2.2635 mod_jk/1.2.32
X-Powered-By: PHP/5.2.17
X-Pingback: http://cheapcosthealthinsurance.com/xmlrpc.php
Link: <http://cheapcosthealthinsurance.com/?p=711>; rel=shortlink
Connection: close
Content-Type: text/html; charset=UTF-8


[...]

PR #1383, opened 2020-09-15
give more information in raw() of WARC indexes 

MXueguang:
It was returning the WARC content only.

By issue https://github.com/castorini/anserini/issues/1259, since other info such as URL might be useful, it is better to keep them in raw().
Now it will return a json dict with `Date`, `ID`, `URL`, `Document`


lintool:
hey @shaneding you're pretty familiar with Anserini at this point, can you take a look?

@MXueguang What I'm trying to decide is whether this is a one-off ugly hack, or if this is generally a good idea, to just start stuffing fields in `raw()` output as a JSON... An alternative, for example, would be for `raw()` to just return the entire WARC crawl record - which obviously has all this stuff in it, right? And let the client parse out whatever they want...

I'm worried that if we start "structuring" the "raw" document, we start down the path of endless document parsing whenever someone want some new field...

cc @ronakice thoughts?

shaneding:
I just took a quick look, why can't we just add `URL`, `date`, and etc as fields in the lucene index? Why do we need to organize it for raw? @lintool

lintool:
@shaneding *slaps forehead* e.g., like a custom generator? e.g., https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/generator/Iso19115Generator.java

@MXueguang I think that's a clean solution?

shaneding:
Exactly, I think thats a more general solution and making them fields just make searching a lot more robust imho @lintool.

MXueguang:
@lintool @shaneding Yeah, custom generator is indeed a much better solution.
Thanks for your suggestions. I'll update it asap

MXueguang:
now the url (or date) are expect to retrieval via:
`hits[0].lucene_document.get("url")`

lintool:
Sorry, I'm missing the broader context: why did `readNextRecord` get removed? and whats this integer offset?
Can we add a simple test case?

MXueguang:
Sorry I reverted something back.

I noticed there might be a bug for the `readNextRecord`. Some of the raws of WARC doc seems missing the last character. I thought it was only for CommonCrawl but I just notice it happened to ClueWeb as well. I need to further investigate this to make sure if it is actually a "bug". It it does, i think i should fix that in a new PR. (It shouldn't influence any current indexing anyway because the last char of `raw` never kept in `contents`)

lintool:
Can we easily add a test to somewhere in https://github.com/castorini/anserini/tree/master/src/test/java/io/anserini/collection ?

MXueguang:
Added test cases for all collections that using `WarcBaseDocument`

lintool (approved):

PR #1385, opened 2020-09-24
Update doc.raw() to return the entire raw JSON

x389liu:


src/main/java/io/anserini/collection/JsonCollection.java
```diff
@@ -164,10 +166,10 @@ public String contents() {
 
     @Override
     public String raw() {
-      if (contents == null) {
-        throw new RuntimeException("JSON document has no \"contents\" field");
+      if (raw == null) {
+        throw new RuntimeException("JSON document has no \"raw\" field");
```
lintool, line 170: I don't think `raw` can ever be `null`? Can you check? If `raw` is `null`, then how did we read the document in the first place...?

lintool (commented):
See comment?

src/main/java/io/anserini/collection/JsonCollection.java
lintool: And either way, I don't think this is the right error message... "raw" isn't a field in the JSON...

x389liu: Thanks for pointing this out, `raw` can never be `null` as indexed docs always have `id`/`contents`.

lintool (approved):

PR #1403, opened 2020-11-08
Refactor Solr/ES scripts + docs

lintool:
+ PEP8 compliance for scripts
+ Went through documentation, updated

@shaneding can you do a CR for me here also? Don't need to run the commands (I have), but at least sanity check to make sure I didn't screw up anything obvious.


lintool:
@adamyy can you take a quick look also, since you recently ran these experiments. Thanks!

src/main/python/run_solr_regression.py
```diff
@@ -29,36 +28,43 @@
...
 
-    def is_alive(self):
+    @staticmethod
+    def is_alive():
         try:
             response = requests.get('http://localhost:8983/')
             response.raise_for_status()
-        except: return False
-        else: return True
+        except requests.exceptions.RequestException:
```
edwinzhng, line 41: There's no other exception that could be raised here? (and elsewhere)

edwinzhng (approved):
LGTM, seems like everything is in order

PR #1411, opened 2020-11-14
More general implementation of segment retrieval (MaxP)

lintool:
Improved implementation of segment retrieval for MS MARCO per-segment doc expansion, so that we don't need a post-processing script to compute MaxP.

A similar feature already existed as `-strip_segment_id`, used by TREC-COVID in the paragraph condition. I improved and generalized in to `-selectMaxSegment` option. Added test case.

Had to refactor old TREC-COVID scripts/document to use new option.


rodrigonogueira4 (approved):
LGTM!

ronakice (approved):
LGTM.

PR #1413, opened 2020-11-16
Renamed -selectMaxSegment to -selectMaxPassage

### LINKED ISSUE #1412: Rename -selectMaxSegment
I just noticed that in the context of -selectMaxSegment, we refer to passages as segments. Should we rename to "passage" instead to be consistent with the literature? So, `-selectMaxPassage` or even `-selectMaxP`?

Let's decide before this feature gets engrained in its usage so we don't have to change much documentation.

lintool:
Closes #1412


rodrigonogueira4 (approved):
LGTM!

PR #1418, opened 2020-11-21
Tweak path to gdeval

lintool:
see #1416 - changes to path need to be propagated to rest of documentation.


rodrigonogueira4 (approved):
LGTM, but any reason for changing the topics (51-100 vs 101-150)?

lintool:
I think that's just weirdness in how the diffs are displayed?

rodrigonogueira4:
> I think that's just weirdness in how the diffs are displayed?

Ah, sorry, you are right!

ronakice (approved):
LGTM!

PR #1421, opened 2020-11-22
Fixed paths to eval tools for fine-tuning experiments

lintool:
Haven't run these regressions in a while... fixed path to eval tools.


rodrigonogueira4 (approved):
LGTM!

PR #1426, opened 2020-11-24
add 20newsgroups replication

stephaniewhoo:


lintool:
hey @stephaniewhoo you checked out the prebuilt indexes also? If not, can you pull them and sanity check? Or I suppose you'll do that implicitly when you move over to the pyserini end.

lintool (approved):

PR #1429, opened 2020-11-27
change sortedmap

qguo96:
Please don't review. This is a draft.

### DIFF
--- a/src/test/java/io/anserini/search/topicreader/CacmTopicReaderTest.java
+++ b/src/test/java/io/anserini/search/topicreader/CacmTopicReaderTest.java
@@ -32,7 +32,7 @@ public class CacmTopicReaderTest {
         Paths.get("src/main/resources/topics-and-qrels/topics.cacm.txt"));
 
     Map<Integer, Map<String, String>> topics = reader.read();
-    Integer[] keys = (Integer[]) topics.keySet().toArray();
+    Integer[] keys = topics.keySet().toArray(new Integer[0]);
     Integer firstKey = keys[0];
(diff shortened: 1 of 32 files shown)


src/test/java/io/anserini/search/topicreader/CacmTopicReaderTest.java
```diff
@@ -32,16 +31,19 @@ public void test() throws IOException {
     TopicReader<Integer> reader = new CacmTopicReader(
         Paths.get("src/main/resources/topics-and-qrels/topics.cacm.txt"));
 
-    SortedMap<Integer, Map<String, String>> topics = reader.read();
-
+    Map<Integer, Map<String, String>> topics = reader.read();
```
lintool, line 34: This seems pretty ugly... there's gotta be a better way to do this?

lintool:
High-level initial comment: If I understand correctly, you want to change the type of a set of topics from `SortedMap` to just `Map`. This is so that subclasses can choose their implementations, for example, a map backed by a linked list.

However, this means that wrt the test cases, the concept of a "first" and "last" key becomes ambiguous - is it wrt to the sort order of the keys or the ordering of the underlying topics in the file? If the topic file is sorted, then these are the same. But in the case of MS MARCO, these are different. So, we need to rethink the testing strategy, e.g., test both?

Perhaps we should just move away from `Map` completely? That is, introduce an intermediate interface, e.g., `Topics<T>` where `T` is the type of the topic id - and hide implementation details, so we have something like `topics.getId(5).getField("title")`? Then, in principle, implementations don't even need to be Map-based.

Issue #1437, opened 2020-12-11
unexpected docvalues type NONE for field 'id'


AndyTheFactory:
I indexed a small test using TweetCollection :

>target/appassembler/bin/IndexCollection.bat -collection TweetCollection  -input collections/test -index indexes/lucene-index.test  -generator TweetGenerator  -threads 44 -storePositions -storeDocvectors -storeRaw -uniqueDocid -tweet.keepUrls -tweet.stemming

It worked fine :

```
2020-12-11 00:16:06,371 INFO  [main] index.IndexCollection (IndexCollection.java:650) - Setting log level to INFO
2020-12-11 00:16:06,374 INFO  [main] index.IndexCollection (IndexCollection.java:653) - Starting indexer...
2020-12-11 00:16:06,375 INFO  [main] index.IndexCollection (IndexCollection.java:654) - ============ Loading Parameters ============
2020-12-11 00:16:06,375 INFO  [main] index.IndexCollection (IndexCollection.java:655) - DocumentCollection path: collections/test
2020-12-11 00:16:06,375 INFO  [main] index.IndexCollection (IndexCollection.java:656) - CollectionClass: TweetCollection
2020-12-11 00:16:06,376 INFO  [main] index.IndexCollection (IndexCollection.java:657) - Generator: TweetGenerator
2020-12-11 00:16:06,376 INFO  [main] index.IndexCollection (IndexCollection.java:658) - Threads: 44
2020-12-11 00:16:06,376 INFO  [main] index.IndexCollection (IndexCollection.java:659) - Stemmer: porter
2020-12-11 00:16:06,377 INFO  [main] index.IndexCollection (IndexCollection.java:660) - Keep stopwords? false
2020-12-11 00:16:06,377 INFO  [main] index.IndexCollection (IndexCollection.java:661) - Stopwords:  null
2020-12-11 00:16:06,377 INFO  [main] index.IndexCollection (IndexCollection.java:662) - Store positions? true
2020-12-11 00:16:06,378 INFO  [main] index.IndexCollection (IndexCollection.java:663) - Store docvectors? true
2020-12-11 00:16:06,378 INFO  [main] index.IndexCollection (IndexCollection.java:664) - Store document "contents" field? false
2020-12-11 00:16:06,378 INFO  [main] index.IndexCollection (IndexCollection.java:665) - Store document "raw" field? true
[...]

PR #1439, opened 2020-12-13
Added -searchtweets option to SimpleSearcher

AndyTheFactory:
Fix for #1437 tie break ruled for Tweets by ID


lintool:
hi @AndyTheFactory thanks for the PR!

I'm a bit backed up right now, but I'll add this to my queue.
If you could add test cases, that would be great.

lintool:
hi @AndyTheFactory I didn't remember until now, the first time you asked... but we have: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SimpleTweetSearcher.java

So this isn't necessary...

Sorry for the trouble.

Issue #1445, opened 2021-01-04
Replicate MS MARCO Document Leaderboard


MXueguang:
https://github.com/castorini/anserini/blob/master/docs/experiments-msmarco-doc-leaderboard.md

seems we don't have this index anymore
`index-msmarco-doc-expanded-per-passage-20201126-1b4d0a.54ea30c64515edf3c3741291b785be53`
```
target/appassembler/bin/SearchCollection -topicreader TsvString -index ~/.cache/pyserini/indexes/index-msmarco-doc-expanded-per-passage-20201126-1b4d0a.54ea30c64515edf3c3741291b785be53 \
  -topics src/main/resources/topics-and-qrels/topics.msmarco-doc.dev.txt -output runs/doc2query-t5-per-passage/dev.trec.txt \
  -bm25 -bm25.k1 2.56 -bm25.b 0.59 -hits 1000 -selectMaxPassage -selectMaxPassage.delimiter "#" -selectMaxPassage.hits 100 &
```

I tried to replace with `index-msmarco-doc-per-passage-20201204-f50dcc.797367406a7542b649cefa6b41cf4c33`
but the result gives:
```
#####################
MRR @100: 0.2727185871733738
QueriesRanked: 5193
#####################
```
which is not what we expected.

lintool:
Closed by #1446 - @MXueguang please send PR with replication log entry.

PR #1446, opened 2021-01-06
Update experiments-msmarco-doc-leaderboard.md

lintool:
Fixed minor typo that should address #1445

@MXueguang please take a look?


MXueguang:
LGTM, replicated and get exactly same scores

PR #1453, opened 2021-01-14
Updated documentation about MS MARCO doc

lintool:
I realized that we hadn't updated this documentation since the doc leaderboard became available. Updating accordingly.


ronakice (commented):
LGTM!

rodrigonogueira4 (approved):

PR #1456, opened 2021-01-17
Add MS MARCO doc per-passage condition

lintool:
Both default and tuned BM25.


ronakice:
Actually, quick question, why no multi-threaded search in the per-doc settings?

lintool:
@ronakice `SearchCollection` has multi-threading, but it's per-run - for parameter sweeps. We haven't implemented intra-run parallelism yet... I've been meaning to...

Issue #1461, opened 2021-01-19
Qrels test cases


lintool:
Hey @MXueguang - while you're revamping the `Qrels` classes, can you add in a few simple tests?

<img width="594" alt="Screen Shot 2021-01-18 at 9 37 35 PM" src="https://user-images.githubusercontent.com/313837/104980778-66a61a80-59d5-11eb-9af4-91a25138b9a1.png">

Like a few tests cases to verify some of the above stats (from the book)? E.g., number of queries, qrels, avg qrels per query, etc. Should be pretty straightforward - iterate through the data structures and count, etc.

MXueguang:
sure

MXueguang:
The size of the topics has been tested in `TopicReaderTest`
so we only need to test the size of `qrels` and some qrels values in `QrelsTest` right?

lintool:
Correct!

PR #1463, opened 2021-01-19
Add method in Qrels to read file as string & define Qrel IDs

MXueguang:
added unit test re https://github.com/castorini/anserini/issues/1461


lintool (commented):
Since this part of the code base isn't being actively used, we can redesign without danger of breaking anything.

How about we move into a new package `.eval`?

How about we rename `Qrels` to `RelevanceJudgments` and `QrelsID` can just become `Qrels`?

`RelevanceJudgments` fits well with methods we currently have, e.g., `isDocJudged` and `getRelevanceGrade`.

Open to discussion... thoughts?

MXueguang:
yeah! I was thinking about renaming the original `Qrels` to something before. `RelevanceJudgments` make sense.

lintool (approved):
Please add the standard boilerplate to the new files...

```
/*
 * Anserini: A Lucene toolkit for replicable information retrieval research
 *
...
```

Otherwise, lgtm!

PR #1467, opened 2021-01-21
Add TREC19DL topics&qrels

MXueguang:


lintool (commented):
Check out: https://arxiv.org/abs/2003.07820

There are no separate doc/passage topics (queries) - they're the same...

lintool:
Wait:

```
$ diff topics.dl19-passage.txt topics.dl19-doc.txt 
7d6
< 168216	does legionella pneumophila cause pneumonia
24a24
> 1132213	how long to hold bow in yoga
28a29
> 287683	how many liberty ships were built in brunswick
33d33
< 1121709	what are the three percenters?
```

Huh, they are _not_ identical. @MXueguang Can you read the paper and see what it says?

lintool:
And here: https://microsoft.github.io/msmarco/TREC-Deep-Learning-2019

MXueguang:
> The track has two tasks: Document retrieval and passage retrieval. Participants were allowed to submit up to three
> runs per task, although this was not strictly enforced. Participants were provided with an initial set of 200 test queries,
> then NIST later selected 43 queries during the pooling and judging process, based on budget constraints and with the
> goal of producing a reusable test collection. **The same 200 queries were used for submissions in both tasks, while the
> selected 43 queries for each task were overlapping but not identical.** The full judging process is described in Section 5

lintool:
Ah, you're right!

lintool (approved):
Can you add a test case to differentiate between passage and doc? I.e., to catch if the two topic sets are being confused?

Otherwise, lgtm.

MXueguang:
just added qrels

lintool (approved):
🚢 it!

lintool:
@MXueguang please check in here: https://github.com/castorini/anserini-tools/tree/master/topics-and-qrels

We replicate in that location also. (we had a long debate on this duplicate a while ago... but tl;dr - yes, duplication is okay.)

PR #1470, opened 2021-01-22
add dpr nq-dev topic & topic reader

MXueguang:


lintool (approved):
lgtm

should we add test set while we're at it?

MXueguang:
yes, we should. will add in this PR

PR #1472, opened 2021-01-25
Do not load all vectors into memory while indexing

tteofili:
Avoid loading all vectors in a `Map` within `IndexVectors` for unnecessary memory consumption, especially with large vector models.


lintool:
hey @tteofili I think your locale settings changed all `.` to `,` in the regression tests... can you update your PR to not do this?

tteofili:
very weird, sure I'll fix it.

lintool:
There doesn't seem to be any progress on this, closing for now.

PR #1473, opened 2021-01-27
Allow stopwords to be configured at search time.

jtibshirani:
Stopwords can currently be configured at index-time. This PR adds support for
specifying stopwords at search-time as well.

This is helpful because generally the same stopwords are used at both index and
search time. If the stopword 'where' is dropped from the index but still
included in searches, then a search for 'where' only matches words with the same
stem like 'wheres'. This is tricky behavior and usually undesirable.


src/test/java/io/anserini/analysis/DefaultEnglishAnalyzerTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.Test;
+
+import java.util.Arrays;
+
+import static io.anserini.analysis.DefaultEnglishAnalyzer.fromArguments;
+import static org.junit.Assert.assertEquals;
+
+public class DefaultEnglishAnalyzerTest {
+
+    @Test
```
lintool, line 14: two space indent to be consistent with rest of code base?

lintool (approved):
minor nit, please fix indent to be consistent, and then I'll merge.

thanks!

src/test/java/io/anserini/analysis/DefaultEnglishAnalyzerTest.java
jtibshirani: Oops, fixed! I had also forgotten the license header.

PR #1478, opened 2021-02-12
Get DocumentCollection to follow all symbolic links and make sure TrecCollection ignores the right files and directories

eelstretching:
DocumentCollection was not following symbolic links correctly when indexing TREC disks 1&2. The inline SimpleFileVisitor in the discover method attempted to determine if a given path was a symbolic link and deal with it there, but it was not correctly detecting a symbolic link with a relative path.

The four argument version of Files.walkFileTree will take a set of FileVisitOption that you can use to tell the visitor to follow symbolic links. Using this version, and eliding the check for symbolic links in visitFile results in the correct behavior for both relative and absolute symbolic links.

While working at getting TREC disks 1 and 2 indexed, I noticed that the values in the skippedFilePrefix and skippedDir sets are lower cased, while the file names from the TREC disks are upper cased. Rather than do something more drastic (e.g., making these sets of regexes to match case-insensitively) I added upper case variants of each of the members of these sets.

After these changes, I ran the TREC 1 and 2 topics regressions and got the expected results for map and P@30, as detailed in anserini/docs/regressions-disk12.md.


lintool:
Hi @eelstretching thanks for your contributions! We're in the middle of cutting a release right now... but this will get merged in the next release.

lintool (approved):

lintool:
Hi @eelstretching thanks for your contributions! Replications check out.

There was a previous report of an issue with TREC disks 1 & 2, see #1033

We don't actively use those test collections any more, so haven't had time to follow up... perhaps you'd be interested in taking a look? Thanks in advance!

eelstretching:
Sure, I'll have a look.

PR #1482, opened 2021-02-19
Ensure that multi-line TREC topic titles are fully parsed. Update the…

eelstretching:
Incorporated the change from @axlliu for ensuring that multi-line TREC topic titles are processed correctly. Added a test for title parsing to the set of tests.

Note that one of the existing tests (`testNewswireTopics`) happened to hit on a multi-line title (topic 200), and that test needed to be fixed to include the second line of the title.

As these changes change the queries, the numbers in the regressions for disk 1 and 2 need to change. I re-ran the index and the evaluations using the new code, which resulted in quite a few changes for the regression numbers (some better, some worse.)


lintool:
Great! Can you actually change the regression numbers here? https://github.com/castorini/anserini/blob/master/src/main/resources/regression/disk12.yaml

The documentation is auto-gen'ed based on yaml above.

eelstretching:
The YAML file's connected to the MD file, the MD file's connected to the...

Updated the YAML and it re-generated the MD file, so I committed both.

lintool:
@eelstretching great, thanks! I'll queue up a run of all regressions on my end to make sure it doesn't break anything else, and then I'll merge.

lintool (approved):

PR #1485, opened 2021-03-19
add qrels + topics dl20, update tools

ronakice:


lintool (approved):
@ronakice we also mirror in https://github.com/castorini/anserini-tools/tree/master/topics-and-qrels - please send PR there also?

PR #1488, opened 2021-03-20
[WIP] C4 indexer + generator

shaneding:
Added index+generator for C4.


src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+
+        @Override
+        public void readNext() throws NoSuchElementException {
+            if (node == null) {
+                throw new NoSuchElementException("JsonNode is empty");
+            } else {
+                bufferedRecord = new C4Collection.Document(node, fileName, count);
+                if (iterator.hasNext()) { // if bufferedReader contains JSON line objects, we parse the next JSON into node
```
lintool, line 79: I don't think you need this logic here because C4 doesn't change formats?

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -0,0 +1,138 @@
...
+        private String id;
+        private String contents;
+        private String raw;
+        private String url;
+        private String timestamp;
+
+        public Document(JsonNode json, String filename, int jsonLoc) {
+            this.raw = json.toPrettyString();
+            this.contents = json.get("text").asText();
+            this.id = filename +  '-' + jsonLoc;
```
lintool, line 99: add a comment here how you're constructing the id.

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        public String getTimestamp() {
+            return timestamp;
+        }
+
+        @Override
+        public String id() {
+            if (id == null) {
+                throw new RuntimeException("JSON document has no \"id\" field");
```
lintool, line 115: make the error collection specific...

src/main/java/io/anserini/index/generator/C4Generator.java
```diff
@@ -0,0 +1,53 @@
...
+        super(args);
+        this.args = args;
+    }
+
+    public Document createDocument(C4Collection.Document doc) throws GeneratorException {
+        Document document = super.createDocument(doc);
+
+        // contents and id stored else where
+        document.add(new StringField(C4Field.URL.name, doc.getUrl(), Field.Store.YES));
+        document.add(new StringField(C4Field.TIMESTAMP.name, doc.getTimestamp(), Field.Store.YES));
```
lintool, line 49: Can we normalize the date into a numeric field? something like epoch s or ms so we can do range queries etc. down the road...

lintool (changes requested):
Various comments...

Can we get a test case also?

src/main/java/io/anserini/collection/C4Collection.java
shaneding: don't we need this for when we reach eof though?

shaneding:
Tests are done, currently the test files are uncompressed since I am assuming that `GZIPInputStream` will behave correctly. But I can compress the testing file if needed.

lintool:
> Tests are done, currently the test files are uncompressed since I am assuming that `GZIPInputStream` will behave correctly. But I can compress the testing file if needed.

I think it's okay to check in a small gzipped file as part of the resources for the test.

src/main/java/io/anserini/collection/C4Collection.java
lintool: The comment is a bit confusing, I think? The entire file is JSON line, right?

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -0,0 +1,153 @@
...
+        private String url;
+        private long timestamp;
+
+        public Document(JsonNode json, String filename, int jsonLoc) {
+            this.raw = json.toPrettyString();
+            this.contents = json.get("text").asText();
+            // id is the filename appended by the document count
+            this.id = filename +  '-' + jsonLoc;
+            this.url = json.get("url").asText();
+            String dateTime = json.get("timestamp").asText();
```
lintool, line 110: you probably want to do some null checking here?

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -0,0 +1,153 @@
...
+
+        public String getUrl() {
+            return url;
+        }
+
+        public long getTimestamp() {
+            return timestamp;
+        }
+
+        private long getNumericTime(String dateTime) {
```
lintool, line 122: I think this method can be inlined? Do you call it elsewhere?

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -0,0 +1,153 @@
...
+            if (id == null) {
+                throw new RuntimeException("C4 document has no \"id\" field");
+            }
+            return id;
+        }
+
+        @Override
+        public String contents() {
+            if (contents == null) {
+                throw new RuntimeException("JSON document has no \"contents\" field");
```
lintool, line 138: field name is not correct...

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -0,0 +1,153 @@
...
+
+        private long getNumericTime(String dateTime) {
+            Instant i = Instant.parse(dateTime);
+            return i.getEpochSecond();
+        }
+
+        @Override
+        public String id() {
+            if (id == null) {
+                throw new RuntimeException("C4 document has no \"id\" field");
```
lintool, line 130: this shouldn't happen right? and the exception message is incorrect.

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -799,8 +772,19 @@ public Counters run() throws IOException {
     LOG.info("Thread pool with " + numThreads + " threads initialized.");
 
     LOG.info("Initializing collection in " + collectionPath.toString());
-    final List segmentPaths = collection.getSegmentPaths();
+    List<?> segmentPaths = collection.getSegmentPaths();
+
+    // for C4 specifically we filter through segmentPaths to only take ones that we want based on file #
+    if (args.collectionClass.equals("C4Collection") && args.shardCount > 1) {
```
lintool, line 778: I think it's clearer to do an outer `if (args.shardCount > 1)` and then an inner `if-else` checking the collection. If we add more collection-specific processing, it'll be cleaner.

src/main/java/io/anserini/index/generator/C4Generator.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public C4Generator(IndexArgs args) {
+        super(args);
+        this.args = args;
+    }
+
+    public Document createDocument(C4Collection.Document doc) throws GeneratorException {
+        Document document = super.createDocument(doc);
+
+        // contents and id stored else where
```
lintool, line 47: instead of "else where", say "in superclass method"

src/test/java/io/anserini/collection/C4CollectionTest.java
```diff
@@ -0,0 +1,40 @@
+package io.anserini.collection;
```
lintool, line 1: Add boilerplate?

lintool (changes requested):
I believe we do 2-space indent rest of the code base?

I'm also fine with long lines (say, up to 120 chars) unless they really get out of hand...

src/main/java/io/anserini/collection/C4Collection.java
shaneding: idt, for those fields they will always be there and in other json collections we never did null checking on fields.

src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -638,6 +612,15 @@ public void destroyObject(PooledObject<RestHighLevelClient> pooled) throws Excep
   private ObjectPool<SolrClient> solrPool;
   private ObjectPool<RestHighLevelClient> esPool;
 
+  private int getFileNumber(String fileName) {
```
lintool, line 615: hrm... maybe move this into part of C4Collection somewhere?

Doesn't really belong here...

shaneding: moved into `DocumentCollection` class

lintool (approved):

PR #1491, opened 2021-03-24
LTR features cleanup + unit test

stephaniewhoo:
+ LTR features clean up according to version described in the paper
+ Unit test added to bring back coverage


src/main/java/io/anserini/ltr/feature/DocumentFieldContext.java
```diff
@@ -274,6 +274,13 @@ public void generateBM25Stat(List<String> terms){
         double k1 = 0.9f;
         double b = 0.4f;
 
+        double avg_hmean_score = 0;
```
nsndimt, line 277: remove generateBM25Stat since we do not use any bm25stat feature?

lintool (approved):

PR #1496, opened 2021-03-31
LTR feature rename + docs

stephaniewhoo:
+ feature description list
+ rename features for consistency


docs/ltr-features.md
```diff
@@ -0,0 +1,90 @@
+# Features
+| Feature No.     | Description                                     |
+|-----------------|-------------------------------------------------|
+|      1          | [IBM Model1 in title\_unlemm](../src/main/java/io/anserini/ltr/feature/IBMModel1.java)
```
lintool, line 4: Based on my understanding, we specify computing the features on _which field_ separately, right?

So I think the table should only contain the features... and then in the Pyserini docs we say, we compute the scores over _x, y, z_ fields?

Also, let's _not_ number the features, because if we add, we'll need to renumber. Give the features names - similar to Java Class is fine, but let's refer to them uniquely.

src/main/java/io/anserini/ltr/feature/BM25Stat.java
```diff
@@ -99,7 +99,9 @@ public float postEdit(DocumentContext context, QueryContext queryContext) {
 
   @Override
   public String getName() {
-    return String.format("%s_%s_%s_BM25_k1_%.2f_b_%.2f",field, qfield, collectFun.getName(), k1, b);
+    String className = this.getClass().getName();
+    String name = className.substring(24,className.length());
```
lintool, line 103: I think there's a cleaner way to get just the class name without the prefix?

lintool (changes requested):
I'm really torn now... `DFR_GL2Stat` -> `DFRGL2Stat` looks terrible.

I think I would even prefer `DrfGl2Stat`? For example, `TfIdfStat` looks fine to me...

Thoughts? Even Huggingface does `BertTokenizer`... which is kinda ugly but consistent...

docs/ltr-features.md
stephaniewhoo: I use DFR_GL2 convention not jave consistent way in the doc, since I think this is easier for people to read? Or do you think we want DfrGl2 in doc as well?

lintool: In the table, `DFR_GL2` is good.

lintool (approved):

PR #1513, opened 2021-04-11
Calvin's replication

printfCalvin:
Run both experiments on Ubuntu in Windows. No problem occurred.


lintool:
Same issue here: "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

lintool (approved):

PR #1514, opened 2021-04-12
add pretokenized option in index and search collection

stephaniewhoo:


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -753,11 +754,12 @@ public Counters run() throws IOException {
         config = new IndexWriterConfig(germanAnalyzer);
       } else if (args.language.equals("es")) {
         config = new IndexWriterConfig(spanishAnalyzer);
-      } else if (args.language.equals("en_ws")) {
+      } else if (args.language.equals("en_ws") || args.pretokenized) {
```
lintool, line 757: hrm... not sure what `en_ws` is... I don't think it's a valid ISO code. Let's kill?

src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -84,6 +84,10 @@
           usage = "Path to file with stopwords.")
   public String stopwords = null;
 
+  @Option(name = "-pretokenized",
+          usage = "Path to file with stopwords.")
```
lintool, line 88: This isn't a path... it's a boolean option...

lintool (commented):
Initial comments

lintool:
@stephaniewhoo can you hook the test to https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/IndexReaderUtils.java#L436

so we can actually retrieve the doc from the index, and then you can iterate through and verify the exact contents...

might need modification to the test harness.

src/test/java/io/anserini/integration/PretokenizedIndexEndToEndTest.java
```diff
@@ -0,0 +1,70 @@
...
+                        "}"
+        ));
+
+        fieldNormStatusTotalFields = 1;
+        // whitespace analyzer keeps everything, includes docid
+        // this is ##a simple pretokenized test some time extra ##vert ##ing and intro 2000000 2000001
+        termIndexStatusTermCount = 15;
+        // this is ##a simple pretokenized test some|2 time|2 extra ##vert|2 ##ing|2 and intro 2000000 2000001  ?
+        //todo: check this field?
+        termIndexStatusTotFreq = 15;
```
stephaniewhoo, line 53: I don't see why this is 15...

stephaniewhoo:
> @stephaniewhoo can you hook the test to https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/IndexReaderUtils.java#L436
> 
> so we can actually retrieve the doc from the index, and then you can iterate through and verify the exact contents...
> 
> might need modification to the test harness.

I am thinking if this https://github.com/castorini/anserini/blob/d808d4a50ee69f44493e19a87dc36c7eb99402a9/src/test/java/io/anserini/integration/EndToEndTest.java#L195 has already checked the exact contents for us by using `document`? 
I also checked the index by document while debugging, but I found `termIndexStatusTotFreq` this field is a bit weird 


> @stephaniewhoo can you hook the test to https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/IndexReaderUtils.java#L436
> 
> so we can actually retrieve the doc from the index, and then you can iterate through and verify the exact contents...
> 
> might need modification to the test harness.

src/test/java/io/anserini/integration/PretokenizedIndexEndToEndTest.java
lintool: not sure either... what I generally do is make sure it's indexed correctly, and then just set the value to the number that comes out...

so at least the test case is guarding against the count changing due to a bug later on...

lintool:
I am thinking that we pull out the document and check token by token - not just the counts... we should check also if, for example, the `##` prefix is preserved.

stephaniewhoo:
> I am thinking that we pull out the document and check token by token - not just the counts... we should check also if, for example, the `##` prefix is preserved.

yes, I think we already have `assert` to check whether `contents` field of document is `this was ##a simple pretokeinzed test` by doing this ` assertEquals(documents.get(collectionDocid).get("contents"), IndexReaderUtils.documentContents(reader, collectionDocid))`.
But this is comparing two strings. Do you want a list of tokens from the document and to see if they are correct

src/test/java/io/anserini/integration/PretokenizedIndexEndToEndTest.java
```diff
@@ -0,0 +1,73 @@
...
+        // this is ##a simple pretokenized test some time extra ##vert ##ing and intro 2000000 2000001
+        termIndexStatusTermCount = 15;
+        termIndexStatusTotFreq = 15;
+        storedFieldStatusTotalDocCounts = 2;
+        termIndexStatusTotPos = 17 + storedFieldStatusTotalDocCounts;
+        storedFieldStatusTotFields = 6; // 1 docs * (1 id + 1 contents + 1 raw) *2
+    }
+
+    @Override
+    protected void setSearchGroundTruth() {
```
stephaniewhoo, line 63: do we want to compare every query tokens here?

lintool: you mean the doc tokens right?

yea, it's not that much to check, e.g.,

> id: 2000000
> "this", "was", "##a", "simple", "pretokenized", "test"

stephaniewhoo: oh, no. Sorry for the confusion. I think https://github.com/castorini/anserini/blob/b51f7a14bea31018057aa4907caf7d67657198e0/src/test/java/io/anserini/integration/EndToEndTest.java#L205 has checked every token in the doc along with its freq.

I mean since we also add -pretokenized option to search collection, say we have topics as `"1   ##vert ##ing`, we should expect while we processing this query, we did not stem or filter stopwords. Query tokens while we performing search should be something like `"##vert","##ing"` not `"vert","ing"`. Do we want to check this here? (I think this is the way to test the pretokenized option for search collection?)

lintool: yes, I think so?

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -700,6 +703,10 @@ private void loadQrels(String rf_qrels) throws IOException {
     return cascade.run(scoredFbDocs,  context);
   }
 
+  public Map<String,List<String>> getQueries(){
```
lintool, line 706: space after comma

src/test/java/io/anserini/integration/EndToEndTest.java
```diff
@@ -62,6 +64,8 @@
   protected String searchOutputPrefix = "e2eTestSearch";
   protected Map<String, String[]> referenceRunOutput = new HashMap<>();
   protected Map<String, Map<String, String>> documents = new HashMap<>();
+  protected Map<String, Map<String,Map<String,Long>>> tokens = new HashMap<>();
```
lintool, line 67: space after comma

src/test/java/io/anserini/integration/PretokenizedIndexEndToEndTest.java
```diff
@@ -0,0 +1,77 @@
...
+                        "  \"contents\" : \"this was ##a simple pretokenized test\"\n" +
+                        "}"));
+        documents.put("2000001", Map.of(
+                "contents", "some time extra ##vert ##ing and some time intro ##vert ##ing",
+                "raw","{\n" +
+                        "  \"id\" : \"2000001\",\n" +
+                        "  \"contents\" : \"some time extra ##vert ##ing and some time intro ##vert ##ing\"\n" +
+                        "}"
+        ));
+        tokens.put("2000000",Map.of("contents",
```
lintool, line 48: space after comma...

lintool:
lgtm now... ready for merging?

remove [WIP] when ready for final review?

lintool (approved):
minor linting comments, otherwise lg

src/test/java/io/anserini/integration/PretokenizedIndexEndToEndTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+
+public class PretokenizedIndexEndToEndTest extends EndToEndTest {
+
+    @Override
```
lintool, line 19: I just noticed - 2 space indent please.

lintool: And add boilerplate while you're at it.

PR #1528, opened 2021-04-27
MS MARCO doc/passage replication

saileshnankani:
Ran both experiments on MacOS. No problem occurred.


MXueguang:
@saileshnankani Thanks for the reproduction. Just to confirm, all the results are exactly same to the numbers in the documents?

saileshnankani:
> @saileshnankani Thanks for the reproduction. Just to confirm, all the results are exactly same to the numbers in the documents?

Yes, all the numbers were exactly the same as in the provided READMEs.

lintool (approved):

PR #1542, opened 2021-05-07
Replicate MS MARCO passage and document dataset

Albert-Ma:


lintool:
Can you confirm that you are able to replicate the results?

lintool (approved):

PR #1549, opened 2021-05-27
Changed SearchCollection output writer to UTF-8 encoding and updated FEVER experiments

kelvin-jiang:
Updated FEVER experiment documentation to use `SearchCollection` accordingly.

### DIFF
--- a/src/main/python/fever/evaluate_doc_retrieval.py
+++ b/src/main/python/fever/evaluate_doc_retrieval.py
@@ -59,7 +59,7 @@ def evaluate_retrieval(args):
         curr_query = -1
         predicted_docs = []
         for i, line in enumerate(f):
-            query_id, doc_id, rank = line.strip().split('\t')
+            query_id, _, doc_id, rank, _, _ = line.strip().split(' ')
             query_id = int(query_id)
 
             # if we reach a new query in the run file, evaluate the previous query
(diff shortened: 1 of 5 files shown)


lintool (approved):

src/main/python/fever/evaluate_doc_retrieval.py
```diff
@@ -59,7 +59,7 @@ def evaluate(query_id, predicted_docs):
         curr_query = -1
         predicted_docs = []
         for i, line in enumerate(f):
-            query_id, doc_id, rank = line.strip().split('\t')
+            query_id, _, doc_id, rank, _, _ = line.strip().split(' ')
```
ronakice, line 62: Can't we just go with `line.strip().split()`

ronakice (approved):
@kelvin-jiang small comment

src/main/python/fever/evaluate_doc_retrieval.py
kelvin-jiang: that should work too, i guess what's currently there is more explicit about what exactly we're splitting on

ronakice: Yeah, I see people arbitrarily picking between tabs/spaces in this case so that's why I was wondering if we are consistent here it is fine.

PR #1555, opened 2021-06-02
Replicated BM25 baselines for MS MARCO passage and document ranking

jpark621:
Environment: 2020 Macbook Pro, macOS Big Sur, 8GC RAM, 500 GB HD

All results were replicated.


lintool (approved):
Great, thanks!

PR #1560, opened 2021-06-12
Add ability to index custom vectors

lintool:
Behind the scenes, Anserini transparently generates pseudo-documents, and then also uses sum-of-impact for scoring.


lintool:
Here's an overview of the implementation:
+ `JsonVectorCollection` transparently takes the user-defined vector and creates a pseudo-document. If a feature has weight _n_, it is repeated _n_ times. With whitespace tokenization, each feature name gets a tf equal to _n_.
+ During indexing, we make sure we turn off "norms" in Lucene (i.e., norm = 1) - in BM25, this is what computes doc length normalization. This is done in `ImpactSimilarity`.
+ At search time, the search scores is simply the sum of the tf values (which are the vector values specified by the user). This is also done in `ImpactSimilarity`.

MXueguang (approved):
lgtm

PR #1563, opened 2021-06-15
Added tests for JsonVectorCollection

lintool:
Along the way, I discovered a code path in `JsonCollection` that is never taken, and hence can be killed.


MXueguang (approved):

lintool:
CI appears to be borked... probably due to this:

> Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.

I will override and merge for now. Will try to sort out later.

Issue #1570, opened 2021-06-23
RM3 only works for English


cash:
This is related to the JHU workshop.

I have a corpus that is not in English, but the Rm3Reranker has this line in it:

```java
if (!term.matches("[a-z0-9]+")) continue;
```
Would it be possible to have an argument on the constructor that either turns this off or supports passing in a custom filter?  For our use cases, we're probably fine with not having any filter, but since this is here it must have boosted performance on some data set and so I could see other use cases for custom filtering.

lintool:
Hrm. Thanks for pointing that out. Turning off the filter is easy, but the tricky bit is figuring out how to expose it...

My initial reaction is to add a new option in `SearchArgs`: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SearchArgs.java#L265

Something like `-rm3.termFilter` set to true by default?

Thoughts?

cash:
I agree that adding a boolean flag is the easiest way to turn filtering on or off. Having it default to true is necessary to not break anything. The name of the flag is reasonable. I can create a PR for this.

How do you handle propagating changes to pyserini? Do you limit changes to pyserini to what is exposed in the most recent release of anserini or do you work with snapshots and synchronize releases?

lintool:
> I agree that adding a boolean flag is the easiest way to turn filtering on or off. Having it default to true is necessary to not break anything. The name of the flag is reasonable. I can create a PR for this.

Thanks!

> How do you handle propagating changes to pyserini? Do you limit changes to pyserini to what is exposed in the most recent release of anserini or do you work with snapshots and synchronize releases?

We build anserini and physically copy over the fat-jar - see instructions on a dev copy here:
https://github.com/castorini/pyserini/#development-installation

To expose to pyserini you'll need to expose the option in `SimpleSearcher` - something like this:
https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SimpleSearcher.java#L318

lintool:
Closed by #1580

PR #1580, opened 2021-06-28
adds boolean flag to control ascii filtering (#1570)

cash:
This adds a boolean parameter `filter_Terms` to Rm3Ranker where true means removing terms that have a character outside [a-z0-9].

It exposes this parameter through SearchArgs as `-rm3.noTermFilter` with a default of false. This was done so that it would not change current usage. 

I looked at writing a unit test, but they weren't any currently for reranking and it would be quite involved. I tested with a Russian collection and confirmed that with filtering turned on the query was expanded with numbers only and with it turned off it was expanded with Russian words.

I did not modify the tag() method to include this flag. I wasn't sure how tag() was being used, but it does not seem like filtering is a parameter that one would tune. It will either be right or wrong based on the collection.


lintool:
hi @cash thanks for the PR. We have a detailed set of regressions, which I'll run now: https://github.com/castorini/anserini/blob/master/docs/regressions.md

Will merge after successful run.

lintool (approved):
lg

PR #1589, opened 2021-07-01
Update max_docs in perform_fusion()

lhbonifacio:
max_docs parameter in perform_fusion() was set as 10000, fixed it to use the correct variable


lintool (approved):

lintool:
Thanks for the fix, but note that fusion has been directly integrated into Pyserini: https://github.com/castorini/pyserini/tree/master/pyserini/fusion

So, this script is not being used anymore, except for legacy purposes.

PR #1592, opened 2021-07-03
Simple impact searcher

MXueguang:


MXueguang:
for PR in pyserini, please see https://github.com/castorini/pyserini/pull/684

src/main/java/io/anserini/search/SimpleImpactSearcher.java
```diff
@@ -0,0 +1,425 @@
...
+     * Batch version of {@link #searchFields(String, Map, int)}.
+     *
+     * @param queries list of queries
+     * @param qids list of unique query ids
+     * @param k number of hits
+     * @param threads number of threads
+     * @param fields map of fields to search with weights
+     * @return a map of query id to search results
+     */
+    public Map<String, Result[]> batchSearchFields(List<Map<String, Float>> queries, List<String> qids, int k, int threads,
```
lintool, line 173: Let's not implement fielded search yet until we have a realistic use case? Otherwise, it's dead code that's untested...

pom.xml
```diff
@@ -129,6 +129,10 @@
               <mainClass>io.anserini.search.SimpleTweetSearcher</mainClass>
               <id>SimpleTweetSearcher</id>
             </program>
+            <program>
+              <mainClass>io.anserini.search.SimpleImpactSearcher</mainClass>
```
lintool, line 133: I think we should remove this - see more below...

src/main/java/io/anserini/search/SimpleImpactSearcher.java
```diff
@@ -0,0 +1,324 @@
...
+    }
+  
+    // Note that this class is primarily meant to be used by automated regression scripts, not humans!
+    // tl;dr - Do not use this class for running experiments. Use SearchCollection instead!
+    //
+    // SimpleImpactSearcher is the main class that exposes search functionality for Pyserini (in Python).
+    // As such, it has a different code path than SearchCollection, the preferred entry point for running experiments
+    // from Java. The main method here exposes only barebone options, primarily designed to verify that results from
+    // SimpleImpactSearcher are *exactly* the same as SearchCollection (e.g., via automated regression scripts).
+    public static void main(String[] args) throws Exception {
```
lintool, line 320: why even have a main? let's kill?

lintool:
@MXueguang is this ready for review? we need detailed tests for this since it's going to be at the core of lots of things moving forward...

pom.xml
MXueguang: seems if we remove this, it will not be included in fatjar?

MXueguang: seems if we remove this, it will not included in fatjar?

lintool: Really? I guess I might have forgotten this before? Where is this documented?

lintool:
We need some tests to make sure that `SearchCollection` and `SimpleImpactSearcher` give _exactly_ the same output. Should this be done as a regression test in pyserini?

Currently, do we get _exactly_ the same as the uniCOIL and DeepImpact repro in Anserini?

If so, ready to merge?

MXueguang:
yes, ready to merge.

lintool (approved):

PR #1596, opened 2021-07-09
C4 No Clean Collection

MarcosFP97:
Hi!

I have implemented a new collection able to read the no clean C4 Collection used in the [TREC Health 2021 Misinformation Track](https://trec-health-misinfo.github.io/). 

The main difference with the C4 Collection resides in the ability to read the docno from the JSON or to generate it automatically under the specs of the track.

I have also done two different tests, one with docno as a JSON field and another without it. Hope you find this useful :)


MarcosFP97:
Hi again, 

I have just realized that another `Generator` class is also needed. Maybe some coding could be saved by inheriting `NoCleanC4Collection` from `C4Collection` class, but I wanted a first solution that worked fine, and I didn't know which solution followed best Anserini's philosophy.

Another option could be renaming these new classes to TREC 2021 Health Misinfo since the docno are custom generated for this task. I am open to any suggestions :)

src/main/java/io/anserini/collection/NoCleanC4Collection.java
```diff
@@ -0,0 +1,179 @@
...
+    private String contents;
+    private String raw;
+    private String url;
+    private long timestamp;
+
+    public Document(JsonNode json, String filename, int jsonLoc) {
+      this.raw = json.toPrettyString();
+      this.contents = json.get("text").asText();
+      
+      try{
```
lintool, line 136:
```suggestion
      try {
```

src/main/java/io/anserini/collection/NoCleanC4Collection.java
```diff
@@ -0,0 +1,179 @@
...
+        this.id = json.get("docno").asText();
+      }catch(Exception e){
```
lintool, line 138:
```suggestion
      } catch(Exception e) {
```

lintool:
(1) Seems like there's a lot of code duplication. Why not just subclass `C4Collection`?
(2) Can we rename to `C4NoCleanCollection`? Lexicographically, this provides a nice sort order putting `C4` and `C4NoClean` together.

MarcosFP97:
Hi @lintool,

(1) Yes, of course. I just wanted to have one first functional version, but I have made some new commits towards this "clean code" idea. 
The problem is that I am finding myself with the following error in both tests and I don't understand why:
```
[ERROR] C4NoCleanCollectionWithDocNoTest.java:[30,18] incompatible types: io.anserini.collection.C4NoCleanCollection cannot be converted to io.anserini.collection.DocumentCollection<io.anserini.collection.C4NoCleanCollection.Document>
```
(2) Done! :)

PS: My apologies in advance if something is not 100% as it should be, I tried to follow the philosophy of the pre-existing code (inner classes, static methods, etc.).

Best,
Marcos

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -80,8 +80,8 @@ private int getFileNumber(String fileName) {
   public static class Segment extends FileSegment<C4Collection.Document>{
     private MappingIterator<JsonNode> iterator; // iterator for JSON line objects
     private JsonNode node = null;
-    private String filePath;
-    private String fileName;
+    private static String filePath;
```
lintool, line 83: I don't think you want static here? `protected`?

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -120,10 +120,18 @@ public void readNext() throws NoSuchElementException {
...
+    }
+
+    public static void setFileName(String filename){
+      fileName = filename;
+    }
   }
 
   public static class Document implements SourceDocument {
-    private String id;
+    private static String id;
```
lintool, line 134: same

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -171,5 +179,9 @@ public String raw() {
     public boolean indexable() {
       return true;
     }
+
+    public static void setId(String idS){
+      id = idS;
```
lintool, line 184: common Java idiom is
`this.id = id` = okay to have the method arg have the same name.

src/main/java/io/anserini/collection/C4Collection.java
MarcosFP97: Yes, but I didn't want to change attribute visibility of previously developed code without first consulting with you. I have already changed them :)

MarcosFP97:
Thanks for your assistance @lintool! So now I believe that the only aspect left is to identify why the `C4NoCleanCollectionTest.java` and `C4NoCleanCollectionWithDocNoTest.java`, are throwing this exception:

```
[ERROR] C4NoCleanCollectionWithDocNoTest.java:[30,18] incompatible types: io.anserini.collection.C4NoCleanCollection cannot be converted to io.anserini.collection.DocumentCollection<io.anserini.collection.C4NoCleanCollection.Document> 
```

Since  I don't identify anything strange there. I am just trying to create a test for the new class in the same way as the rest of the `Collection` classes. Any thoughts?

MarcosFP97:
I have made a change: in the new test classes instead of inheriting from `DocumentCollectionTest<C4NoCleanCollection.Document>`, the exception disappears if we inherit from `DocumentCollectionTest<C4Collection.Document>`. 

However, I am not sure if this is correct, and the assertion fails. I am a bit lost at this point with this error... I would appreciate any help!

MarcosFP97:
Hi again @lintool,
Good news! I managed to solve it. The problem was that I was trying to create my Segment class in the subclass and this wasn't necessary (it was never used indeed). The most elegant manner that I found to solve this was to check the type of the instance in the upper class and call the corresponding Document. In this way, tests are perfectly working :) (like in my original version, but without code duplication).
Tomorrow I will try to index a toy example to check that the C4Generator works fine with the new class. I let you know once it is finished :)

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -77,19 +77,24 @@ private int getFileNumber(String fileName) {
...
+    protected String fileName;
+    protected int count = 0;
 
     public Segment(Path path) throws IOException {
       super(path);
       filePath = path.toString();
       int fileNumStart = filePath.indexOf("c4-train.") + 9;
-      // plus one to remove leading zero
-      fileName = filePath.substring(fileNumStart + 1, fileNumStart + 5);
+      if(C4Collection.this instanceof C4NoCleanCollection){
```
lintool, line 91: This should be in the subclass...

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -111,7 +116,12 @@ public void readNext() throws NoSuchElementException {
       if (node == null) {
         throw new NoSuchElementException("JsonNode is empty");
       } else {
-        bufferedRecord = new C4Collection.Document(node, fileName, count);
+        if(C4Collection.this instanceof C4NoCleanCollection){
```
lintool, line 119: same

src/test/java/io/anserini/collection/C4NoCleanCollectionWithDocNoTest.java
```diff
@@ -0,0 +1,56 @@
+/*
```
lintool, line 1: `WithDocNo` should probably be `WithDocno`?

src/main/java/io/anserini/collection/C4Collection.java
```diff
@@ -77,19 +77,24 @@ private int getFileNumber(String fileName) {
...
 
     public Segment(Path path) throws IOException {
       super(path);
       filePath = path.toString();
       int fileNumStart = filePath.indexOf("c4-train.") + 9;
-      // plus one to remove leading zero
-      fileName = filePath.substring(fileNumStart + 1, fileNumStart + 5);
+      if(C4Collection.this instanceof C4NoCleanCollection){
+        // plus one to remove leading zero
+        fileName = filePath.substring(fileNumStart + 1, fileNumStart + 14);
```
lintool, line 93: this can be much cleaner with a regular expression...

src/test/java/io/anserini/collection/C4NoCleanCollectionWithDocNoTest.java
MarcosFP97: Yes, done

src/main/java/io/anserini/collection/C4Collection.java
MarcosFP97: Done! My apologies, I didn't realize that I was necessary to override one method in the subclass in order for the pipeline to work properly.

src/main/java/io/anserini/collection/C4Collection.java
MarcosFP97: Done too :)

MarcosFP97:
Ok, so I think that if you approve it, the code is ready to merge :) The tests are working, and I have also tried to index the "toy examples" with the following commands:

```
nohup sh target/appassembler/bin/IndexCollection -collection C4NoCleanCollection -input ~/anserini/src/test/resources/sample_docs/c4_noclean_with_docno -index indexes/C4_TEST -generator C4Generator -threads 44 -storePositions -storeDocvectors -storeRaw >& logs/log.C4_TEST &
```
```
nohup sh target/appassembler/bin/IndexCollection -collection C4NoCleanCollection -input ~/anserini/src/test/resources/sample_docs/c4_noclean -index indexes/C4_TEST -generator C4Generator -threads 44 -storePositions -storeDocvectors -storeRaw >& logs/log.C4_TEST &
```

And they worked perfectly. Looking forward for final comments @lintool :)

ronakice (approved):
LGTM! @lintool let's merge this, so Arthur Chen and I can make changes.

MarcosFP97:
Thank you very much for your approval @ronakice! 

Looking forward for the merge :)

lintool (approved):

PR #1600, opened 2021-07-16
UniCOIL reproduced by ArvinZhuang

ArvinZhuang:
Got the same MRR@10 score by using provided collection and queries.
More than that, I actually followed the uniCOIL repo (https://github.com/luyug/COIL/tree/main/uniCOIL) retrained the model with a different seed, and recreated the collection and queries, I got the MRR@10 score of 0.350, pretty close, I'm happy with that.

One question, I notice that uniCOIL retrieval is much slower than bm25, it actually took me more than 1.5 hours on my old laptop. Any insight why this is the case? My guess is because of the small vocab size and super long average posting list?


MXueguang:
Hi @ArvinZhuang,
Thank you very much for the replication.

There are some reasons can cause it slower I guess.
1. vocab size is smaller, bert-base-cased has ~30k tokens while BM25 usually uses analyzers gives 500k vocab.
2. the posting list could be longer but not super long, since it should be the same as BM25+doc2query.
3. Since we are creating the pseudo doc, the term frequency distribution is very different than regular docs. Fail to get optimized in Lucene could be an issue make things slow.

On my machine, SSD + i7 single thread, it took around 15~20 min, if I do multithreads in [Pyserini](https://github.com/castorini/pyserini/blob/master/docs/experiments-unicoil.md), it just took 3-4 minutes.

ArvinZhuang:
Hi @MXueguang,
Thanks for the feedback! I'm curious about whats the doc distribution of those frequent special subtokens such as "##s"? It is likely lots of docs have this subtoken right? same for queries.  If this is the case, then it makes sense that many queries, unfortunately, have to iterate millions of docs.

ArvinZhuang:
Also for the stopwords such as "is", I think anserini bm25 removes stopwords by default right? but bert tokenizer doesn't.

MXueguang:
yes, I guess some post process,( e.g. filter out stopwords, remove extremely large df token) can speed up the retrieval without influence the effectiveness too much.

lintool:
@ArvinZhuang thanks for your replication and we're glad this has worked out for you!

lintool (approved):

PR #1603, opened 2021-07-21
Added support for MS MARCO doc v2 segmented corpus

lintool:


src/main/java/io/anserini/collection/MsMarcoDocV2Collection.java
```diff
@@ -131,12 +131,21 @@ public String id() {
...
+      // This difference allows us to easily figure out which version we're processing,
+      // and to (potentially) process them differently (although we're not doing so currently).
+      if (!fields.containsKey("url") || !fields.containsKey("title") || !fields.containsKey("headings") ||
+              (!fields.containsKey("body") && !fields.containsKey("segment")) ) {
         throw new RuntimeException("Document is missing required fields!");
       }
 
-      return fields.get("url") + " " + fields.get("title") + " " + fields.get("headings") + " " + fields.get("body");
+      if (fields.containsKey("body")) {
+        return fields.get("url") + " " + fields.get("title") + " " + fields.get("headings") + " " + fields.get("body");
```
justram, line 145: Would we circle back to use options to control the fields to be encoded like Pyserini's encoding scheme?
https://github.com/castorini/pyserini/blob/eb05f1426ebef6f0e477413ca13a35f7397b5501/pyserini/encode/__main__.py#L97-L99

lintool: I shared the results of indexing different fields in our internal Slack channel - and the tl;dr is that indexing everything is still the best. Those experiments were done in a janky way - by changing the code physically each time and re-compiling.

Yes, in principle it would be good to be able to specify this... but it's unclear how we would pass in the information in cmdline params... any ideas?

justram: Yeah, everything looks great, and we can merge it now, I think.
Just wondering if we need to leave a note on the experiment.md and some comments in line 144 like TODO: ... for future reference.

MXueguang: I'll try to figure out a way to handle this format in pyserini.encode
we can merge it for now.

MXueguang: yeah I think we can merge it for now.

MXueguang (approved):

src/main/java/io/anserini/collection/MsMarcoDocV2Collection.java
jacklin64: Yes, it looks good to me.

jacklin64 (approved):
Although now we direct add all the field in the code, I also think in the future, we should add option for flexibility. For example, when running passage index, the text field will be "passage" instead of "segment".

src/main/java/io/anserini/collection/MsMarcoDocV2Collection.java
lintool: Okay, I added notes in the code and documentation.

PR #1615, opened 2021-08-11
use 7-char sha1

t-k-:


t-k-:
The later minor change has some conflict, I think it is better off creating another branch from my end, so just closed this one.

New PR: https://github.com/castorini/anserini/pull/1616

PR #1619, opened 2021-08-18
whitespace analyzer for sw and te

MXueguang:


src/main/java/io/anserini/index/IndexCollection.java
```diff
@@ -805,6 +805,8 @@ public Counters run() throws IOException {
         config = new IndexWriterConfig(turkishAnalyzer);
       } else if (args.language.equals("zh") || args.language.equals("ko")) {
         config = new IndexWriterConfig(chineseAnalyzer);
+      } else if (args.language.equals("sw") || args.language.equals("te")) {
```
lintool, line 808:
```suggestion
      } else if (args.language.equals("sw") || args.language.equals("te")) {
        // For Mr.TyDi: sw and te do not have custom Lucene analyzers, so just use whitespace analyzer.
```

src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -281,6 +282,8 @@ public void setLanguage(String language) {
       this.analyzer = new TurkishAnalyzer();
     } else if (language.equals("zh") || language.equals("ko")) {
       this.analyzer = new CJKAnalyzer();
+    } else if (language.equals("sw") || language.equals("te")) {
```
lintool, line 285: Add comment, same as above?

lintool (approved):

PR #1648, opened 2021-10-14
Change msmarco v2 name

MXueguang:


lintool:
Interesting. We don't actually tests the Qrels enum... we probably should.

PR #1658, opened 2021-10-18
Fix score overflow issue in ScoreTiesAdjusterReranker for SPLADEv2

lintool:
The method of rounding used in ScoreTiesAdjusterReranker causes overflow issues
because SPLADEv2 scores can get very big. Doesn't affect the actual ordering of
docs, hence everything is fine with MS MARCO output format, but does cause issues
with TREC format and trec_eval.

With this bug fixed, added SPLADEv2 regressions.


lintool:
hi @MXueguang I've asked @JMMackenzie to take a look since we've discussed this on Slack.

However, another pair of 👀 on this wouldn't hurt either. Thanks!

PR #1678, opened 2021-11-19
Add intra-run multithreading to SearchCollection

lintool:
Context: `SearchCollection` currently has the ability to do parameter sweeps, with different configurations in parallel using a thread pool. However, it currently has no ability to run different queries in a single configuration in parallel. This means that some sparse models take a long time to run.

This PR adds a threadpool within an individual configuration to run queries in parallel.


lintool:
All regressions successfully passed with the multi-threaded implementation.

MXueguang (approved):

PR #1679, opened 2021-11-19
Add uniCOIL noexp zero-shot results on MS MARCO V2 and DL 21

lintool:


lintool:
@MXueguang - I'm waiting for encoded queries for dev2 and DL 21. I'll add and update I get them from you.

lintool:
@MXueguang @ronakice This is ready for review.

I reorganized the main README entry point. I forgot that each regression page can have multiple query sets, so we don't need separate pages for dev and dev2 on MS MARCO V2. This makes our README cleaner.

MXueguang (approved):

Issue #1680, opened 2021-11-20
Clean up of regression pipeline


lintool:
Some notes for myself:

+ `name` in the yaml should actually be corpus
+ probably should simplify the way we construct the name of run files
+ docgen could be simplified wrt pointers to yaml and actual template

lintool:
+ KILT regression needs a better name

lintool:
All done (for now)!

PR #1690, opened 2021-12-07
Refactor regression framework

lintool:
Work on #1680


lintool:
@MXueguang this is ready for review. It's a huge patch, but I've verified regressions, so a quick sanity check will be sufficient, I think.

MXueguang (approved):

PR #1708, opened 2021-12-17
More regression refactoring

lintool:
More work on #1680


lintool:
@MXueguang This is a huge patch, but most changes are cosmetic, so a quick sanity check will suffice. I have verified regressions still work.

MXueguang (approved):

PR #1710, opened 2021-12-18
Cleanup build warnings

lintool:
hey @stephaniewhoo can you take a look? some of this is your code, I think.

> Map<String, Object> json = new HashMap();

Just suppressing may not be the cleanest collection... is there a more appropriate specialized type than `Object`?

---

Otherwise, fixed a bunch of minor build warnings.


stephaniewhoo:
The Object may be `String` or `List<String>`, so I am afraid that it cannot be further specified.

lintool:
Okay, in that case can you do a CR and give me a 🚢 it of it looks good?

stephaniewhoo (approved):
Lgtm

PR #1717, opened 2021-12-28
Final round of refactoring of regression framework

lintool:
Final installment of #1680 

+ Refactored regression script, cleaned up dead code.
+ Renamed YAML params.


lintool:
@MXueguang This is ready for review. 

I've confirmed that regressions still work; most of the changes are cosmetic, so a quick sanity check will suffice.

Most of the substantive changes are in the regression script `src/main/python/run_regression.py`.

MXueguang (approved):

PR #1722, opened 2022-01-07
Reproduction of MS MARCO passage and doc ranking

vivianliu0:
Reproduced results from MS MARCO passage and doc ranking exercises as part of initial [screening exercise](https://github.com/lintool/guide/blob/master/ura.md)

Was able to reproduce the expected values, below the issues I encountered:
- For the MS MARCO passage ranking, using the provided alternative mirror was much faster (approximately 30 min compared to 2h+ suggested completion time through msmarco.blob.core.windows.net) to download the MS MARCO passages.
- Minor issues: since I was working on a Windows machine, there were small differences in the commands such as using backslashes for file paths, and gzip rather than gunzip for decompression.

Environment Setup:
Windows 10 Pro Version 20H2
Java: openjdk 11.0.9.1
Python: 3.8.4


lintool (approved):

lintool:
@vivianliu0 actually, same comment as with your pyserini PR... I think the date is wrong? Also, can you please resolve conflicts? Thanks!

PR #1731, opened 2022-01-13
Fix bug in run_regression re: scoping of variables

lintool:
Interesting bug. `args.dry_run` is not in scope within the function `run_search`, but on Python 3.6 this works... whereas _not_ on Python 3.8.


ronakice (approved):
LGTM! Interesting this happens!

MXueguang (approved):

lintool:
Apparently, `__main__` counts a global scope: https://stackoverflow.com/questions/4775579/main-and-scoping-in-python

Did not know this...

PR #1756, opened 2022-02-07
Initial Screen

jh8liang:


lintool (changes requested):
This does not appear to be a valid URL:
https://github.com/castorini/pyserini/commit/5cdf9ec3600bf693cf8d80a9c869444b4e29ff75

I think you meant `anserini` here?

Also, please note your machine specs (OS, etc.). Also indicate if you were successful or if you had any issues.

PR #1766, opened 2022-02-14
Geosearch test

d1shs0ap:
Files in resources/sample_docs/geosearch:
greenland_rivers.wkt: this file isn’t used in any of the tests. It is used to test the file_to_wkt.py script--it converted the HydroRIVER dataset (shp files) for Greenland into the form of a wkt file. The HydroRIVER dataset is omitted from sample_docs.
grand_river.wkt is converted from grand_river.geojson
Other wkt files were written by me to test different functionalities of the Lucene API
Tests (used SimpleWKTShapeParser to load files):
Included 1 test for each of the relevant shapes: LineString, MultiLineString, Polygon, MultiLinePolygon
Included a test for a polygon with “holes” in it
Included a test for Lake Ontario, loaded from GeoJson
Included a test for a segment of Grand River, equivalent to the following:
FYI: There is a bijection between the “relevant” Lucene types and WKL types:
LineString = Line, MultiLineString = Line[]
Polygon = Polygon (supports holes), MultiPolygon = Polygon[]
As a interesting side note, seems like Kibana doesn’t support the latest Lucene “Contains” query feature yet (released in 8.5.0), where we search for polygons containing a certain point/line/polygon


src/test/java/io/anserini/GeoIndexerTestBase.java
```diff
@@ -0,0 +1,162 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.text.ParseException;
+import java.util.List;
+
+
+public class GeoIndexerTestBase extends LuceneTestCase {
+    protected Path tempDir1;
```
lintool, line 41: House style - two space indent. Please reformat everything?

src/test/java/io/anserini/GeoIndexerTestBase.java
```diff
@@ -0,0 +1,162 @@
...
+        try {
+            Directory dir = FSDirectory.open(tempDir1);
+
+            IndexWriterConfig config = new IndexWriterConfig();
+            config.setOpenMode(IndexWriterConfig.OpenMode.CREATE);
+
+            IndexWriter writer = new IndexWriter(dir, config);
+
+
+            Document doc1 = new Document();
```
lintool, line 53: Before each of these blocks, can you please add a simple comment answer the question: what feature is this illustrating?

src/test/java/io/anserini/search/GeoSearcherTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+public class GeoSearcherTest extends GeoIndexerTestBase {
+
+    @Test
+    public void testGetLakeOntarioGeoJson() throws Exception {
+        Directory directory = FSDirectory.open(tempDir1);
+        DirectoryReader reader = DirectoryReader.open(directory);
+        IndexSearcher searcher = new IndexSearcher(reader);
+
+        Query q = LatLonShape.newBoxQuery("geometry", ShapeField.QueryRelation.INTERSECTS,
```
lintool, line 23: We generally don't mind wide lines - I think this will look better on a single line. Same below?

lintool (changes requested):
Let's not check `sample_docs/geosearch/greenland_rivers.wkt` into the repo... too large... adds to bloat. What if we did a subset of the file, e.g., first 100 lines? Or enough just for testing?

Minor comments, mostly pedantic.

Great job!

d1shs0ap:
@lintool I think in that case, it makes sense to remove `sample_docs/geosearch/greenland_rivers.wkt` altogether. It isn't used in any of the Java test cases since its sole purpose was to test the Python script. But the results of the Python script is already shown with `grand_river.geojson` and `grand_river.wkt`, so it may be redundant to do it again over `greenland_rivers`--let me know what you think.

lintool:
> @lintool I think in that case, it makes sense to remove `sample_docs/geosearch/greenland_rivers.wkt` altogether. It isn't used in any of the Java test cases since its sole purpose was to test the Python script. But the results of the Python script is already shown with `grand_river.geojson` and `grand_river.wkt`, so it may be redundant to do it again over `greenland_rivers`--let me know what you think.

Agreed. Let's remove `sample_docs/geosearch/greenland_rivers.wkt`.

lintool (approved):

PR #1774, opened 2022-02-23
MS MARCO reproduction tasks

mayankanand007:
System information: Apple M1 Silicon chip, 16GB RAM, JDK-17 (compatible with M1) / Java 17.0.2

Notes: No issues in build, and able to replicate all results with reference times same as given in the experiment description.


lintool:
Nice! I think this is our first documented repro on Apple Silicon. JDK-17 sounds so futuristic.

lintool (approved):

PR #1776, opened 2022-02-24
Add HydroRIVERS index feature

d1shs0ap:
Changes:
- `GeoGenerator.java`: generator for geo documents
- `JsonCollectionGeoRiverTest.java`: tests if JsonCollection works on the data prepared for HydroRIVERS
- `GeoGeneratorTest.java`: tests for `GeoGenerator.java`
- `rivers.json`: test data used in `JsonCollectionGeoRiverTest.java`

Ran the indexer:
```
sh target/appassembler/bin/IndexCollection -threads 1 -collection JsonCollection \
 -generator GeoGenerator -input collections/hydrorivers \
 -index indexes/hydrorivers -storeRaw 
```
on **all** HydroRIVERS data (8.5 million rivers) with no problems. However, haven't tried querying on the indexed data yet, will do so after building the search class.


src/main/java/io/anserini/index/generator/GeoGenerator.java
```diff
@@ -0,0 +1,85 @@
...
+          for (Field f: fields) {
+            doc.add(f);
+          }
+        } catch (ParseException | IOException e) {
+          LOG.error("Error parsing unknown shape using SimpleWKTShapeParser: " + v);
+        } catch (IllegalArgumentException e) {
+          LOG.error("Error casting shape to any of the types Line, Line[], Polygon, Polygon[]: " + v);
+        }
+
+      } else {
```
lintool, line 65: We can circle back in a subsequent patch, but the cleaner way to handle is to build a specific `HydroRiversGenerator` that subclasses `GeoGenerator` - to take care of the "other" fields?

src/main/java/io/anserini/index/generator/GeoGenerator.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.logging.log4j.Logger;
+import org.apache.lucene.document.*;
+import org.apache.lucene.geo.Line;
+import org.apache.lucene.geo.Polygon;
+import org.apache.lucene.geo.SimpleWKTShapeParser;
+
+import java.io.IOException;
+import java.text.ParseException;
+
+public class GeoGenerator implements LuceneDocumentGenerator<JsonCollection.Document> {
```
lintool, line 15: Is this better named `WktGenerator`? Or maybe `WktGeometryGenerator`? Thoughts?

src/test/java/io/anserini/collection/JsonCollectionGeoRiverTest.java
```diff
@@ -0,0 +1,104 @@
...
+            Map.entry("ORD_FLOW", "8"),
+            Map.entry("HYBAS_L12", "9120016580"),
+            Map.entry("geometry", "LINESTRING (-29.737500000000722 83.54583333333295, -29.731250000000642 83.55208333333294, -29.731250000000642 83.57291666666629)"),
+            Map.entry("id", "90000003")
+    ));
+  }
+
+  @Override
+  void checkDocument(SourceDocument doc, Map<String, String> expected) {
+    // Note that we need an id in addition to HYRIV_ID to distinct between different docs
```
lintool, line 85: to distinct -> to distinguish

src/main/java/io/anserini/index/generator/GeoGenerator.java
lintool: If you do this, then I don't think you need to duplicate `HYRIV_ID` in the document; the generator can just create both fields.

lintool (approved):
I love how this fits nicely into the existing infra, and we get all the features of `IndexCollections` (e.g., multi-threading) for free.

src/main/java/io/anserini/index/generator/GeoGenerator.java
d1shs0ap: Actually, it used to be named `WktGenerator`. But in order to make the raw data fit with the JsonCollection, the format is no longer `wkt`. It is more like a custom `json` file instead.

src/main/java/io/anserini/index/generator/GeoGenerator.java
d1shs0ap: This is something I was also bouncing back and forth with. Perhaps we could decide later when we deal with HydroBASINS and HydroLAKES?

Issue #1777, opened 2022-02-24
Add missing bindings to missing uniCOIL queries for DL


lintool:
We currently only have:

```
  TREC2019_DL_DOC(TsvIntTopicReader.class,"topics-and-qrels/topics.dl19-doc.txt"),
  TREC2019_DL_PASSAGE(TsvIntTopicReader.class,"topics-and-qrels/topics.dl19-passage.txt"),
  TREC2020_DL(TsvIntTopicReader.class,"topics-and-qrels/topics.dl20.txt"),
  TREC2021_DL(TsvIntTopicReader.class,"topics-and-qrels/topics.dl21.txt"),
```

Let's go through and add missing ones.

@MXueguang can you please do this?

PR #1779, opened 2022-02-24
Add unicoil encoded DL topics

MXueguang:
to close https://github.com/castorini/anserini/issues/1777
add unicoil encoded dl topics


lintool (changes requested):
Please add test cases also?

And just a clarification - noexp for DL19/20 - we need to separately build regressions for those, right?

MXueguang:
> And just a clarification - noexp for DL19/20 - we need to separately build regressions for those, right?
yes.

Ah I'll add test cases in this PR

lintool (approved):

PR #1781, opened 2022-02-27
Hydrorivers search

d1shs0ap:
Changes:
- `GeoGenerator.java`, `GeoGeneratorTest.java`: Updated GeoGenerator to also include a string Id field to stay consistent with the `SimpleSearcher.Result` class
- `SimpleGeoSearcher.java`: Bridge searcher class for GeoSearch
- `GeoSearcherTest.java`: Tests the bridge class
- `GeoSearchExplorationTest.java`: This file was renamed from `GeoSearcherTest.java` to `GeoSearchExplorationTest.java`. It comprises of the tests I added before from the Lucene Geo API exploration
- `GeoIndexerTestBase.java`: Updated the indexer base to also include a string Id field


src/main/java/io/anserini/search/SimpleGeoSearcher.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public SimpleGeoSearcher(String indexDir) throws IOException {
+    Path indexPath = Paths.get(indexDir);
+
+    if (!Files.exists(indexPath) || !Files.isDirectory(indexPath) || !Files.isReadable(indexPath)) {
+      throw new IllegalArgumentException(indexDir + " does not exist or is not a directory.");
+    }
+
+    reader = DirectoryReader.open(FSDirectory.open(indexPath));
+    searcher = new IndexSearcher(reader);
```
lintool, line 67: close everything for better hygiene?

lintool (commented):
So the bridge method you came up with is:

```
public Result[] searchGeo(int k, ShapeField.QueryRelation relation, LatLonGeometry... geometries)
```

From what I can tell?

What do you think of the alternative of just exposing `search` method that takes a Lucene query? And then we just wrap Java geo-based Lucene queries from the Python side?

lintool:
So to be more concrete, my suggestion is to expose a bunch of builder methods from the Python end, e.g., `createQueryTypeX`, `createQueryTypeY`, etc. These create the appropriate Lucene Query object, so then the bridge method just needs a simple `search` method.

Thoughts about this alternative design?

d1shs0ap:
> So to be more concrete, my suggestion is to expose a bunch of builder methods from the Python end, e.g., `createQueryTypeX`, `createQueryTypeY`, etc. These create the appropriate Lucene Query object, so then the bridge method just needs a simple `search` method.
> 
> Thoughts about this alternative design?

This alternative design sounds good to me. To confirm my understanding, we'd call something like `LatLonShape.newGeometryQuery()`(or any of the Query constructors [here](https://lucene.apache.org/core/8_11_0/core/org/apache/lucene/document/LatLonShape.html)) to construct the query on the Python end?

lintool:
Yup!

lintool (approved):

PR #1792, opened 2022-03-03
Hydrorivers sorted search

d1shs0ap:
Changes:
- `GeoGenerator.java`, `GeoGeneratorTest.java`: Added the "segment points" of lines/polygons during indexing so that we can query by distance later
- `GeoIndexerTestBase.java`, `line_sorted.wkt`: Add the later file as a document into the test index
- `SimpleGeoSearcher.java`: Added the ability to do a sorted search (sorted based on the distance to a given point)
- `SimpleGeoSearcherTest.java`: Demonstrates (and tests) sorted search function


src/main/java/io/anserini/index/generator/GeoGenerator.java
```diff
@@ -42,17 +42,39 @@ public Document createDocument(JsonCollection.Document geoDoc) {
 
           Field[] fields = new Field[0];
           if (shape instanceof Line) {
-            fields = LatLonShape.createIndexableFields("geometry", (Line) shape);
+            Line line = (Line) shape;
+            fields = LatLonShape.createIndexableFields("geometry", line);
+
+            for (int i = 0; i < line.numPoints(); ++i) {
```
lintool, line 48: Can you change to `i++`.
Also, consider if a functional style using `forEach()` will make the code look cleaner?

src/main/java/io/anserini/search/SimpleGeoSearcher.java
```diff
@@ -24,11 +25,20 @@
   private IndexSearcher searcher = null;
 
   public Result[] searchGeo(Query query, int k) throws IOException {
+    return searchGeo(query, k, null);
```
lintool, line 28: Don't recall from previous PR. Did we want just `search` here?

lintool (commented):
Don't feel too strongly... comments for you to think about.

src/main/java/io/anserini/search/SimpleGeoSearcher.java
d1shs0ap: @lintool Naming convention-wise, the previous PR also had the method named `searchGeo.` I chose to call it `searchGeo` instead of `search` to make it analogous to `searchTweets` in `SimpleTweetSearcher,` which also extends `SimpleSearcher.`

Functionally, the previous PR didn't have an overloaded method to support the sort argument and had all the logic in `searchGeo(Query query, int k).` In this PR, I refactored the logic into `searchGeo(Query query, int k, Sort sort)` so `searchGeo(Query query, int k)` becomes a wrapper function for when we don't care about the order of results.

Does this address what you had in mind?

src/main/java/io/anserini/index/generator/GeoGenerator.java
d1shs0ap: Changed to `i++`.

Perhaps `forEach()` would be hard to use in this case, since we are iterating over two arrays (`getLats()` and `getLons()`)?

src/main/java/io/anserini/search/SimpleGeoSearcher.java
lintool: Sounds good!

lintool (approved):

PR #1804, opened 2022-03-24
Add Rocchio reranking option

yuki617:
* Add script for rocchio reranking 
* Add rocchio option and update the regression table for disk45


src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -291,6 +291,41 @@
       usage = "RM3 parameter: turn off English term filter")
   public boolean rm3_noTermFilter = false;
 
+    // --------------------------
```
lintool, line 294: extra space here.

src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -291,6 +291,41 @@
...
+  // query expansion model: rocchio
+  // --------------------------
+
+  // Anserini uses the same default options as in Indri.
```
lintool, line 298: Change this to say that you got these default settings from vector-prf - give cite, etc.

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  @Override
+  public ScoredDocuments rerank(ScoredDocuments docs, RerankerContext context) {
+    assert(docs.documents.length == docs.scores.length);
+
+    IndexSearcher searcher = context.getIndexSearcher();
+    IndexReader reader = searcher.getIndexReader();
+
+    FeatureVector qfv = FeatureVector.fromTerms(AnalyzerUtils.analyze(analyzer, context.getQueryText())).scaleToUnitL1Norm();
```
lintool, line 77: Interesting. `scaleToUnitL1Norm()` isn't part of the original definition of Rocchio. I wonder what happens if you remove?

lintool (commented):
I'll take a more careful look again later, but just some initial comments.

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,211 @@
...
+
+    boolean useRf = (context.getSearchArgs().rf_qrels != null);
```
lintool, line 79: this is a no-op throughout this impl, so remove?

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,211 @@
...
+      }
+    }
+
+    f.pruneToSize(fbTerms);
+    f.scaleToUnitL1Norm();
+
+    return f;
+  }
+
+  private FeatureVector createdFeatureVector(Terms terms, IndexReader reader, boolean tweetsearch, int fbDocs) {
```
lintool, line 182: `tweetsearch` flag isn't used - removed from args list?

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,211 @@
...
+
+      BytesRef text;
+      while ((text = termsEnum.next()) != null) {
+        String term = text.utf8ToString();
+
+        if (term.length() < 2 || term.length() > 20) continue;
+        if (this.filterTerms && !term.matches("[a-z0-9]+")) continue;
+
+        int freq = (int) termsEnum.totalTermFreq();
+        f.addFeatureWeight(term, (float) freq / fbDocs);
```
lintool, line 196: not sure why we're doing this here?

lintool: Where are we actually getting the BM25 weights of the vector?

lintool (commented):
More comments...

src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -291,6 +291,41 @@
...
+
+  @Option(name = "-rocchio.fbTerms", handler = StringArrayOptionHandler.class,
+      usage = "Rocchio parameter: number of expansion terms")
+  public String[] rocchio_fbTerms = new String[]{"256"};
+
+  @Option(name = "-rocchio.fbDocs", handler = StringArrayOptionHandler.class,
+      usage = "Rocchio parameter: number of expansion documents")
+  public String[] rocchio_fbDocs = new String[]{"5"};
+
+  @Option(name = "-rocchio.originalQueryWeight", handler = StringArrayOptionHandler.class,
```
lintool, line 317: why don't we change to alpha, beta, gamma, as in the original definition?

yuki617: Changed into alpha, beta=1-alpha. In the future, if we add negative document, will name the weight to gamma

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
yuki617: Updated the freq into BM25 weight

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
yuki617: scaleToUnitL1Norm rises the score a bit, so I still kept it there.

src/main/java/io/anserini/search/SearchArgs.java
yuki617: Updated the comments. Btw, I found the score is usually better with shorter fbTerms, e.g 20,50. Should we still use the default setting(fbTerms=256) from vector-prf?

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,200 @@
...
+    IndexSearcher searcher = context.getIndexSearcher();
+    IndexReader reader = searcher.getIndexReader();
+
+    FeatureVector qfv = FeatureVector.fromTerms(AnalyzerUtils.analyze(analyzer, context.getQueryText())).scaleToUnitL1Norm();
+
+    FeatureVector rm = estimateRelevanceModel(docs, reader);
+
+    // Rocchio weight = alpha * original query + beta * mean(top n doc embedding)
+    // Here beta = 1-alpha
+    rm = FeatureVector.interpolate(qfv, rm, alpha);
```
lintool, line 86: Can we actually implement alpha and beta as per the actual definition?
Let's not use `interpolate` method in this case.

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
lintool: did you remove the `scaleToUnitL1Norm()` on the document vector also? You do so below....

lintool (commented):
More comments.

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
yuki617: No problem. Added the beta option.

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -606,7 +607,30 @@ public void close() throws IOException {
           }
         }
       }
-    } else {
+    } else if (args.rocchio) {
+      for (String fbTerms : args.rocchio_fbTerms) {
+        for (String fbDocs : args.rocchio_fbDocs) {
+          for (String alpha : args.rocchio_alpha) {
+            String tag;
+            float beta = 1 - Float.valueOf(alpha);
```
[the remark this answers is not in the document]
yuki617: Queensland paper still follow beta=1-alpha for all their experiments. But for consistence reason, let's still use both alpha and beta parameter.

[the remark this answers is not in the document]
lintool: agreed, so if we impl gamma, it'll make more sense.

src/main/java/io/anserini/search/SearchArgs.java
```diff
@@ -291,6 +291,41 @@
...
+  // --------------------------
+  // query expansion model: rocchio
+  // --------------------------
+
+  // Anserini uses the same default options as the Vector-clprf in the following work:
+  // Li, Hang, et al. "Pseudo relevance feedback with deep language models and dense retrievers: 
+  // Successes and pitfalls." arXiv preprint arXiv:2108.11044 (2021).
+  //
+  //   int fbDocs = _param.get( "fbDocs" , 5 );
+  //   int fbTerms = _param.get( "fbTerms" , 50 );
```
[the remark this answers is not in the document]
yuki617: Yes, this is from Queensland paper, but max-depth depends on the topic. The max-depth here is using the one from DL20-passage.

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,200 @@
...
+
+    if (this.outputQuery) {
+      LOG.info("QID: " + context.getQueryId());
+      LOG.info("Original Query: " + context.getQuery().toString(this.field));
+      LOG.info("Running new query: " + feedbackQuery.toString(this.field));
+    }
+
+    TopDocs rs;
+    try {
+      Query finalQuery = feedbackQuery;
```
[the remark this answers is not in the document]
yuki617: This is being used.

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
yuki617: No I didn't remove the both scaleToUnitL1Norm() for query and document vector

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,211 @@
...
+    FeatureVector f = new FeatureVector();
+
+    Set<String> vocab = new HashSet<>();
+    int numdocs;
+    numdocs = docs.documents.length < fbDocs ? docs.documents.length : fbDocs;
+
+    List<FeatureVector> docvectors = new ArrayList<>();
+    for (int i = 0; i < numdocs; i++) {
+      try {
+        FeatureVector docVector = createDocumentVector(
```
lintool, line 143: I'm fine with longish lines... code would look better on a single line.

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -606,7 +607,31 @@ public void close() throws IOException {
...
       }
-    } else {
+    } else if (args.rocchio) {
+      for (String fbTerms : args.rocchio_fbTerms) {
+        for (String fbDocs : args.rocchio_fbDocs) {
+          for (String alpha : args.rocchio_alpha) {
+            for (String beta : args.rocchio_beta) {
+              String tag;
+              if (this.args.rf_qrels != null){
+                tag = String.format("rocchioRf(fbTerms=%s,alpha=%s,beta=%s)",
```
lintool, line 617: I'm fine with longish lines... code would look better on a single line.

And below.

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
yuki617: Gotcha, fixed.

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  @Override
+  public ScoredDocuments rerank(ScoredDocuments docs, RerankerContext context) {
+    assert(docs.documents.length == docs.scores.length);
+
+    IndexSearcher searcher = context.getIndexSearcher();
+    IndexReader reader = searcher.getIndexReader();
+
+    FeatureVector qfv = FeatureVector.fromTerms(AnalyzerUtils.analyze(analyzer, context.getQueryText()));
```
lintool, line 80: `qfv` -> `queryVector`

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,212 @@
...
+
+    FeatureVector rm = computeMeanOfDocumentVectors(docs, reader);
```
lintool, line 82: `rm` -> `documentVector`

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,212 @@
...
+
+    // Rocchio weight = alpha * original query + beta * mean(top n doc embedding)
+    // Here beta = 1-alpha
```
lintool, line 85: "Here beta = 1-alpha" -> this comment is outdated, because we're reading beta from a param now.

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,212 @@
...
+    assert(docs.documents.length == docs.scores.length);
+
+    IndexSearcher searcher = context.getIndexSearcher();
+    IndexReader reader = searcher.getIndexReader();
+
+    FeatureVector qfv = FeatureVector.fromTerms(AnalyzerUtils.analyze(analyzer, context.getQueryText()));
+
+    FeatureVector rm = computeMeanOfDocumentVectors(docs, reader);
+
+    // Rocchio weight = alpha * original query + beta * mean(top n doc embedding)
```
lintool, line 84: "Rocchio weight" is inaccurate.
"Rocchio Algorithm"?

"top n doc embedding" -> "top n document vector"

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,212 @@
...
+    // Here beta = 1-alpha
+    rm = interpolateQueryDocVector(qfv, rm, alpha, beta);
```
lintool, line 86: I don't think the method name is accurate anymore? Maybe `computeWeightedVector`?

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -0,0 +1,212 @@
...
+
+    Set<String> vocab = new HashSet<>();
+    int numdocs;
+    numdocs = docs.documents.length < fbDocs ? docs.documents.length : fbDocs;
+
+    List<FeatureVector> docvectors = new ArrayList<>();
+    for (int i = 0; i < numdocs; i++) {
+      try {
+        FeatureVector docVector = createDocumentVector(reader.getTermVector(docs.ids[i], field), reader, docs.ids[i]);
+        docVector.pruneToSize(fbTerms);
```
lintool, line 144: I don't think we should prune each doc vector to size... since we already prune at the end.

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
lintool: Also, scale query vector also? `scaleToUnitL2Norm()`

lintool (changes requested):
More edits...

src/main/java/io/anserini/rerank/lib/RocchioReranker.java
```diff
@@ -102,7 +101,7 @@ public ScoredDocuments rerank(ScoredDocuments docs, RerankerContext context) {
       LOG.info("Running new query: " + feedbackQuery.toString(this.field));
     }
 
-    TopDocs rs;
+    TopDocs searchResult;
```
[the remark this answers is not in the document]
lintool: or just `results`?

lintool (commented):
Shouldn't the query weights be bm25 also?

lintool (approved):

lintool:
Since I worked with @yuki617 extensively on this, I wanted to sketch out for the record the process we went through, primarily to address the issue of "tuning on the test set". Were we doing it? To be honest, scientifically, kinda. But we did the best we could in terms of being "honest".

The fundamental issue is that we need to evaluate on _something_ to check the correctness of our implementation, and if we're changing the implementation based on the results, then _yes_, strictly speaking, we are tuning on the test set. This is less of a concern if we're fixing bugs, but other design choices... could go either way.

For example, it made sense to normalize the query and the mean of the doc vectors; because Rocchio would make little sense otherwise since the magnitudes of the vectors are so different (note that this isn't an issue with vector-prf with dense vectors). So how to normalize? I'd say L2 makes the most sense. But we tried both L1 and L2 and L2 worked better, so one might say this is a post-doc justification.

Note that in the process of implementation there was a bit of confusion on whether the query and mean of doc vectors we compute should be BM25 weighted. They should not - because we are using Lucene boosts for the 2nd round query, and the query itself handles the BM25 weighting. Whether or not implementation with Lucene boosts is "correct" - that's a separate matter... but it's how RM3 is implemented (using Lucene boosts), so this implementation won't be any more "wrong" than the RM3 implementation.

So, not BM25 weighted - that leaves two alternatives: TF weighted and boolean; this only makes a difference on the doc vectors end, since we're aggregating multiple doc vectors. We tried both, and ended up with the boolean implementation because it worked better. I would say that boolean makes more sense... but to be honest, had TF come out on top, I would have probably found some explanation as to why it made more sense.

Nevertheless, during the course of the implementation, we _did not_ examine effectiveness on any collection other than Disk 4/5.

Finally, there is the issue of Rocchio parameters. We tried three sets:

+ The set from Vector PRF (parameters from Queensland work on dense vectors)
+ The parameters from RM3.
+ The alpha and beta settings from the [Manning et al. textbook](https://nlp.stanford.edu/IR-book/html/htmledition/the-rocchio71-algorithm-1.html) (alpha = 1, beta = 0.75)

In the end, we used RM3 parameters for fbTerms and fbDocs, and alpha/beta weights from Manning et al. I think this can be characterized as zero-shot parameter transfer. We did not do any further tuning.

PR #1819, opened 2022-03-31
Refactor Rocchio implementation

lintool:
+ Simplify creation of original query vector
+ Add more detailed comments
+ Add heuristics for terms to include in doc vector, taken directly from RM3


yuki617 (approved):
Looks good to me! And I can confirm all rocchio scores can be reproduced.

PR #1847, opened 2022-04-19
Query Encoding Using Unicoil Onnx Model Test

jh8liang:


jh8liang:
@lintool 

This code review covers the test cases of bringing both the bert tokenizer and also the unicoil query encoder model on huggingface (castorini/unicoil-msmarco-passage model) to java.

The examples used in the two test classes are the same set of queries chosen from msmarco-v2-passage.dev and topics.dl20. I randomly picked a few queries with different lengths to be used for testing. The examples array in query tokenization test class includes the query string itself and the token ids of the query string generated by huggingface's bert tokenizer in the same way in pyserini. The examples array in query inference test includes the same set of query strings but in token ids format, and the unicoil model's output weights for each token from pyserini.

I am having some trouble including the UnicoilEncoder.onnx model in this code review because the file size is >100mb and github public fork does not support LFS. I want to ask you for you opinion/help on how I should include the onnx model file into anserini. I ran the tests locally and all test cases should work and I just need to find a way to include the onnx model file. I will send you a separate email with the onnx model for you to take a look at as well. Please let me know. Thanks.

lintool:
hi @jh8liang - thanks for looking into this!

For both the model itself and the vocab, can we dynamically download from a known location (e.g., Dropbox) into `~/.cache/anserini`? This is what Pyserini does. Shouldn't be too hard to write the download jig?

jh8liang:
I just updated the PR to pull directly from my dropbox for the vocab file and onnx model. Let me know if you think these tests are alright and I will update this PR according to your comments

jh8liang:
Hi @lintool ,
Here is my brief update for the PR:

- Query encoding (unicoil msmarco passage model) can now be done on the fly for sparse retrieval tasks by specifying the following SearchArgs:
   - -encoder flag specifies name of the encoder model e.g) UnicoilMsmarcoPassage (default is null)
   - -weightRange specifies range of weights (default to 5)
   - -quantRange specifies range of quantization (default to 256)
- Onnx encoder integration with Anserini is done in "SearchCollections" class where each of the query string read by TopicReader is directly passed into the specified encoder model if specified to get the encoded query string to be used in downstream search tasks

For testing:
- I am able to generate the exact same encoded query file for topics-msmarco-passage-dev-subset as the pyserini.encode.query module using the onnx unicoil model in java (with parameters weightRange=5, quantRange=256)
- But the encoded query generated from pyserini or onnx unicoil model seems to be different from the tsv file in anserini: "src/main/resources/topics-and-qrels/topics.msmarco-passage.dev-subset.unicoil.tsv.gz" which I find to be kinda strange. The "topics.msmarco-passage.dev-subset.unicoil.tsv.gz" seems to mismatch from encoding generated by pyserini using unicoil-msmarco-passage model with parameters weightRange=5 and quantRange=256?
- As a result the [unicoil regression](https://github.com/castorini/anserini/blob/master/docs/regressions-msmarco-passage-unicoil.md#anserini-regressions-ms-marco-passage-ranking) results is slightly different than expected:
  - AP@1000: 0.3568
  - RR@10: 0.3509
  - R@100: 0.8619
  - R@1000: 0.9583
  - These are the results I got from feeding anserini either the encoded queries generated from pyserini myself, or performing query encoding on the fly with onnx model directly in anserini
- I am able to reproduce the original regression values however using the pre-encoded queries under anserini resources. I also appended myself under the reproduction log

Follow ups/Next steps:
- Want to ask for your opinion on the PR to see if you think there are updates that I should make to it
-  Want to ask if you know how the topics file in anserini: "src/main/resources/topics-and-qrels/topics.msmarco-passage.dev-subset.unicoil.tsv.gz" is generated? It doesnt seem to match the results generated by pyserini's encode.query module

lintool:
Reference regressions:
+ anserini pre-encoded queries: https://github.com/castorini/anserini/blob/master/docs/regressions-msmarco-passage-unicoil.md
+ pyserini pre-encoded queries: https://github.com/castorini/pyserini/blob/master/docs/experiments-unicoil.md

jh8liang:
Hi @lintool , yes I am still interested and have merged in the latest trunk

lintool:
hi @jh8liang I pulled your branch and built - your fatjar comes out to be quite large:

```
% ls -lh target/anserini-0.15.1-SNAPSHOT-fatjar.jar
-rw-r--r--  1 jimmylin  staff   176M Sep 25 07:46 target/anserini-0.15.1-SNAPSHOT-fatjar.jar
```

It's over our 150 MB PyPI limit. Can you try to slim down?

Use `mvn dependency:tree` as a start?

lintool:
Superseded by #2077

PR #1856, opened 2022-04-25
Beir splade patch

justram:
- [x] Add topics, regressions, and tests for splade distill cocodenser medium; collections: cqadupstack subsets, bioasq, trec-news, signal-1m, and robust04
- [x] Revise collection namings for splade distill cocodenser medium


src/main/resources/regression/beir-v1.0.0-robust04-flat.yaml
```diff
@@ -8,9 +8,9 @@ generator_class: DefaultLuceneDocumentGenerator
 index_threads: 1
 index_options: -storePositions -storeDocvectors -storeRaw
 index_stats:
-  documents: 528036
```
lintool, line 11: I think there's a bug here?

lintool (approved):
lg, but double check [beir-v1.0.0-robust04-flat.yaml](https://github.com/castorini/anserini/pull/1856/files#diff-363732a25844f016e11b40d1d53b6b7ffd98c93235cdf9343ac7f8d87a4a85e3)... looks wrong - should have 528k docs...

src/main/resources/regression/beir-v1.0.0-robust04-flat.yaml
justram: Right, accidentally changed the old yaml when creating the new ones. I am fixing it.

lintool (approved):

Issue #1778, opened 2022-02-24
Multithread method to get document from index


lintool:
Currently, `SimpleSearcher` has `document` method to fetch document. This is on a single thread. We should make it mulit-threaded for a "bulk get". Needed for this: https://github.com/castorini/pyserini/issues/370

HAKSOAT:
Hi @lintool 

Just putting it here that I'm on this task.

PR #1857, opened 2022-04-26
Add Multi-threaded Batch Document

HAKSOAT:
This addresses #1778

### DIFF
--- a/src/main/java/io/anserini/search/SimpleSearcher.java
+++ b/src/main/java/io/anserini/search/SimpleSearcher.java
@@ -682,6 +682,45 @@ public class SimpleSearcher implements Closeable {
   }
 
   /**
+   * Returns a map of internal Lucene docid to Lucene {@link Document}.
+   * Batch version of {@link #document(int)}.
+   *
+   * @param docids list of docids
+   * @return a map of docid to corresponding Lucene {@link Document}
+   */
+  public Map<Integer, Document> batchDocument(List<Integer> docids, int threads){
+    ThreadPoolExecutor executor = (ThreadPoolExecutor) Executors.newFixedThreadPool(threads);
+    ConcurrentHashMap<Integer, Document> results = new ConcurrentHashMap<>();
+
+    for (Integer docid: docids) {
+      executor.execute(() -> {
+        Document result = null;
+        try {
+          result = reader.document(docid);
+          results.put(docid, result);
+        } catch (Exception e){}
+      });
+    }
+
+    executor.shutdown();
+
+    try {
+      // Wait for existing tasks to terminate
+      while (!executor.awaitTermination(1, TimeUnit.MINUTES)) {
+        LOG.info(String.format("%.2f percent completed",
+                (double) executor.getCompletedTaskCount() / docids.size() * 100.0d));
+      }
+    } catch (InterruptedException ie) {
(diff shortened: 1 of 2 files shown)


HAKSOAT:
Hi, @lintool I have a PR in progress. I need your thoughts on some things:

1. The original `document` methods [L663](https://github.com/castorini/anserini/blob/b429218e52a385eabf3fd81979e221111fbc4a19/src/main/java/io/anserini/search/SimpleSearcher.java#L663) and [L680](https://github.com/castorini/anserini/blob/b429218e52a385eabf3fd81979e221111fbc4a19/src/main/java/io/anserini/search/SimpleSearcher.java#L680) take in an integer and a string respectively. However, I'm not very sure how I can go about doing this without creating two `batchDocument` methods with different names. Hence, I'd like to know what the best practice is for Anserini. The PR only currently tackles batch for internal lucene docs and I'd like to add something for collection docid too.

2. After compiling I saw a bunch of modified markdown files. Will I push all of them as well? I didn't push because I didn't want them to mess up the PR review.

lintool:
hi @HAKSOAT - thanks for this PR.

Let's go with a single method for now, called `batchGetDocument`. It should take a list of (external) collection ids, which is the more common case.

If you update and merge in current master, it should get rid of the new markdown edits?

Thanks!

src/test/java/io/anserini/search/SimpleSearcherTest.java
```diff
@@ -54,6 +55,29 @@ public void testGetDoc() throws Exception {
...
+
+    ArrayList<Integer> ldocIds = new ArrayList<Integer>();
+    ldocIds.add(0);
+    ldocIds.add(1);
+    ldocIds.add(2);
+    ldocIds.add(3);
+
+    Map<Integer, Document> results = searcher.batchDocument(ldocIds, 2);
+
+    assertEquals("here is some text here is some more text. city.",
```
lintool, line 70: I prefer _not_ to break these lines...

src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -681,6 +681,45 @@ public Document document(String docid) {
...
   }
 
+  /**
+   * Returns a map of collection docid to Lucene {@link Document}.
+   * Batch version of {@link #document(String)}.
+   *
+   * @param docids list of docids
+   * @return a map of docid to corresponding Lucene {@link Document}
+   */
+  public Map<String, Document> batchGetDocument(List<String> docids, int threads){
```
lintool, line 691: space before { here.

lintool:
lg, but can you please build a fatjar, copy over to pyserini, and make sure it connects up on the pyserini end? i.e., let's build the bindings on the pyserini end just to confirm everything works e2e.

HAKSOAT:
> lg, but can you please build a fatjar, copy over to pyserini, and make sure it connects up on the pyserini end? i.e., let's build the bindings on the pyserini end just to confirm everything works e2e.

Noted.

HAKSOAT:
> lg, but can you please build a fatjar, copy over to pyserini, and make sure it connects up on the pyserini end? i.e., let's build the bindings on the pyserini end just to confirm everything works e2e.

It works. I'll put a PR through for this on Pyserini.

lintool (approved):

PR #1858, opened 2022-04-26
Rename msmarco-passage -> msmarco-passage-wp in YAML for wp-based regressions

lintool:
Otherwise, if we run the "base" and "wp" versions concurrently, they clash with each other.


yuki617 (approved):
lgtm!

PR #1873, opened 2022-05-13
Indexing CC-100 Corpus (https://data.statmt.org/cc-100/)

ToluClassics:


src/main/java/io/anserini/collection/CommonCrawl100Collection.java
```diff
@@ -0,0 +1,169 @@
...
+    private List<JsonNode> jsonNodeArray = null;
+    private Iterator<JsonNode> iterator; // iterator for JSON line objects
+  
+    public Segment(Path path) throws IOException {
+      super(path);
+  
+      if (path.toString().endsWith(".xz")) {
+        bufferedReader = new BufferedReader(new InputStreamReader(
+            new XZInputStream(new FileInputStream(path.toString()))));
+      }else{
```
lintool, line 66:
```suggestion
      } else {
```

src/main/java/io/anserini/collection/CommonCrawl100Collection.java
```diff
@@ -0,0 +1,169 @@
...
+  
+    private List<JsonNode> getTxtFiles(BufferedReader reader) throws IOException {
+      List<JsonNode> jsonNodeArray = new ArrayList<>();
+      ObjectMapper objectMapper = new ObjectMapper();
+    
+      String line;
+      int i=0;
+      while ((line = reader.readLine()) != null) {
+        try{
+          String json = "{ \"id\" : \"doc_"+i+"\", \"contents\" : \""+line.replaceAll("[-+\"\'^[\\\\p{C}]\\\\]*","").strip()+"\" }";
```
lintool, line 88: will this generate unique docids across a large corpus?

lintool (commented):
hi @ToluClassics can you explain the rationale for generating a "fake" JSON document internally? Seems a bit janky... we should document the reason why this approach is justified?

src/main/java/io/anserini/collection/CommonCrawl100Collection.java
```diff
@@ -0,0 +1,169 @@
...
+          JsonNode jsonNode = objectMapper.readTree(json);
+          jsonNodeArray.add(jsonNode);
+          i++;
+        }catch (IOException e) {
```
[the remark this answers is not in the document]
ToluClassics: > hi @ToluClassics can you explain the rationale for generating a "fake" JSON document internally? Seems a bit janky... we should document the reason why this approach is justified?

Since it's a monolingual corpus; i'm just using the line ids as the document id;
doing same for the Afriberta corpus collection

src/main/java/io/anserini/collection/CommonCrawl100Collection.java
ToluClassics: Since it's a monolingual corpus; i'm just using the line ids as the document id;
doing same for the Afriberta corpus collection

ToluClassics: @lintool I have taken out the fake JSON document

src/main/java/io/anserini/collection/CommonCrawl100Collection.java
```diff
@@ -0,0 +1,125 @@
...
+    @Override
+    public void readNext() throws NoSuchElementException, IOException {
+      String line;
+      String id = "";
+      String contents = "";
+
+      while ((line = bufferedReader.readLine())!=null){
+        id = "doc_"+i;
+        contents = line;
+        LOG.error(id+"\n"+contents);
```
lintool, line 78: I don't think you want to do this? esp for large collections?

src/main/java/io/anserini/collection/CommonCrawl100Collection.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public void readNext() throws NoSuchElementException, IOException {
+      String line;
+      String id = "";
+      String contents = "";
+
+      while ((line = bufferedReader.readLine())!=null){
+        id = "doc_"+i;
```
lintool, line 76: from what I can tell, you're not incorporating the name of the file as part of the docid?

which means that if you collection has multiple files, you're going to have collisions?

PR #1881, opened 2022-05-20
Add regression for MS MARCO passage ranking BM25-Quantized

manveertamber:


src/main/resources/docgen/templates/msmarco-passage-bm25-b8.template
```diff
@@ -0,0 +1,52 @@
+# Anserini Regressions: MS MARCO Passage Ranking With Quantized BM25 Weights
+
```
lintool, line 2: Add something like "Models: BM25 with quantized weights (8 bits)" to be consistent with:
https://github.com/castorini/anserini/blob/master/docs/regressions-msmarco-passage.md

lintool (approved):
Also add link to `.md` in main README:
https://github.com/castorini/anserini#ms-marco-v1-passage-corpus

Something like "quantized BM25".

Go ahead and ship after that!

Issue #1484, opened 2021-02-28
Batching for IndexReader's methods


alexlimh:
Does anyone know if there's a batching method for IndexReader.compute_query_document_score in pyserini? The original method seems to run very slow, roughly 6 hrs for ~500k query-doc pairs on an Intel(R) Xeon(R) CPU E5-2699 v4 @ 2.20GHz.

PR #1882, opened 2022-05-21
Added batch equivalent of computeQueryDocumentScore

HAKSOAT:
Addresses #1484

### DIFF
--- a/src/main/java/io/anserini/index/IndexReaderUtils.java
+++ b/src/main/java/io/anserini/index/IndexReaderUtils.java
@@ -735,18 +735,17 @@ public class IndexReaderUtils {
    * @param reader index reader
    * @param docids A list of docids of the documents to score
    * @param q query
-   * @param threads number of threads
    * @return a map of document ids to their scores with respect to the query
    * @throws IOException if error encountered during query
    */
   public static Map<String, Float> batchComputeQueryDocumentScore(
-          IndexReader reader, List<String> docids, String q, int threads)
+          IndexReader reader, List<String> docids, String q)
           throws IOException {
 
     SearchArgs args = new SearchArgs();
     return batchComputeQueryDocumentScoreWithSimilarityAndAnalyzer(reader, docids, q,
             new BM25Similarity(Float.parseFloat(args.bm25_k1[0]), Float.parseFloat(args.bm25_b[0])),
-            IndexCollection.DEFAULT_ANALYZER, threads);
+            IndexCollection.DEFAULT_ANALYZER);
   }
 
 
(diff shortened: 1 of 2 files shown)


HAKSOAT:
Hi @lintool I'd appreciate a review here when chanced.

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -726,7 +733,153 @@ public static float computeQueryDocumentScoreWithSimilarityAndAnalyzer(
...
+    ConcurrentHashMap<String, Float> results = new ConcurrentHashMap<>();
+
+    for (String docid: docids) {
+      executor.execute(() -> {
+        try {
+          Query query = new BagOfWordsQueryGenerator().buildQuery(IndexArgs.CONTENTS, analyzer, q);
+
+          Query filterQuery = new ConstantScoreQuery(new TermQuery(new Term(IndexArgs.ID, docid)));
+          BooleanQuery.Builder builder = new BooleanQuery.Builder();
+          builder.add(filterQuery, BooleanClause.Occur.MUST);
```
lintool, line 852: What you want to do is to move the docids here: In the non-batch impl, the filter clause restricts to a single docid. Here, in the batch impl, you want to restrict to a _set_ of docids - i.e., add multiple sub-clauses in the filter query.

lintool (changes requested):
Suggesting a better impl.

src/main/java/io/anserini/index/IndexReaderUtils.java
HAKSOAT: Hi @lintool I took a look at this and tried testing with a set of documents from `robust04`. However, I came across the error: `org.apache.lucene.search.BooleanQuery$TooManyClauses: maxClauseCount is set to 1024`.

This resulted from doing:

```
    for (String docid: docids){
      // Setting default result value for all docids.
      results.put(docid, 0.0f);
      Query filterQuery = new ConstantScoreQuery(new TermQuery(new Term(IndexArgs.ID, docid)));
      builder.add(filterQuery, BooleanClause.Occur.SHOULD);
    }
```

What are your thoughts on this? 

Am I doing the right thing? I tried this with tests and it works when the clause count is less than 1024.

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -20,6 +20,8 @@
 import io.anserini.search.SearchArgs;
 import io.anserini.search.query.BagOfWordsQueryGenerator;
 import io.anserini.search.query.PhraseQueryGenerator;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
```
HAKSOAT, line 24: Note to @HAKSOAT: Remove these logger imports if not needed in the final implementation.

PR #1900, opened 2022-06-09
Add Rocchio to SimpleSearcher

yuki617:


src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -841,7 +911,18 @@ public static void main(String[] args) throws Exception {
         searcher.setRM3(searchArgs.rm3_fbTerms, searchArgs.rm3_fbDocs, searchArgs.rm3_originalQueryWeight);
       } else {
         LOG.info("Testing code path of default RM3 parameters.");
-        searcher.setRM3();
+         searcher.setRM3();
```
lintool, line 914: I think there's an extra space here?

lintool (commented):
Can we add a few test cases? E.g., here - https://github.com/castorini/anserini/blob/master/src/test/java/io/anserini/search/SimpleSearcherTest.java

lintool (approved):

PR #1907, opened 2022-06-15
Add regression for HC4 monolingual retrieval BM25

ToluClassics:
Add regression for HC4 monolingual retrieval with human translated queries BM25


lintool (approved):

lintool:
@ToluClassics Before you merge, add a link from the main README?

PR #1920, opened 2022-06-26
Improve alignment in effectiveness tables in regression documentation

lintool:
As we add more conditions in a regression test, the effectiveness tables increase in width - and at some point, they require horizontal scrolling. Combining separate tables (for each metric) into a single table improves alignment in this case and greatly increases readability.


yuki617 (approved):
lgtm.

PR #1924, opened 2022-07-11
Update HC4 Regressions with Test Topics

ToluClassics:


lintool:
Discussed over slack, but adding here for reference:

https://arxiv.org/pdf/2201.09992.pdf

table 1 calls the splits train/eval - should we rename the same way?

docs/regressions-hc4-v1.0-fa.md
```diff
@@ -1,6 +1,6 @@
 # Anserini Regressions: HC4 (v1.0) &mdash; Persian
 
-This page documents BM25 regression experiments for [HC4 (v1.0) &mdash; Persian](https://arxiv.org/pdf/2201.09992.pdf).
+This page documents BM25 regression experiments for [HC4 (v1.0) &mdash; Persian](https://github.com/hltcoe/HC4), [Paper](https://arxiv.org/pdf/2201.09992.pdf).
```
lintool, line 3:
```suggestion
This page documents BM25 regression experiments for [HC4 (v1.0) &mdash; Persian](https://github.com/hltcoe/HC4), ([paper](https://arxiv.org/pdf/2201.09992.pdf)).
```

I think this will look better?
If so, do the same for each? (but in the template, I guess...)

PR #1961, opened 2022-08-16
Update Pyserini bindings to Lucene 9

lintool:
Exposes Lucene 8 backwards compatibility bindings in `SimpleSearcher` and `SimpleImpactSearcher`.

Basically, if we detect Lucene 8 indexes, we disable consistent tie-breaking, which depends on docvalues; see #1952

Additional changes:
+ general cleanup (fixed code formatting in SimpleImpactSearcher)
+ killed main in SimpleSearcher
+ Changed to Python method names (snake_case)


MXueguang:
we change the name style to python style in SimpleSearchers?

lintool:
> we change the name style to python style in SimpleSearchers?

Yes... since this is a major breaking upgrade, I thought I'd do some cleanup while I was at it. This would bring closer alignment to the pyserini end.

Thoughts? Let me know if you _don't_ think this is a good idea?

MXueguang:
I think its good, simple(impact)searchers are for python anyway

MXueguang (approved):

PR #1974, opened 2022-09-16
MrTydi Updated regressions (MIRACL)

ToluClassics:


lintool:
I don't think name is accurate: `mrtydi-v1.1-miracl-ar`; should probably be something like `miracl-v1.0-ar`?

PR #1992, opened 2022-10-14
Fix batch_search for RM3, made BM25prf thread-safe

lintool:
batch_search for RM3 doesn't appear to work, as reported in https://github.com/castorini/pyserini/issues/831

I tracked down the issue to the fact that analyzers have state, and so concurrent calls to analyze the query causes issues.
Solution is to analyze the query just once.

I also fixed the thread safety issue with BM25prf. The issue is that we swap in a different similarity in the searcher,
so interleaved execution confuses the reranker wrt what similarity we're actually using.


lintool:
This is ready for review.

Interestingly, this PR also fixes a bug in bglinking regressions; i.e., the implementation before was actually not correct.

PR #1999, opened 2022-10-23
MIRACL Dev Set Regressions and Test Cases

ToluClassics:


src/main/resources/regression/miracl-v1.0-ar.yaml
```diff
@@ -0,0 +1,48 @@
+---
+corpus: miracl-v1.0-ar
+corpus_path: collections/miracl-corpus/miracl-corpus-v1.0-ar
+
+index_path: indexes/lucene-index.miracl-v1.0-arabic/
```
lintool, line 5: Change to `lucene-index.miracl-v1.0-ar` to be consistent with corpus? i.e. two-letter ISO.

lintool (changes requested):
We need an update to the README, and also https://github.com/castorini/anserini/blob/master/docs/regressions.md#invocations

lintool:
For the edits to README, let's create an entirely new section, like this https://github.com/castorini/anserini#regressions-for-beir-v100

Put each lang in rows. Right now we only have BM25 (col) - but I'm sure we'll have more conditions as we move forward.

lintool (approved):

PR #2004, opened 2022-10-26
Fix bug in HC4 doc translation regressions

lintool:
These are document translation conditions - i.e., the documents are in English. 
So, we shouldn't be using a non-English tokenizer.


ToluClassics:
oh wow

ToluClassics (approved):

Issue #2018, opened 2022-11-13
Refactoring HNSW Lucene classes


lintool:
Hi @MXueguang - I propose the following renaming:

+ `IndexVectorCollection` -> `IndexDenseVectors`
+ `VectorCollection` -> `JsonDenseVectorCollection` (later, we can have `BinaryDenseVectorCollection`, etc.)
+ `LuceneVectorDocumentGenerator` -> `LuceneDenseVectorDocumentGenerator`
+ `IndexVectorArgs` -> make it an inner class `IndexDenseVectors.Args`
+ `SearchVectorCollection` -> `SearchDenseVectors`

The key distinction here is between "dense" and "sparse"... because BM25 is a vector also...

So we can have in parallel (perhaps later):

+ `IndexCollection` -> `IndexSparseCollection`
+ `IndexArgs` -> `IndexSparseCollection.Args`

PR #2020, opened 2022-11-14
Change name hnsw and add HNSW indexing config

MXueguang:
now we can config hnsw parameters
```
./target/appassembler/bin/IndexDenseVectors -collection JsonDenseVectorCollection \
 -generator LuceneDenseVectorDocumentGenerator -threads 1 -input <> \
 -index <> -M 32 -efC 100
```


MXueguang:
also solve part of https://github.com/castorini/anserini/issues/2018

src/main/java/io/anserini/index/IndexDenseVectorsArgs.java
```diff
@@ -20,7 +20,7 @@
 import org.kohsuke.args4j.spi.StringArrayOptionHandler;
 
 
-public class IndexVectorArgs {
+public class IndexDenseVectorsArgs {
```
lintool, line 23: I thought we wanted to make this an inner class?

src/main/java/io/anserini/search/SearchDenseVectors.java
```diff
@@ -316,12 +316,12 @@ public static void main(String[] args) throws Exception {
...
 
     final long start = System.nanoTime();
-    SearchVectorCollection searcher;
+    SearchDenseVectors searcher;
 
     // We're at top-level already inside a main; makes no sense to propagate exceptions further, so reformat the
     // exception messages and display on console.
     try {
-      searcher = new SearchVectorCollection(searchArgs);
+      searcher = new SearchDenseVectors(searchArgs);
```
lintool, line 324: Same of the args here... did we want to make them inner classes?

src/main/java/io/anserini/index/IndexDenseVectors.java
```diff
@@ -364,19 +372,19 @@ public Counters run() throws IOException {
   }
 
   public static void main(String[] args) throws Exception {
-    IndexVectorArgs indexCollectionArgs = new IndexVectorArgs();
+    IndexDenseVectorsArgs indexCollectionArgs = new IndexDenseVectorsArgs();
```
MXueguang, line 375: a java question

I'm a bit unsure how to properly initialize the Args here if we make it subclass
@lintool 
I don't think I can do
```
IndexDenseVectors.Args args = new IndexDenseVectors.Args()
```

MXueguang: we need to parse the args first and then initialize the index class as IndexDenseVectors(args)

lintool (approved):

PR #2032, opened 2022-12-07
Add JsonStringTopicReader

justram:
- Add support for JSON Lines topics
- Sample format: `{"id": "topic1", "contents": "this is the contents 1.", "field1": "topic1 field1 content", "field2": "topic1 field2 content"}`
- Usage: `${ANSERINI_PATH}/target/appassembler/bin/SearchCollection` `-topicfield <field>`
- Tested with transformed [topics.dl19-passage.jsonl](https://github.com/justram/anserini/blob/json_string-id_topicreader/src/test/resources/sample_topics/topics.dl19-passage.jsonl) and [stringID_topics.jsonl](https://github.com/justram/anserini/blob/json_string-id_topicreader/src/test/resources/sample_topics/stringID_topics.jsonl)


src/main/resources/regression/dl19-passage-jsonl.yaml
```diff
@@ -0,0 +1,91 @@
+---
```
lintool, line 1: I don't think we need this since we have until tests?

Adding a regression means that I have to test it forever...

lintool (approved):

src/main/resources/regression/dl19-passage-jsonl.yaml
justram: Yes, it's for my sanity check. I can remove this file.

lintool: yes please

PR #2037, opened 2022-12-14
SearchDenseVectors pass the ThreadPoolExecutor to IndexSearcher for parallelizing searches

tteofili:
Make it possible to leverage parallel index search in `SearchDenseVectors` by passing the `Executor` to `IndexSearcher` via `-parallel` param.


lintool:
hey @tteofili let's completely redo the parallelism options into something that actually makes sense? the `-threads` and `-parallelism' options is technical debt and confusing, copied over from `SearchCollection`. Since we're starting with a clean slate, we should just implement it "right".

lintool:
So _Anserini_ would be able to run _n_ queries in parallel, and for each query you'd have a threadpool of _m_ threads? So that's a total of n * m threads?

For modern hardware, with so many cores, that makes sense.

tteofili:
yes that is the idea. That way we can more easily evaluate the performance speedup provided by parallelizing each part.

tteofili:
moved the `-parallel` option to `-threadsPerSearch`.

lintool:
> moved the `-parallel` option to `-threadsPerSearch`.

How about `-threadsPerQuery`?

tteofili:
> How about `-threadsPerQuery`?

👍

src/main/java/io/anserini/search/SearchDenseVectors.java
```diff
@@ -341,22 +344,43 @@ public <K> void runTopics() throws IOException {
 
 
     LOG.info("============ Launching Search Threads ============");
-
+    ThreadPoolExecutor searcherExecutor;
+    if (args.threadsPerQuery > 1) {
+      searcherExecutor = (ThreadPoolExecutor) Executors.newFixedThreadPool(args.threadsPerQuery);
+    } else {
+      searcherExecutor = null;
```
lintool, line 351: if you set `ThreadPoolExecutor searcherExecutor = null` above, you won't need this branch?

lintool (approved):
Minor suggestions...

PR #2047, opened 2023-01-14
Update NeuCLIR 2022 regressions to new qrels

lintool:


ToluClassics (approved):

ToluClassics:
The scores are much lower than before with the new qrels

PR #2048, opened 2023-01-15
Update reproduction log

Cathrineee:
Computer: Apple MacBook Pro (with M1Pro)

Maven version: Apache Maven 3.8.7
Java version: 11.0.16.1
OS name: "mac os x", version: "13.0.1", arch: "aarch64"


lintool (commented):
Something's off... your entry says you reproduced 2023-01-14, but the commit id is from after that date...? Maybe you updated the repo in the meantime?

Cathrineee:
> Something's off... your entry says you reproduced 2023-01-14, but the commit id is from after that date...? Maybe you updated the repo in the meantime?

Updated

lintool (approved):

PR #2054, opened 2023-01-24
Fix neuclir queries

ToluClassics:
Remove unecessary quotes from neuclir queries


lintool:
I've added separately to my own PR.

Issue #2060, opened 2023-02-04
Update SimpleIndexer Args 


theyorubayesian:
SimpleIndexer class needs to be updated to use args such as analyzeWithHuggingFaceTokenizer, storePositions, storeDocVectors etc.

PR #2061, opened 2023-02-04
update SimpleIndexer to use IndexCollection.Args

theyorubayesian:
A couple of notes:
- AnalyzerMap.java helps avoid a lengthy if/else in SimpleIndexer as we have currently in IndexCollection when selecting language-specific analyzers
- `storeDocvectors`, `storePositions` etc are usually passed through IndexCollection.Args to the generator. Rather than update all generators with new constructors able to use SimpleIndexer.Args to collect this argument, I currently allow SimpleIndexer to reuse `IndexCollection.Args`. 
- A downside of this decision is that when SimpleIndexer is initialized from Args, `-collection` and `-threads` must be passed (because they are marked as required in IndexCollection.Args). However, they are not used by Simplelndexer.

See #2060

### DIFF
--- a/src/main/java/io/anserini/analysis/AnalyzerMap.java
+++ b/src/main/java/io/anserini/analysis/AnalyzerMap.java
@@ -1,3 +1,19 @@
+/*
+ * Anserini: A Lucene toolkit for reproducible information retrieval research
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.anserini.analysis;
 
 import org.apache.lucene.analysis.Analyzer;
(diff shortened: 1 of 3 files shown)


src/main/java/io/anserini/analysis/AnalyzerMap.java
```diff
@@ -0,0 +1,42 @@
+package io.anserini.analysis;
```
lintool, line 1: Add package header boilerplate please.

src/main/java/io/anserini/analysis/AnalyzerMap.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.apache.lucene.analysis.Analyzer;
+import java.util.HashMap;
+import java.util.Map;
+
+public class AnalyzerMap {
+    public static final Map<String, String> analyzerMap = new HashMap<String, String>() {
```
lintool, line 8: two space indent

lintool (changes requested):
Can you please add test cases to https://github.com/castorini/anserini/blob/master/src/test/java/io/anserini/index/SimpleIndexerTest.java ?

src/main/java/io/anserini/analysis/AnalyzerMap.java
```diff
@@ -0,0 +1,58 @@
...
+      put("sv", "org.apache.lucene.analysis.sv.SwedishAnalyzer");
+      put("te", "org.apache.lucene.analysis.te.TeluguAnalyzer");
+      put("th", "org.apache.lucene.analysis.th.ThaiAnalyzer");
+      put("tr", "org.apache.lucene.analysis.tr.TurkishAnalyzer");
+      put("uk", "org.apache.lucene.analysis.uk.UkrainianMorfologikAnalyzer");
+      put("zh", "org.apache.lucene.analysis.cjk.CJKAnalyzer");
+    }
+  };
+
+  public static Analyzer getLangSpecificAnalyzer(String language) throws Exception {
```
lintool, line 54: `getLangSpecificAnalyzer` -> `getLanguageSpecificAnalyzer`?

Personally, I like things fully written out...

lintool (approved):

PR #2062, opened 2023-02-23
Add ability to append to existing index

lintool:


src/main/java/io/anserini/analysis/AnalyzerMap.java
```diff
@@ -51,8 +53,24 @@ public class AnalyzerMap {
...
+    } catch (InstantiationException e) {
+      e.printStackTrace();
+    } catch (IllegalAccessException e) {
+      e.printStackTrace();
+    } catch (InvocationTargetException e) {
+      e.printStackTrace();
+    } catch (NoSuchMethodException e) {
+      e.printStackTrace();
+    } catch (ClassNotFoundException e) {
+      e.printStackTrace();
```
tteofili, line 70: what about adding an _ERROR_ log rather than printing the stacktrace?

theyorubayesian (approved):
LGTM

PR #2068, opened 2023-02-27
Add all current analyzers to DumpAnalyzedQueries

JMMackenzie:
Allows the query dump tool to use analyzers beyond the simple English/Porter combination.

### DIFF
--- a/src/main/java/io/anserini/util/DumpAnalyzedQueries.java
+++ b/src/main/java/io/anserini/util/DumpAnalyzedQueries.java
@@ -29,6 +29,9 @@ import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
 import org.kohsuke.args4j.ParserProperties;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.file.Path;
@@ -42,6 +45,8 @@ import java.util.SortedMap;
  * analyzed query) tuples; the analyzed query comprises space-delimited tokens.
  */
 public class DumpAnalyzedQueries {
+
+  private static final Logger LOG = LogManager.getLogger(DumpAnalyzedQueries.class);
 
   public static class Args {
     @Option(name = "-topicreader", metaVar = "[class]", usage = "topic reader")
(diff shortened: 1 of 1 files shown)


src/main/java/io/anserini/util/DumpAnalyzedQueries.java
```diff
@@ -63,6 +94,91 @@ public static void main(String[] argv) throws IOException {
       return;
     }
 
+    // Read the analyzer from the command line, default is en 
+    if (args.language.equals("ar")) {
```
theyorubayesian, line 98: Can you use [AnalyzerMap](https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/analysis/AnalyzerMap.java) here instead?

JMMackenzie: What about `sw` or `te` though? Perhaps we should add a "whitespace" language, or just allow a default to be set? Otherwise using the AnalyzerMap wont allow all languages to be dealt with correctly I guess.

I lifted this from the SimpleSearcher just in case this helps with deciding how to deal with whitespace: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SimpleSearcher.java#L196

theyorubayesian: You could use a `getAnalyzer` method similar to [SimpleIndexer.getAnalyzer](https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/SimpleIndexer.java#L127-L150)

JMMackenzie: Hey @theyorubayesian - how about this?

src/main/java/io/anserini/util/DumpAnalyzedQueries.java
```diff
@@ -48,12 +52,35 @@ public static class Args {
...
+    public String language = "en";
+  }
+
+  static Analyzer getAnalyzer(Args args) {
+    try {
+      if (AnalyzerMap.analyzerMap.containsKey(args.language)) {
+        System.out.println("Using language-specific analyzer");
+        System.out.println("Language: " + args.language);
+        return AnalyzerMap.getLanguageSpecificAnalyzer(args.language);
+      } else if (args.language.equals("sw") || args.language.equals("te")) {
```
theyorubayesian, line 66: I think there's an analyzer for Telugu now so can you make it `yo` instead?

src/main/java/io/anserini/util/DumpAnalyzedQueries.java
```diff
@@ -48,12 +52,35 @@ public static class Args {
...
     public String output;
+
+    @Option(name = "-language", usage = "Analyzer Language")
+    public String language = "en";
+  }
+
+  static Analyzer getAnalyzer(Args args) {
+    try {
+      if (AnalyzerMap.analyzerMap.containsKey(args.language)) {
+        System.out.println("Using language-specific analyzer");
```
theyorubayesian, line 63: Can you swap these print statements for logs? See (https://github.com/castorini/anserini/blob/ec03f8cdf8bd9f51ce9f45233350cda32a6d23f6/src/main/java/io/anserini/analysis/AnalyzerMap.java#L28)

lintool (approved):

PR #2071, opened 2023-03-07
Add regressions for SPLADE++ ED/SD

lintool:
Entry point for README branch: https://github.com/castorini/anserini/tree/splade

Sample doc page: https://github.com/castorini/anserini/blob/splade/docs/regressions-msmarco-passage-splade-pp-ed.md
(all others similarly designed/formatted)


cadurosar (approved):
LGTM

PR #2075, opened 2023-03-10
megatokenizer-regression-docs

theyorubayesian:
Add regression docs for megatokenizer & composite tokenizer

### DIFF
--- a/docs/regressions-miracl-v1.0-ar-mega.md
+++ b/docs/regressions-miracl-v1.0-ar-mega.md
@@ -1,0 +1,61 @@
+# Anserini Regressions: MIRACL (v1.0) &mdash; Arabic
+
+**Models**: bag-of-words approaches using MegaTokenizer
+
+This page documents BM25 regression experiments for [MIRACL (v1.0) &mdash; Arabic](https://github.com/project-miracl/miracl) using MegaTokenizer.
+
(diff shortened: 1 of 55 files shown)


src/main/resources/docgen/templates/msmarco-doc-composite-tokenizer.template
```diff
@@ -0,0 +1,53 @@
+# Anserini Regressions: MS MARCO Document Ranking
```
lintool, line 1: There is no corresponding yaml here? Don't understand what this template is for?

lintool: Ah I see... So why just MS MARCO doc? Why not passage?

Do we want to circle back and do add composite {ms marco, dl19, dl20} x {passage, doc} in a separate PR?

theyorubayesian: Okay, sure! I'll try to get it out over the weekend

theyorubayesian: Do we need dl19 & dl20? I don't think we have those for `HuggingFaceTokenizerAnalyzer`

lintool (approved):

PR #2077, opened 2023-03-13
On-the-fly query encoder for Unicoil and Splade++ through ONNX

ArthurChen189:
**Notes**
- Unicoil query encoder for Msmarco Passage
- Splade++ (both ensemble-distilled and self-distilled) query encoder for Msmarco Passage
- Unittests added for both

**Usage**
```
target/appassembler/bin/SearchCollection -index <index path> \ 
-topics src/main/resources/topics-and-qrels/topics.msmarco-passage.dev-subset.txt \
-topicreader TsvInt -output <output path> \
-impact -pretokenized \
-encoder <UnicoilMsmarcoPassage/SpladePpEdMsmarcoPassage/SpladePpSdMsmarcoPassage>
```

**Jar Size**
`anserini-0.20.1-SNAPSHOT-fatjar.jar`: 191M


pom.xml
```diff
@@ -468,7 +468,17 @@
     <dependency>
       <groupId>ai.djl.huggingface</groupId>
       <artifactId>tokenizers</artifactId>
-      <version>0.21.0</version>
+      <version>0.18.0</version>
```
justram, line 471: Could you help to check if we can make this PR work for `0.21.0`?
`0.18.0` has some issues in other env, e.g., CentOS 7, so we bumped up the version to the latest one.

ArthurChen189: Resolved!
Just did some magic to the integration test.

src/main/java/io/anserini/search/queryencoder/SpladePpEdMsmarcoPassageEncoder.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private float[] toArray(List<Float> input) {
+        float[] output = new float[input.size()];
+        for (int i = 0; i < output.length; i++) {
+            output[i] = input.get(i);
+        }
+        return output;
+    }
+    private LinkedHashMap<String, Float> getTokenWeightMap(float[] computedWeights){
```
lintool, line 120: Do we need to specify implementation in the return sig? Why not just `Map<String, Float>`?

src/main/java/io/anserini/search/queryencoder/SpladePpEdMsmarcoPassageEncoder.java
```diff
@@ -0,0 +1,161 @@
...
+            tokenWeightMap.remove("[SEP]");
+        }
+        if (tokenWeightMap.containsKey("[PAD]")) {
+            tokenWeightMap.remove("[PAD]");
+        }
+
+        return tokenWeightMap;
+    }
+
+    private String generateEncodedQuery(LinkedHashMap<String, Float> tokenWeightMap) {
```
lintool, line 146: Same comment as above.

src/main/java/io/anserini/search/queryencoder/SpladePpEdMsmarcoPassageEncoder.java
```diff
@@ -0,0 +1,161 @@
+package io.anserini.search.queryencoder;
```
lintool, line 1: Add boilerplate - for all files also?

lintool (changes requested):
Some initial comments...

btw, why are there diffs in `tools/`?

src/main/java/io/anserini/search/queryencoder/SpladePpEdMsmarcoPassageEncoder.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.*;
+
+public class SpladePpEdMsmarcoPassageEncoder extends QueryEncoder{
+    static private final String MODEL_URL = "https://dl.dropboxusercontent.com/s/8eq7pjkm1lrwe1r/spladePpEd.onnx?dl=0";
+
+    static private final String VOCAB_URL = "https://dl.dropboxusercontent.com/s/2kgkvw6gm37ghc8/vocab.txt?dl=0RYd";
+
+    static private final String MODEL_NAME = "spladePpEdEncoder.onnx";
+
+    static private final String VOCAB_NAME = "vocab.txt";
```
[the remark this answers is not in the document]
ArthurChen189: Since all three encoders use the same vocab list, should we name them using the same name? (i.e., so the user would only need to download one file instead of 3)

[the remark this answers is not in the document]
ArthurChen189: Like naming all three vocab as 'WordpieceVocab.txt'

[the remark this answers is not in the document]
lintool: Space is free... so let's make three copies and give them different names?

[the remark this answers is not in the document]
ArthurChen189: Sounds good

src/test/java/io/anserini/onnx/splade/SpladePpEdEncoderQueryInferenceTest.java
```diff
@@ -0,0 +1,221 @@
+package io.anserini.onnx.splade;
```
lintool, line 1: Rename to match main classes?

src/main/java/io/anserini/search/query/QueryEncoder.java
```diff
@@ -0,0 +1,34 @@
+package io.anserini.search.query;
```
lintool, line 1: Boilerplate... for all files also?

lintool:
Minor nit... `Unicoil` should probably be `UniCoil`? Since the model is named "uniCOIL"?

src/test/java/io/anserini/search/query/SpladePlusPlusSelfDistilEncoderQueryInferenceTest.java
```diff
@@ -0,0 +1,259 @@
...
+import java.io.IOException;
+import java.net.URL;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.assertArrayEquals;
+
+public class SpladePlusPlusSelfDistilEncoderQueryInferenceTest {
```
lintool, line 19: two space indent?

src/test/java/io/anserini/search/query/UniCoilEncoderQueryInferenceTest.java
```diff
@@ -0,0 +1,112 @@
...
+                inputs.put("inputIds", OnnxTensor.createTensor(env, tokenIds));
+                try (Result results = session.run(inputs)) {
+                    float[] computedWeights = flatten(results.get(0).getValue());
+                    assertArrayEquals(expectedWeights, computedWeights, 1e-5f);
+                }
+            }
+        }
+    }
+
+    private float[] flatten(Object obj) {
```
lintool, line 93: two space indent?

lintool (changes requested):
+ two-space indent everywhere please
+ make sure new files have boilerplate
+ latest PR still says "Submodule tools" removed.

```
    @Option(name = "-weightRange", usage = "range of weights for sparse query encoding")
    public int weightRange = 5;

    @Option(name = "-quantRange", usage = "range of quantization for sparse query encoding")
    public int quantRange = 256;
```

I had commented before - shouldn't each model know it's own settings? and hence can be hard-coded?

ArthurChen189:
- new boilerplate added for tests
- two-space indentation applied
- submodules recovered
- hardcoded quant range + weight range applied

src/test/java/io/anserini/search/query/UniCoilEncoderQueryInferenceTest.java
```diff
@@ -0,0 +1,135 @@
...
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertArrayEquals;
+
+public class UniCoilEncoderQueryInferenceTest {
+  static private final String MODEL_URL = "https://dl.dropboxusercontent.com/s/39jqt27b6efuyry/UnicoilEncoder.onnx?dl=0";
```
lintool, line 20: Is there a `https://rgw.cs.uwaterloo.ca/pyserini/data` URL for this?

lintool (approved):

Issue #1890, opened 2022-05-29
Better implementation of JsonVectorCollection than the "fake words" approach


lintool:
To index JsonVectorCollection sparse vectors, we currently use the "fake words" trick - just duplicate the word _X_ times, where _X_ is the score. This might be a better solution: https://lucene.apache.org/core/8_11_0/analyzers-common/org/apache/lucene/analysis/miscellaneous/DelimitedTermFrequencyTokenFilter.html

tteofili:
I did an experiment for this a while ago, the behavior for some reason is not exactly the same (expected) as the fw one. I can dig into this.

JMMackenzie:
Possibly different due to #1843?

PR #2080, opened 2023-03-23
Faster indexing for learned sparse retrieval 

thongnt99:
Related to #1890
On-going work: Using FeatureField to directly index terms and weights

The indexing works and returns the same metrics as the token repeating method, but three tests (for the repeating method) are currently failing. Please let me know how to fix the tests or create new tests. 

Indexing: 
```bash 
./anserini-lsr/target/appassembler/bin/IndexCollection \
-collection JsonTermWeightCollection \
-input collections/msmarco-passage/lsr_collection_jsonl \
-index indexes/msmarco-passage/lsr-index-msmarco \
-generator TermWeightDocumentGenerator \
-threads 60 -impact -pretokenized
```
Retrieval: 
```bash
./anserini-lsr/target/appassembler/bin/SearchCollection \
-index path_to_index \
-topics path_to_topic \
-topicreader TsvString \
-output path_to_output_file \
-impact -pretokenized -hits 1000 -parallelism 60 
```


lintool:
Hi @thongnt99 very interesting and thanks for the PR!

Can you provide a sense of the performance improvement?

thongnt99:
Hi @lintool , 

These are some comparison points I collected from [our recent reproduction attempt](https://arxiv.org/pdf/2303.13416v1.pdf) with LSR methods. 
The degree of speed up would depend on the magnitude of document term weights, but at least twice faster than the term repeating method. We saw, for example,  a huge improvement for indexing EPIC since EPIC does not use sparse regularizers during training, therefore produces generally larger weights (the term repeating method has to repeat more).  
| LSR method | Old | New|
|-------------|----|-----|
| QMLP_DMLM|0:10:25|0:04:09|
| EPIC (top_k=400) | 1:23:53	| 0:04:02|
|Splade (0.01, 0.08) | 0:17:41	| 0:03:52|
|uniCOIL	| 0:05:11 |	0:02:18|

MXueguang:
> Hi @lintool ,
> 
> These are some comparison points I collected from [our recent reproduction attempt](https://arxiv.org/pdf/2303.13416v1.pdf) with LSR methods. The degree of speed up would depend on the magnitude of term weights, but at least twice faster than the term repeating method. We saw, for example, a huge improvement for indexing EPIC since EPIC does not use sparse regularizers during training, therefore produces generally larger weights (the term repeating method has to repeat more).
> 
> LSR method	Old	New
> QMLP_DMLM	0:10:25	0:04:09
> EPIC (top_k=400)	1:23:53	0:04:02
> Splade (0.01, 0.08)	0:17:41	0:03:52
> uniCOIL	0:05:11	0:02:18

this is cool!

src/main/java/io/anserini/collection/JsonTermWeightCollection.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+
+/**xxw
```
lintool, line 27: `xxw`?

src/main/java/io/anserini/collection/JsonTermWeightCollection.java
```diff
@@ -0,0 +1,79 @@
...
+public class JsonTermWeightCollection extends DocumentCollection<JsonTermWeightCollection.Document> {
+  public JsonTermWeightCollection(Path path) {
+    this.path = path;
+  }
+
+  @Override
+  public FileSegment<JsonTermWeightCollection.Document> createFileSegment(BufferedReader bufferedReader) throws IOException {
+    return new JsonTermWeightCollection.Segment<>(bufferedReader);
+  }
+  @Override
```
lintool, line 39: space between method declarations... and below also.

src/main/java/io/anserini/index/generator/TermWeightDocumentGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+ * Converts a {@link SourceDocument} into a Lucene {@link Document}, ready to be indexed.
+ *
+ * @param <T> type of the source document
+ */
+public class TermWeightDocumentGenerator<T extends  SourceTermWeightDocument & SourceDocument> implements LuceneDocumentGenerator<T> {
+  protected IndexCollection.Args args;
+
+  protected TermWeightDocumentGenerator() {
+  }
+  /**
```
lintool, line 39: space

lintool (commented):
Initial comments.

lintool:
Instead of `TermWeightDocument`... why not just call it `VectorDocument`? Vector as `Map<String,Float>` seems pretty intuitive?

thongnt99:
> Instead of `TermWeightDocument`... why not just call it `VectorDocument`? Vector as `Map<String,Float>` seems pretty intuitive?
Yes, I also think that TermWeightDocument isn't an ideal name. Probably `SparseVectorDocument` is more suitable than `VectorDocument`. The formers says that we should store indices (or terms) and values (similar to SparseMatrix or DenseMatrix format).

lintool:
I like `SparseVectorDocument`!

thongnt99:
@lintool
I changed class names and fixed issues in your previous comments. 

```
./anserini-lsr/target/appassembler/bin/IndexCollection \
-collection JsonSparseVectorCollection \
-input collections/msmarco-passage/lsr_collection_jsonl \
-index indexes/msmarco-passage/lsr-index-msmarco \
-generator SparseVectorDocumentGenerator \
-threads 60 -impact -pretokenized
```

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -1250,7 +1251,10 @@ public <K> ScoredDocuments search(IndexSearcher searcher, K qid, String queryStr
 
     if (args.sdm) {
       query = new SdmQueryGenerator(args.sdm_tw, args.sdm_ow, args.sdm_uw).buildQuery(Constants.CONTENTS, analyzer, queryString);
-    } else {
+    } else if (args.impact){
+        query = new BagOfWordsQueryGenerator().buildFeatureQuery(Constants.CONTENTS, analyzer, queryString);
+    }
+    else {
```
lintool, line 1257: else with previous line, e.g.,

```
} else {
```

src/main/java/io/anserini/search/query/FeatureGenerator.java
```diff
@@ -0,0 +1,15 @@
+package io.anserini.search.query;
+
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.search.Query;
+
+public interface FeatureGenerator{
+    /**
```
lintool, line 7: Two space indent.

lintool (commented):
How about some tests?

thongnt99:
> How about some tests?

@lintool  I am gonna add the tests after ECIR.

Issue #2069, opened 2023-03-01
ClassCastException when indexing ACL Anthology


ygorg:
When following the ["Indexing the ACL Anthology with Anserini"](https://github.com/castorini/anserini/blob/master/docs/acl-anthology.md) the actual indexing raises the following traceback (see [AclAnthology.java:158](https://github.com/castorini/anserini/blob/1609686b047f76045e3735cb5aca94d99cdd7947/src/main/java/io/anserini/collection/AclAnthology.java#L158)):
```
java.lang.ClassCastException: class com.fasterxml.jackson.databind.node.TextNode cannot be cast to class com.fasterxml.jackson.databind.node.ArrayNode (com.fasterxml.jackson.databind.node.TextNode and com.fasterxml.jackson.databind.node.ArrayNode are in unnamed module of loader 'app')
	at io.anserini.collection.AclAnthology$Document.<init>(AclAnthology.java:158) ~[anserini-0.20.0-fatjar.jar:?]
	at io.anserini.collection.AclAnthology$Segment.readNext(AclAnthology.java:115) ~[anserini-0.20.0-fatjar.jar:?]
	at io.anserini.collection.FileSegment$1.hasNext(FileSegment.java:136) ~[anserini-0.20.0-fatjar.jar:?]
	at io.anserini.index.IndexCollection$LocalIndexerThread.run(IndexCollection.java:298) [anserini-0.20.0-fatjar.jar:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
	at java.lang.Thread.run(Thread.java:829) [?:?]
```
There seem to have been a change on how the venues are processed in the [acl-org/acl-anthology](https://github.com/acl-org/acl-anthology) which breaks anserini's [`collection.ACLAnthology`](https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/collection/AclAnthology.java).

My use case is using [castorini/covidex]() with ACL documents. My workaround was to index the bibtex of the aclanthology, but there is a lot of brackets and LaTeX things in the text, so I'd rather go with this solution.

Steps to reproduce:
```bash
git clone https://github.com/acl-org/acl-anthology
conda create -n acl_anth python=3.8
conda activate acl_anth
cd acl-anthology
pip install -r bin/requirements.txt
[...]

PR #2084, opened 2023-03-27
Issues when indexing AclAnthology

ygorg:
This PR is linked to #2069.

There are some issues when following the ["Indexing the ACL Anthology with Anserini"](https://github.com/castorini/anserini/blob/master/docs/acl-anthology.md) tutorial:

1. The `venues` attribute is now `venue` and is stored in paper rather than volume.
2. The `page_first` and `page_last` attributes are sometimes in roman notation and thus strings, not number. I change the schema, because translating roman to int should happen in `acl-anthology`.
3. Some files are too big for `fasterxml.jackson`. [By default](https://github.com/FasterXML/jackson-dataformats-text/tree/2.15/yaml#maximum-input-yaml-document-size-3-mb) files should be less that 3mb (~1,345,000 char). I increased this size to 10mb.
4. The script `create_hugo_yaml.py` produces a yaml file containing aliases and references (to prevent data redundancy). These cannot be parsed by `fasterxml.jackson` (https://github.com/FasterXML/jackson-dataformats-text/issues/98). One way to prevent this is modifying the script to prevent [this behavior](https://ttl255.com/yaml-anchors-and-aliases-and-how-to-disable-them/#override) (see the updated doc). An other way is to postprocess the data in `acl-anthology/build/data` using 
```bash
pip install ruamel.yaml.cmd
mv volumes.yaml volumes_old.yaml
yaml merge-expand volumes_old.yaml volumes.yaml
mv papers papers_old
for yaml_old in papers_old/*.yaml; do  # this is ~10minutes !
  yaml_new=${yaml_old##*/}
  yaml merge-expand $yaml_old papers/$yaml_new
done
```

Point number 4 is hacky, maybe there is a way to preprocess and remove the aliases in java in AclAnthology.java ? Maybe this dataset should be read using another format (jsonl, xml, bibtex) ?
I still need to update the unittests.


ygorg:
I updated the unittests with the same sample documents as before, but from a recent extract of acl-anthology.

aryamancodes:
I followed the instructions on this PR and the related issue (including both the "hacks" described in point 4) and was able to successfully index ACL Anthology. The unit tests ran successfully too. Just a small nitpick with the post-processing script, the correct version should look like:
```
pip install ruamel.yaml.cmd
mv volumes.yaml volumes_old.yaml
yaml merge-expand volumes_old.yaml volumes.yaml
mv papers papers_old

# since we renamed the papers dir, it no longer exists
mkdir papers 

for yaml_old in papers_old/*.yaml; do  # this is ~10minutes !
  yaml_new=${yaml_old##*/}
  yaml merge-expand $yaml_old papers/$yaml_new
done
```

lintool:
@ygorg thoughts? let's sort through these final issues and then merge in your PR?

ygorg:
@aryamancodes My comment wasn't clear, the solution you used is very slow, my primary recommandation was to modify `create_hugo_yaml.py` with [this line](https://github.com/castorini/anserini/pull/2084/files#diff-2cc23ee7be05d2c1f405fe4e78a5eea20bedf154834c870d589d03ffb4f816a9R36) `Dumper.ignore_aliases = lambda self, data: True`
Which simplifies the process so much.

lintool (approved):

lintool:
Thanks @ygorg for the contribution!

PR #2089, opened 2023-04-01
Culling PRF regressions with MS MARCO sparse judgments

lintool:
This is an effort to reduce the running times of our regressions.
Commenting out instead of removing so that we can re-create if needed for experiments.


lintool:
Note that we don't need to worry about the underlying code breaking because we're still testing PRF with dense judgments.

theyorubayesian (approved):

Issue #2066, opened 2023-02-26
Naming for index and search classes


lintool:
We currently have three sets of indexing/search entry points:
1. "normal" inverted indexing and search: `io.anserini.index.IndexCollection`, `io.anserini.search.SearchCollection`
2. indexing and searching dense vectors using HNSW.
3. indexing and searching dense vectors using inverted indexes, with LSH or FW analyzers, currently in the `io.anserini.ann` package.

We need a consistent naming scheme... How about:

For 1, keep as before:
+ `io.anserini.index.IndexCollection`
+ `io.anserini.search.SearchCollection`

For 2:
+ `io.anserini.index.IndexHnswDenseVectors`
+ `io.anserini.search.SearchHnswDenseVectors`

For 3:
+ `io.anserini.index.IndexInvertedDenseVectors`
+ `io.anserini.search.SearchInvertedDenseVectors`

tteofili:
the proposal looks good to me

PR #2091, opened 2023-04-04
Better naming for index and search classes

tteofili:
This PR aims at resolving #2066. 
The names of the indexing-related classes all start with `Index` whereas searching-related class names start with`Search`. 
The `IndexInvertedDenseVectors` and `SearchInvertedDenseVectors` have also been moved to `io.anserini.index` and `io.anserini.search` packages accordingly.
Runners' configurations within `pom.xml` have been changed to point to the right classes, however, for the sake of backward compatibility (also considering documentation), the names of the resulting commands have been kept the same; however we can change that if we prefer new naming scheme to be reflected.


src/main/java/io/anserini/ann/EvalInvertedDenseVectors.java
```diff
@@ -53,7 +55,7 @@
 
 import static org.apache.lucene.search.BooleanClause.Occur.SHOULD;
 
-public class ApproximateNearestNeighborEval {
+public class EvalInvertedDenseVectors {
```
lintool, line 58: I think the point was to get rid of the `io.anserini.ann` package... so maybe move this to `io.anserini.search` - and rename to `Evaluate`? Typically, we'd write eval util classes in Python...

src/main/java/io/anserini/util/IOUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+public class IOUtils {
+
+  private IOUtils() {
+  }
+
+  public static Map<String, List<float[]>> readGloVe(File input) throws IOException {
```
lintool, line 31: I don't think this needs to be generalized and abstracted in its own class? Evaluation of Glove seems pretty outdated...

src/main/java/io/anserini/ann/EvalInvertedDenseVectors.java
tteofili: ok, if we're to remove the `ann` package we also need to move `FeatureVectorTokenizer`,  and both`fw` and `lexlsh' packages, e.g., into `io.anserini.analysis` (as they are tokenizers+filters in the end).

lintool: Agreed.

lintool (approved):

PR #2094, opened 2023-04-05
Add initial ONNX regressions

lintool:
Building on @ArthurChen189 's recent PR.

Some initial ONNX regressions so we can start playing around with it...

These work:

```bash
nohup python src/main/python/run_regression.py --index --verify --search --regression msmarco-passage-splade-pp-ed-onnx >& logs/log.msmarco-passage-splade-pp-ed-onnx &
nohup python src/main/python/run_regression.py --index --verify --search --regression msmarco-passage-splade-pp-sd-onnx >& logs/log.msmarco-passage-splade-pp-sd-onnx &
```

As well as:

```bash
nohup python src/main/python/run_regression.py --search-pool 1 --verify --search --regression dl19-passage-splade-pp-ed-onnx >& logs/log.dl19-passage-splade-pp-ed-onnx &
nohup python src/main/python/run_regression.py --search-pool 1 --verify --search --regression dl20-passage-splade-pp-ed-onnx >& logs/log.dl20-passage-splade-pp-ed-onnx &

nohup python src/main/python/run_regression.py --search-pool 1 --verify --search --regression dl19-passage-splade-pp-sd-onnx >& logs/log.dl19-passage-splade-pp-sd-onnx &
nohup python src/main/python/run_regression.py --search-pool 1 --verify --search --regression dl20-passage-splade-pp-sd-onnx >& logs/log.dl20-passage-splade-pp-sd-onnx &
```

I'm encountering some concurrency issues, so without `--search-pool 1`, the regressions fail. Will need to debug more.

cc @cadurosar

### DIFF
--- a/src/main/resources/regression/dl19-passage-splade-pp-ed-onnx.yaml
+++ b/src/main/resources/regression/dl19-passage-splade-pp-ed-onnx.yaml
@@ -1,0 +1,93 @@
+---
+corpus: msmarco-passage-splade-pp-ed
+corpus_path: collections/msmarco/msmarco-passage-splade-pp-ed
+
+download_url: https://rgw.cs.uwaterloo.ca/pyserini/data/msmarco-passage-splade-pp-ed.tar
+download_checksum: e489133bdc54ee1e7c62a32aa582bc77
+
+index_path: indexes/lucene-index.msmarco-passage-splade-pp-ed/
+collection_class: JsonVectorCollection
+generator_class: DefaultLuceneDocumentGenerator
+index_threads: 16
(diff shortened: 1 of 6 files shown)


ArthurChen189:
Just want to make sure we are referring to the same error (on a different OSes), only the RM3 runs failed?
image

cadurosar:
The weirdest part for me is that is not just a small loss of effectiveness, is almost like it is losing the original representation

lintool:
> The weirdest part for me is that is not just a small loss of effectiveness, is almost like it is losing the original representation

I think it's a weird concurrency bug...

PR #2103, opened 2023-04-13
Fix paths for docs, src

ArthurChen189:
Changed paths from `src/main/resources/topics-and-qrels` to `tools/topics-and-qrels`


src/test/java/io/anserini/eval/RelevanceJudgmentsTest.java
```diff
@@ -35,7 +35,7 @@ public int getQrelsCount(RelevanceJudgments qrels) throws IOException{
 
   @Test
   public void testRobust04() throws IOException{
-    RelevanceJudgments qrels = new RelevanceJudgments("src/main/resources/topics-and-qrels/qrels.robust04.txt");
+    RelevanceJudgments qrels = new RelevanceJudgments("tools/topics-and-qrels/qrels.robust04.txt");
```
lintool, line 38: Wait, how are tests passing now?

lintool (changes requested):
Can we revert the changes in `src/main/python` back to the original paths? Rationale is that those are mostly one-off scripts that aren't being used anymore... I'd rather they be left alone - we can fix them if we decided to use them again. That way `git blame` will be (more) accurate.

src/test/java/io/anserini/eval/RelevanceJudgmentsTest.java
ArthurChen189: Previously, all topics were downloaded to the cache before it was searched locally.
After the fix, files are searched locally first and then download from the web if not found.

src/main/resources/ecir2019_axiomatic/collections.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 search_command: target/appassembler/bin/SearchCollection
-topic_root: src/main/resources/topics-and-qrels/
```
lintool, line 3: revert?

src/main/resources/fine_tuning/collections.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 search_command: target/appassembler/bin/SearchCollection
-topic_root: src/main/resources/topics-and-qrels/
```
lintool, line 3: revert?

lintool (changes requested):

lintool:
please revert these two files?

lintool (approved):

PR #2107, opened 2023-04-22
Bump Onnx to v1.12.1

aryamancodes:
Bumping to version 1.12.1 since 1.11.0 is not compatible with M1 Macs, [as described here](https://github.com/microsoft/onnxruntime/issues/11054)


lintool:
@ArthurChen189 is this okay with you?

ArthurChen189 (approved):
that works

Issue #2122, opened 2023-05-23
Add ability to parse raw text into docvectors on-the-fly for impact indexes


lintool:
As part of #1984 we added the ability to re-create docvectors on-the-fly so that we didn't need to store docvectors in the index (but we need to store the raw text, which is smaller).

This feature hasn't been exposed for impact indexes. We should do it so models like uniCOIL and SPLADE can benefit.

lintool:
Building three types of indexes, using uniCOIL as an example:

```
# "base"
target/appassembler/bin/IndexCollection \
  -collection JsonVectorCollection \
  -input /mnt/collections/msmarco/msmarco-passage-unicoil \
  -index indexes/lucene-index.msmarco-passage-unicoil/ \
  -generator DefaultLuceneDocumentGenerator \
  -threads 16 -impact -pretokenized -optimize &

# Store docvectors
target/appassembler/bin/IndexCollection \
  -collection JsonVectorCollection \
  -input /mnt/collections/msmarco/msmarco-passage-unicoil \
  -index indexes/lucene-index.msmarco-passage-unicoil.docvectors/ \
  -generator DefaultLuceneDocumentGenerator \
  -threads 16 -impact -pretokenized -storeDocvectors -optimize &

# Store raw text
target/appassembler/bin/IndexCollection \
  -collection JsonVectorCollection \
  -input /mnt/collections/msmarco/msmarco-passage-unicoil \
  -index indexes/lucene-index.msmarco-passage-unicoil.text/ \
  -generator DefaultLuceneDocumentGenerator \
  -threads 16 -impact -pretokenized -storeRaw -optimize &
```

Saves a lot of space to store only raw text:

```
$ du -h indexes/ | grep unicoil
1.3G	indexes/lucene-index.msmarco-passage-unicoil
47G	indexes/lucene-index.msmarco-passage-unicoil.docvectors
8.0G	indexes/lucene-index.msmarco-passage-unicoil.text
```

lintool:
@AileenLin to help you out - this is currently what's not working on the Pyserini end:

```
python -m pyserini.search.lucene \
  --threads 16 --batch-size 128 \
  --index ../anserini/indexes/lucene-index.msmarco-passage-unicoil.docvectors \
  --topics dl19-passage-unicoil \
  --output runs/run.dl19-rocchio.txt \
  --hits 1000 --impact --rocchio
```

Ultimately, I want to make this work.

AileenLin:
[...]

PR #2133, opened 2023-06-03
issue #2122 Add ability to parse raw text into docvectors on-the-fly for impact indexes

AileenLin:


src/main/java/io/anserini/search/SimpleImpactSearcher.java
```diff
@@ -97,12 +104,23 @@ protected SimpleImpactSearcher() {
   }
 
   /**
-   * Creates a {@code SimpleImpactSearcher}.
+   * Creates a {@code SimpleSearcher}.
```
lintool, line 107:
```suggestion
   * Creates a {@code SimpleImpactSearcher}.
```

lintool:
We have an error getting test coverage: https://app.codecov.io/gh/castorini/anserini/pull/2133

src/main/java/io/anserini/rerank/RerankerContext.java
```diff
@@ -46,6 +47,20 @@ public RerankerContext(IndexSearcher searcher, K queryId, Query query, String qu
     this.queryTokens = queryTokens;
     this.filter = filter;
     this.searchArgs = searchArgs;
+    this.weightedQueryTokens = null;
+  }
+
+  public RerankerContext(IndexSearcher searcher, K queryId, Query query, String queryDocId, String queryText,
```
lintool, line 53: hi @AileenLin can you explain why we need to add this code path?

`SearchCollection` is able to perform Rocchio/RM3 with impact exists, right?
E.g., https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SearchCollection.java#L1152

AileenLin: the SimpleImpact search method take in a Map of token frequency, we can either add the weitedQueryToken here, or add another search method to process text in SimpleImpact. But if for pretokenized Map of token frequency, I prefered the second way to do it

lintool: In that case, would it make sense for `SearchCollection` and `SimpleImpactSearcher` to use the same code path?

Issue #2127, opened 2023-05-30
Verify ONNX repo, add ONNX model conversion documentation


lintool:
@ArthurChen189

lintool:
@cadurosar did repo, merged in as #2128 -

+ https://github.com/castorini/anserini/blob/master/docs/regressions-msmarco-passage-splade-pp-ed-onnx.md
+ https://github.com/castorini/anserini/blob/master/docs/regressions-msmarco-passage-splade-pp-sd-onnx.md

@cadurosar just for completeness, can you verify repro for DL19/20 also?

+ https://github.com/castorini/anserini/blob/master/docs/regressions-dl19-passage-splade-pp-ed-onnx.md
+ https://github.com/castorini/anserini/blob/master/docs/regressions-dl19-passage-splade-pp-sd-onnx.md
+ https://github.com/castorini/anserini/blob/master/docs/regressions-dl20-passage-splade-pp-ed-onnx.md
+ https://github.com/castorini/anserini/blob/master/docs/regressions-dl20-passage-splade-pp-sd-onnx.md

lintool:
Defunct, closing.

PR #2136, opened 2023-06-05
Add ONNX model conversion docs

ArthurChen189:
Add MD file for converting and optimizing SPLADE++ from HuggingFace to ONNX

Related to issue #2127


cadurosar:
Looks good to me, maybe add a test to see if the outputs are close to what is expected? What do you think @lintool ?

lintool:
haha, I usually never say no to unit tests, but this does belong in Anserini... but since the code is in Python, we don't have testing framework set up... so I think good to merge?

lintool:
I think the tests belong on the Pyserini end once we set up the bindings there?

cadurosar:
Yes, I agree is more on the pyserini side, ok seems good to me them.

cadurosar (approved):

PR #2148, opened 2023-07-16
Add ability to parse raw text into docvectors on-the-fly for impact indexes #2122

AileenLin:
#2122

### DIFF
--- a/src/main/java/io/anserini/search/SimpleImpactSearcher.java
+++ b/src/main/java/io/anserini/search/SimpleImpactSearcher.java
@@ -125,7 +125,7 @@ public class SimpleImpactSearcher implements Closeable {
    */
   public SimpleImpactSearcher(String indexDir, String queryEncoder) throws IOException {
     this(indexDir, IndexCollection.DEFAULT_ANALYZER);
-    this.setOnnxQueryEncoder(queryEncoder);
+    this.set_onnx_query_encoder(queryEncoder);
   }
 
   /**
@@ -138,7 +138,7 @@ public class SimpleImpactSearcher implements Closeable {
    */
   public SimpleImpactSearcher(String indexDir, String queryEncoder, Analyzer analyzer) throws IOException {
     this(indexDir, analyzer);
-    this.setOnnxQueryEncoder(queryEncoder);
+    this.set_onnx_query_encoder(queryEncoder);
   }
 
   /**
@@ -177,7 +177,7 @@ public class SimpleImpactSearcher implements Closeable {
    * 
    * @param encoder the query encoder
    */
-  public void setOnnxQueryEncoder(String encoder) {
+  public void set_onnx_query_encoder(String encoder) {
     if (emptyEncoder()) {
       try {
         this.queryEncoder = (QueryEncoder) Class.forName("io.anserini.search.query." + encoder + "QueryEncoder")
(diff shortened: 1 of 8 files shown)


src/main/java/io/anserini/search/SimpleImpactSearcher.java
```diff
@@ -143,6 +192,136 @@ private boolean empty_encoder(){
...
+  /**
+   * Returns the analyzer used.
+   *
+   * @return analyzed used
+   */
+  public Analyzer get_analyzer(){
+    return this.analyzer;
+  }
+
+    /**
```
lintool, line 213:
```suggestion
  /**
```

src/main/java/io/anserini/search/SimpleImpactSearcher.java
```diff
@@ -143,6 +192,136 @@ private boolean empty_encoder(){
...
+  /**
+   * Disables RM3 query expansion.
+   */
+  public void unset_rm3() {
+    this.useRM3 = false;
+    cascade = new RerankerCascade();
+    cascade.add(new ScoreTiesAdjusterReranker());
+  }
+
+    /**
```
lintool, line 231: fix indent, same as above.

lintool (commented):
I thought we were going to change type to `Map<String, Int>` in places where it makes sense?

src/main/java/io/anserini/search/SimpleImpactSearcher.java
```diff
@@ -429,48 +501,47 @@ public Map<String, Float> encode_with_onnx(String queryString) throws OrtExcepti
...
    * @param k number of hits
    * @return array of search results
    * @throws IOException if error encountered during search
+   * @throws OrtException if error encountered during search
    */
-  public Result[] search(Map<String, Float> q, int k) throws IOException, OrtException {
-    Query query = generator.buildQuery(Constants.CONTENTS, q);
-    String encodedQuery = encode_with_onnx(q);
+  public Result[] search(Map<String, Integer> encoded_q, int k) throws IOException, OrtException {
+    Map<String, Float> float_encoded_q = intToFloat(encoded_q);
```
lintool, line 542: I'm still not very happy with this... we get a `Map<String, Integer>`... and then we covert to `Map<String, Float>`... then it comes back as a flat String.

Should we be redesigning the QueryEncoder?

AileenLin: just saw this. this line is not for query encoder thing, its for QueryGenerator to build query, which takes in float weight. so here I just convert Int to Float, for example 10 -> 10.0

src/main/java/io/anserini/search/SimpleImpactSearcher.java
```diff
@@ -545,6 +625,26 @@ public Result[] search(Map<String, Integer> encoded_q, int k) throws IOException
...
+   *
+   * @param q string query
+   * @param k number of hits
+   * @return array of search results
+   * @throws IOException if error encountered during search
+   * @throws OrtException if error encountered during search
+   */
+  public Result[] search(String q, int k) throws IOException, OrtException {
+    // make encoded query from raw query
+    Map<String, Integer> encoded_q = encode_with_onnx(q);
```
lintool, line 639: Same, camelCase?

lintool (approved):

PR #2161, opened 2023-08-08
Make it possible to index 2048 sized vectors with HNSW

tteofili:
Small hack to make it possible to index 2048 sized vectors.
To use it within `appassembler` generated scripts, the fatjar should replace the single jars within `target/appassembler/repo`.


lintool:
superseded by #2235

PR #2168, opened 2023-08-24
Add an entry in the "Reproduction Log" for ugrad onboarding

yilinjz:
**OS**: Ubuntu 18.04 (WSL)
 
**Hardware**:
- Processor: Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz   2.90 GHz
-  RAM: 32.0 GB
- System type: 64-bit operating system, x64-based processor
 
**Python**: Python 3.6.9

**Result**: I was able to replicate the MS MARCO Tour (i.e. download & extra the data, data munging and analyze qrels files) on my machine with the above settings. Outputs are the same as listed in the "start-here" document.


lintool:
hi @yilinjz - can you update your PR with some details - e.g., your OS, hardware, etc., and some indication of whether it worked? See previous PRs as an example.

yilinjz:
Hi Professor Lin @lintool, I've updated the PR with the desired information.

lintool (approved):

PR #2169, opened 2023-08-24
Add an entry in the Passage Ranking Experiment Reproduction Log for ugrad onboarding

yilinjz:
**OS**: macOS Ventura Version 13.5
**Hardware**: M2 MacBook Pro (2023), 16 GB RAM
 
**Java**: 20.0.2
**Maven**: 3.9.4
**Python**: 3.11.4

**Result**: I was able to replicate the Indexing, Retrieval & Evaluation steps on my machine with the above settings. Outputs are the same as listed in the document. I also repeat the steps for two other queries (one with MRR@10=0.33 and the other with MRR@10=0.00) and their MRR@10 scores agree with the ranking lists.

**Additional Comment**: There might be a typo in the start-here.md file. Given the context, it seems 1048585 is the query id rather than the document id.


lintool:
Great catch on the typo!

lintool (approved):

PR #2185, opened 2023-09-03
Results reproduced by @lucedes27

lucedes27:
Fixed minor typo: 'except' --> 'expect' ('accept' also would have worked)

Using MacBook Pro 2019 running macOS Ventura 13.4.1

Everything worked mostly fine.
For the 'python' terminal commands, I had to replace 'python' with 'python3' because I hadn't installed Anaconda


lintool:
Good typo catch!

lintool (approved):

PR #2220, opened 2023-10-11

shayanbali:
### setup:

- Operating System: macOS Ventura Version 13.5
- Chip: Apple M2 Pro
- Memory: 16 GB
- Environment: Mac Terminal
- Python: 3.9.6

### status:
everything worked

### DIFF
--- a/docs/experiments-msmarco-passage.md
+++ b/docs/experiments-msmarco-passage.md
@@ -79,7 +79,7 @@ The output queries file `collections/msmarco-passage/queries.dev.small.tsv` shou
 
 In building a retrieval system, there are generally two phases:
 
-+ In the **indexing** phase, an indexer takes the document collection (i.e., corpus) and builds an index, which is a data structure that supports effcient retrieval.
++ In the **indexing** phase, an indexer takes the document collection (i.e., corpus) and builds an index, which is a data structure that supports efficient retrieval.
 + In the **retrieval** (or **search**) phase, the retrieval system returns a ranked list given a query _q_, with the aid of the index constructed in the previous phase.
 
 (There's also a training phase when we start to discuss models that _learn_ from data, but we're not there yet.)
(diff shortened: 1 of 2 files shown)


shayanbali (commented):
There is a typo mistake in Indexing section (effcient -> efficient).

Providing a detailed example of MRR@10 in this document could be helpful for better understanding

lintool:
Thanks for catching the typo!

> Providing a detailed example of MRR@10 in this document could be helpful for better understanding.

What exactly do you mean? MRR is explained here, right? https://github.com/castorini/anserini/blob/master/docs/experiments-msmarco-passage.md#evaluation

Do you have any suggestions for how to improve?

lintool (approved):

shayanbali:
An example like this can help better understanding:

Scenario:

Imagine we have a search engine that retrieves web pages based on user queries. We want to evaluate the performance of this search engine using MRR@10.

Steps:

### **Generate Sample Queries:**

Let's say we have a set of queries that users might input into the search engine. For instance:

Query 1: "How to bake a cake"
Query 2: "Best hiking trails near me"
Query 3: "Python programming tutorials"
### **Run Queries and Retrieve Results:**

We run these queries through our search engine, and it returns a ranked list of results for each query. For simplicity, let's assume each query returns 10 results.

For Query 1, the top 10 results might be:

Recipe website with detailed instructions on baking a cake (Relevant)
Online store selling cake decorations (Irrelevant)
Blog post about cake decorating techniques (Irrelevant)
...
Cooking forum discussing various recipes (Irrelevant)
Similarly, we have ranked lists for Query 2 and Query 3.

### **Determine Relevance:**

For each query, we manually assess the relevance of the top result. Let's assume:

For Query 1, the most relevant result is at position 1 (as it directly answers the query).
For Query 2, the most relevant result is at position 5 (as it provides information about local hiking trails).
For Query 3, the most relevant result is at position 3 (as it's a high-quality Python programming tutorial).
### **Calculate MRR@10:**

Using the relevance positions we determined, we calculate the Mean Reciprocal Rank at 10:

MRR@10 = (1/1 + 1/5 + 1/3) / 3 ≈ 0.54

lintool:
Hi @shayanbali thanks for the suggestion, but IMO this would be over-explaining and breaking the flow of the current narrative.

Issue #679, opened 2019-06-05
Upgrade to Lucene8


lintool:
Next decision we need to make... upgrade to Lucene8? It's a matter of _when_.

Lucene8 promises to be faster: https://github.com/castorini/anserini/blob/master/docs/lucene7-vs-lucene8.md

However, Lucene8 yields a whole slew of changes to regression numbers... and will likely break a bunch of things.

My recommendation is to rip the bandaid off, accept that things are going to be broken for a bit, and then gradually fix. EMNLP and CIKM deadlines just passed; WSDM isn't for a while; ACL is far off in the distance.

The direct effects I can foresee are:
+ Ongoing MS MARCO experiments (affecting @rodrigonogueira4 and @Victor0118)
+ Disruption for SIGIR demos (affecting @r-clancy @zeynepakkalyoncu @ljj7975 )
+ OSIRRC can use the v0.5.0 release, so no effect here.
+ BERTserini needs to be retuned (affecting @Victor0118 @LuchenTan); but NAACL demo just passed.
+ Birch effectiveness numbers might change (but for OSIRRC can use v0.5.0 release) - affecting @zeynepakkalyoncu and @Victor0118 
+ TREC DL is a bit off, so we can enter with updated version.
+ @emmileaf has a lot of refactoring planned.
+ Lucene8 upgrade will actually make ES integration easier, I think? (@charW and @r-clancy)

Anything else I missed? Downside of a large shared code-base used by the entire group... lots of effects...

cc @Peilin-Yang in case he has an opinion?

Thoughts, everyone? 👍 or 👎 or 🤷‍♂ or 🤷‍♀ ?

r-clancy:
+1 for ripping the band-aid off, seems like a good time to do it.

- SIGIR demos - we can use the `0.5.0` release as the new index type in Lucene 8 will likely effect the results
- `anserini-docker` is now using `0.5.0`
- ElasticSearch - this should let us use the most recent ES version rather than an older one

charW:
+1. Using ES 7+ right now would be pretty cool.

rodrigonogueira4:
+1. As @lintool mentioned, now seems a good time because there is no deadline in the near future.

zeynepakkalyoncu:
+1. I agree it's a good time to introduce a considerable change so that we can work on any issues until the next deadline.

lintool:
Update: I now have all our existing regressions passing on Lucene8, in a branch called `lucene8`:

```
[...]

PR #2226, opened 2023-10-23
Upgrade to Lucene 9.8.0

tteofili:
Upgrade Anserini to use Apache Lucene 9.8.0.


tteofili:
closes #679

src/main/java/io/anserini/ltr/DocumentFieldContext.java
```diff
@@ -23,11 +23,7 @@
 import org.apache.lucene.index.Term;
 import org.apache.lucene.index.Terms;
 import org.apache.lucene.index.TermsEnum;
-import org.apache.lucene.search.DocIdSetIterator;
-import org.apache.lucene.search.DocValuesFieldExistsQuery;
-import org.apache.lucene.search.IndexSearcher;
-import org.apache.lucene.search.Query;
-import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.*;
```
lintool, line 26: Let's change back to explicit imports? I like that better stylistically...

lintool (approved):

lintool:
Let's go ahead and merge! (And I'll run all our regressions from scratch anyway...)

src/main/java/io/anserini/util/ExtractDocumentLengths.java
```diff
@@ -78,7 +78,7 @@ public static void main(String[] args) throws Exception {
 
     out.println("internal_docid\texternal_docid\tdoc_length\tunique_term_count\tlossy_doc_length\tlossy_unique_term_count");
     for (int i = 0; i < numDocs; i++) {
-      Terms terms = reader.getTermVector(i, Constants.CONTENTS);
+      Terms terms = reader.termVectors().get(i, Constants.CONTENTS);
```
jpountz, line 81: Ideally you'd fetch the `reader.termVectors()` object only once, before the loop, and then reuse it in the loop. (And similarly to other places that rely on stored fields and term vectors.)

tteofili: yeah, good point!

tteofili: Will have another PR fix that in different places (and perhaps  storedFields() too)

PR #2231, opened 2023-10-31
Tweaked cosDPR-distil scores

lintool:
There's some variance with cosDPR-distil scores... last time, I ran the 2CRs once and then captured the scores - this is a bit brittle and results in failures sometimes. This PR captures a better approach: I ran the tests 5 times and averaged the scores.


MXueguang (approved):
lgtm

PR #2238, opened 2023-11-04
Add simple script to analyze reproductions

lintool:
Simple script to analyze reproduction logs:

```
% python src/main/python/analyze_reproductions.py
Analyzing docs/start-here.md
 + count: 19
 + start: 2023-07-21
 + end:   2023-11-01
 + total: 103 days
 + ratel: 5.42

Analyzing docs/experiments-msmarco-passage.md
 + count: 108
 + start: 2019-08-12
 + end:   2023-11-03
 + total: 1544 days
 + ratel: 14.30
```


ronakice:
Code looks good and nice to see such stats but what is the meaning of ratel as defined by ‘date diff/cnt’

ronakice (approved):

lintool:
> Code looks good and nice to see such stats but what is the meaning of ratel as defined by ‘date diff/cnt’

Did you think I should flip numerator and denominator and compute "reproductions per day"?

Right now it's actually computing days between reproductions (on avg).

"reproductions per day" will be a value that's pretty small... much less than 1.

Or just keep the current metrics but clarify?

ronakice:
Yeah, maybe this is fine too, just need some clarity then since rate of reproductions I could calculate as the flipped! I think more interesting would be to see seasonal patterns, and look at this metric over time. Like confirm that there is a small bump after around paper announcements and then every now and then when URA cycle begins or something. Else we lump it all into a single number, but this is probably a task for some URA ;)

lintool:
Minor tweaks:

```
% python src/main/python/analyze_reproductions.py
Analyzing docs/start-here.md
 + count: 19
 + start: 2023-07-21
 + end:   2023-11-01
 + total: 103 days
 + freq:  5.42 days between reproductions (on avg)

Analyzing docs/experiments-msmarco-passage.md
 + count: 108
 + start: 2019-08-12
 + end:   2023-11-03
 + total: 1544 days
 + freq:  14.30 days between reproductions (on avg)
```

PR #2253, opened 2023-11-09
Fix splade-pp-sd-onnx regression tests

ArthurChen189:
Just spotted one inconsistent regression test. Scores are updated using the results of a new run on ORCA.


lintool (approved):
good catch!

PR #2259, opened 2023-11-12
Add entry to reproduction log

sjjohnst:
Ran on Compute Canada HPC, Cedar. Linux OS, used a python3.10 virtual environment, single node, 32G memory. Everything worked as expected.


lintool:

PR #2261, opened 2023-11-14
Follow up: add bindings for CIRAL's dev qrels

Mofetoluwa:
Follow up to [castorini/anserini/pull/2258](https://github.com/castorini/anserini/pull/2258) as branch was deleted. 
Including bindings for dev qrels.


lintool (changes requested):
Please add tests to `RelevanceJudgmentsTest`.

PR #2262, opened 2023-11-14
Loosen regression tests tolerance for ONNX runs

lintool:
For ONNX runs, increase tolerance a bit because we observe some minor differences across OSes.

### DIFF
--- a/src/main/python/run_regression.py
+++ b/src/main/python/run_regression.py
@@ -226,9 +226,9 @@ def evaluate_and_verify(yaml_data, dry_run):
 
     if not dry_run:
         if failures:
-            logger.info(f'\033[91mFailed tests!\033[0m Total elapsed time: {end - start:.0f}s')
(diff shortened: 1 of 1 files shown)


src/main/python/run_regression.py
```diff
@@ -222,6 +227,8 @@ def evaluate_and_verify(yaml_data, dry_run):
     if not dry_run:
         if failures:
             logger.info(f'\033[91mFailed tests!\033[0m Total elapsed time: {end - start:.0f}s')
+        elif okish:
+            logger.info(f'\033[94m[OK*]\033[0m Total elapsed time: {end - start:.0f}s')
```
cadurosar, line 231: Nitpick, but why not use the okish_str here instead of the fixed string?

cadurosar (approved):
LGTM, only thing is that small nitpick at the end that could reuse the string, but also it was not a part of the file before (failures should also use) so can be skipped

PR #2275, opened 2023-11-26
Extract common code paths in indexing pipeline

lintool:
Major refactoring of indexing pipeline (IndexCollection, IndexHnswDenseVectors, and IndexInvertedDenseVectors), extracting common code paths into AbstractIndexer.

@tteofili @MXueguang still running tests, but ready for review in parallel.


tteofili (approved):
thx @lintool LGTM

src/main/java/io/anserini/index/AbstractIndexer.java
```diff
@@ -0,0 +1,338 @@
...
+    public boolean uniqueDocid = false;
+
+    @Option(name = "-optimize", usage = "Optimizes index by merging into a single index segment.")
+    public boolean optimize = false;
+
+    @Option(name = "-memoryBuffer", metaVar = "[mb]", usage = "Memory buffer size in MB.")
+    public int memoryBuffer = 16384;
+
+    @Option(name = "-threads", metaVar = "[num]", usage = "Number of indexing threads.")
+    public int threads = 4;
```
jpountz, line 71: Thinking out loud: what about defaulting to `Runtime.getRuntime().availableProcessors()` so that indexing concurrency would automatically try to use all available resources, without requiring users to pass the right number of threads? (Possibly for another PR)

lintool:
Tests are proceeding, but not finished yet. Stand by.

PR #2278, opened 2023-11-27
Fix typo for number of total queries

AndreSlavescu:
# Changes
- fixed small typo in second onboarding task doc


lintool:
Good catch. Can you please consolidate into #2277 ?

PR #2300, opened 2023-12-11
Refactoring search pipeline for HNSW and inverted dense

lintool:
+ Refactored to create HnswDenseSearcher and InvertedDenseSearcher - these will provide bindings from Python
+ SearchHnswDenseVectors and SearchInvertedDenseVectors wrapper the two searchers above to provide main
+ Improved test coverage


lintool:
@tteofili please review when you have a chance - I'll run tests in parallel.

src/main/java/io/anserini/search/ScoredDoc.java
```diff
@@ -0,0 +1,33 @@
...
+
+package io.anserini.search;
+
+import org.apache.lucene.document.Document;
+
+public class ScoredDoc {
+  public String docid;
+  public int lucene_docid;
+  public float score;
+  public Document lucene_document;
```
tteofili, line 25: trivial comment: it'd be nice to keep Java variable naming conventions (e.g., `docId` instead of `docid`, `luceneDocId` instead of `lucene_docid`, `luceneDocument` instead of `lucene_document`).

tteofili (approved):
LGTM (with just a trivial comment)

src/main/java/io/anserini/search/ScoredDoc.java
lintool: Ah, I've gone back and forth on this - this class is meant for Pyserini Python bindings... so the options are to use Python conventions in Java or use Java conventions from Python... I opted for the first.

tteofili: ah yes, that's good then !

PR #2331, opened 2024-01-10
Update linux + WSL trec_eval binaries

jasper-xian:
Adds WSL binary

Updates linux-amd binary. It was previously built on linux.student but didn't work properly in orca, now built in orca and works on both ends


jasper-xian:
@Melissa1412 would you be able to test the WSL binary you built? You can do this by building anserini in WSL and running any trec_eval command like so
```
target/appassembler/bin/trec_eval <qrels> <runfile>
```

lintool (approved):

Melissa1412:
> @Melissa1412 would you be able to test the WSL binary you built? You can do this by building anserini in WSL and running any trec_eval command like so
> 
> ```
> target/appassembler/bin/trec_eval <qrels> <runfile>
> ```

Tested and worked
Screenshot 2024-01-10 184112

PR #2334, opened 2024-01-12
Rollback trec_eval to 9.0.4

jasper-xian:


lintool:
@jasper-xian rollback docs as well?

```
% grep '9.0.8' docs/*.md
```

jasper-xian:
Nothing to rollback in docs I think, I previously only updated to use `target/appassembler/bin/trec_eval` which stays consistent with this @lintool

lintool:
> Nothing to rollback in docs I think, I previously only updated to use `target/appassembler/bin/trec_eval` which stays consistent with this @lintool

My mistake, yes, you are correct.

lintool (approved):

PR #2336, opened 2024-01-13
Add to Reproduction Logs

Tanngent:
Reproduced on Apple M2
8 GB RAM

macOS Sonoma 14.2.1
Python 3.9.6
maven 3.9.6
Java 21.0.1


lintool:
Please continue along the onboarding path until you get to the end.

PR #2348, opened 2024-01-20
Update start-here.md

AmaanAmythAlegend:
Git Bash on Windows used.
Everything worked.
One possible clarification is that wget may not be already available. It is to be installed by pip and wget binaries downloaded for Windows so that it works in Git Bash.


lintool:
Please continue along the onboarding path, and we'll take a look at the end.

lintool:
Having seen no follow up, closing.

PR #2353, opened 2024-01-24
Add BGE regressions for BEIR

lintool:


lintool:
@MXueguang this is ready for review.

PR #2357, opened 2024-01-26
Cohere MS MARCO v1 passage 2CR

jasper-xian:
Ref https://github.com/castorini/anserini-tools/pull/67


lintool (approved):
Please update the `tools/` submodule to pull in the topics also (ref https://github.com/castorini/anserini-tools/pull/67), and then go ahead and merge.

PR #2366, opened 2024-02-10
Refactor MS MARCO passage dev with Cohere V3 embeddings

lintool:
Added int8 indexes, general clean-up, added links to README.

@cadurosar we're missing dl19/dl20 - what do you want to do? Send in separate PR?


lintool:
@cadurosar says we should go ahead and merge this, and circle back to add dl19/dl20 later.

@jasper-xian will wait for your CR first.

jasper-xian (approved):

Issue #2374, opened 2024-02-15
bge-base-en-v1.5 encoder query length issues


lintool:
For arguana, we're going to run into the same query length issues as #2368 for bge-base-en-v1.5

@ArthurChen189 can you take a look please?

lintool:
Closed by #2387

PR #2375, opened 2024-02-15
Add BGE regressions with ONNX on BEIR

lintool:
We're blocked on #2374 - otherwise arguana won't work.


lintool:
with @ArthurChen189 fixing the maxlen issue with ONNX in #2387 this can be merged in now.

lintool:
Note: I've simply copied the scores from the non-ONNX versions - will circle back and re-adjust scores.

MXueguang (approved):

PR #2378, opened 2024-02-17
Update start-here.md

saeedt70:
google colab
python 3
t4 gpu
Due to implementation and cloning in colab, I had to change the special addresses of file in the colab


lintool:
Please continue to the end of the onboarding path and we'll take a look all at once.

lintool:

PR #2385, opened 2024-02-20
Cohere embed-english-v3 DL19/20 2CR

jasper-xian:
WIP (will re-verify when topics are merged to tools)


lintool (approved):
Add links in README also?

Merge when ready.

PR #2399, opened 2024-02-27
Update README to add "try it" section: reproduce directly from fatjar

lintool:


cadurosar (approved):
This is so cool, being able to use the fatjar directly makes things way easier!

PR #2410, opened 2024-03-17
Upgrade to JDK 21

lintool:


pom.xml
```diff
@@ -224,7 +236,7 @@
               </transformers>
               <!-- This fixes the issue "Invalid signature file digest for Manifest main attributes"
                    cf. http://zhentao-li.blogspot.com/2012/06/maven-shade-plugin-invalid-signature.html -->
-              <filters>
+              <!--filters>
```
sahel-sh, line 239: is this change intentional?

sahel-sh (approved):
LGTM with a question

pom.xml
lintool: Yes, commented out intentionally - not sure its effect... so punting for now, will circle back later to figure out implications.

MXueguang (commented):
lgtm

PR #2415, opened 2024-03-23
Tweak 'name' in regression yaml files to make more unique

lintool:
This avoids output runfile collisions, enabling regressions to run concurrently without clobbering each other.


MXueguang (approved):
lgtm

PR #2425, opened 2024-03-28
Frontend

16BitNarwhal:
Demo for anserini made with React.js and Springboot

image

Run with `mvn spring-boot:run` then open http://localhost:8080/


dependency-reduced-pom.xml
```diff
@@ -0,0 +1,705 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lintool, line 1: this file shouldn't be checked in..

src/main/frontend/.gitignore
```diff
@@ -0,0 +1,36 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
lintool, line 1: this should be merged with the `.gitignore` in the root of the repo.

lintool (changes requested):
repo style is two-space indents.
please change all?

src/main/java/io/anserini/demo/QueryResult.java
```diff
@@ -0,0 +1,39 @@
+package io.anserini.demo;
+
+public class QueryResult {
+    private String docId;
```
lintool, line 4: `docId` -> `docid`

lintool:
@16BitNarwhal Can you merge in current `master` and resolve conflicts (if any)?

Also, let's rename the main entry point from `io.anserini.server.Application`?

lintool:
Also, I'm getting:

```
[ERROR] Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:2.7.5:run (default-cli) on project anserini: Execution default-cli of goal org.springframework.boot:spring-boot-maven-plugin:2.7.5:run failed: Unsupported class file major version 65 -> [Help 1]
```

65 = JDK 21... so dependencies need to be upgraded somewherE?

lintool:
Your commit of `tools/` appears to be behind `master`? Doesn't pick up new commits... can you please update?

Another issue - this PR increase the size of the fatjar from 166M to 184M, from the new jars that are pulled in. If they're absolutely necessary, then it's okay... but can you do a pass to see if you can reduce bloat? i.e., remove all the unnecessary optional dependencies.

lintool (approved):

PR #2434, opened 2024-03-31
Completed Onboarding Path 

G-Wang12:
Completed Start Here module on MacOS. Every step was done smoothly.


lintool:

lintool:
Having seen no further follow-up, closing.

PR #2457, opened 2024-04-16
Refactor regressions with prebuilt indexes

lintool:
+ Upgraded to renamed indexes conforming to new schema.
+ Created docs/fatjar-regressions-v0.35.0.md - old instructions moved to here
+ Created docs/fatjar-regressions-v0.35.1-SNAPSHOT.md - new instructions here

Read for review.


docs/fatjar-regressions-v0.35.1-SNAPSHOT.md
```diff
@@ -0,0 +1,251 @@
+# Anserini Fatjar Regresions (v0.35.1-SNAPSHOT)
+
+## MS MARCO V1 Passage
+
+Currently, Anserini provides support for the following models:
+
++ BM25
```
sahel-sh, line 7: IIRC, there are many more choices in Pyserini 2cr. Are they Pyserini only? or there is no regression test for them?

src/test/java/io/anserini/doc/GenerateReproductionDocsTest.java
```diff
@@ -41,12 +41,16 @@ public class GenerateReproductionDocsTest {
...
+      "cos-dpr-distil.fp32.onnx",
+      "cos-dpr-distil.int8.cached_q",
+      "cos-dpr-distil.int8.onnx",
+      "bge-base-en-15.fp32.cached_q",
+      "bge-base-en-15.fp32.onnx",
+      "bge-base-en-15.int8.cached_q",
+      "bge-base-en-15.int8.onnx"
   };
 
   public static String findMsMarcoTableTopicSetKeyV1(String topicKey) {
```
sahel-sh, line 56: looks like some of the newly added topic keys have trec2020 or treck2019 instead of dl20 or dl19, should they be covered in the tests too?

docs/fatjar-regressions-v0.35.1-SNAPSHOT.md
lintool: Yes, Pyserini has more conditions - many of which are not available from Anserini. I haven't though through relationship moving forward... i.e., are they going to have the same (unlikely); is Pyserini going to be a superset (more likely). My current thinking is that Anserini is going to have regressions that are "easily doable" from Lucene.

src/test/java/io/anserini/doc/GenerateReproductionDocsTest.java
lintool: Actually, I already have test cases for aliases: https://github.com/castorini/anserini/blob/master/src/test/java/io/anserini/search/topicreader/TopicsTest.java#L40

Will add more.

PR #2459, opened 2024-04-20
Add bindings for MS MARCO V2.1 prebuilt indexes + qrels

lintool:
Allows us to directly perform MS MARCO V2.1 runs.


ronakice (approved):
LGTM!

PR #2463, opened 2024-04-24
Add `-outputRerankerRequests` option to create input for RankLLM 

ronakice:
As [discussed in RankLLM](https://github.com/castorini/rank_llm/issues/110#issuecomment-2070836397), we can create an intermediate JSONL file with this option.

Example Usage:
`java -cp `ls target/*-fatjar.jar` io.anserini.search.SearchCollection -index msmarco-v2.1-doc-segmented -topics trec2021-dl -output runs/run.msmarco-v2.1-doc-segmented.dl21.txt -outputRerankerRequests runs/retrieve_results_msmarco-v2.1-doc-segmented-dl21_top20.jsonl -bm25 -hits 20 -threads 16 -format trec`

We only use JSONL, since RankLLM supports it. 

The results for a single query can then be visualized by `head -1 runs/retrieve_results_msmarco-v2.1-doc-segmented-dl21_top20.jsonl | jq`.


dependency-reduced-pom.xml
```diff
@@ -0,0 +1,563 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lintool: This file shouldn't be checked in.

src/main/java/io/anserini/search/RunOutputWriter.java
```diff
@@ -16,48 +16,79 @@
...
+      Map<String, Object> queryMap = new LinkedHashMap<>();
+      queryMap.put("query", new LinkedHashMap<>(Map.of("qid", qid, "text", query)));
+      queryMap.put("candidates", candidates);
+      outputRerankerRequests.println(mapper.writeValueAsString(queryMap));
+    }
+    
+    if ("msmarco".equals(format)) {
+      for (ScoredDoc r : results) {
         out.append(String.format(Locale.US, "%s\t%s\t%d\n", qid, r.docid, rank));
-      } else {
```
lintool, line 44: Please keep this comment.

dependency-reduced-pom.xml
ronakice: yep, noticed it just before this 👍🏼

src/main/java/io/anserini/search/RunOutputWriter.java
lintool: i.e., outlining TREC format.

src/main/java/io/anserini/search/SearchHnswDenseVectors.java
```diff
@@ -16,6 +16,8 @@
```
lintool, line 16: Can we remove this option in this class for now? We don't have a way to test it, and so I would rather _not_ have a feature than an _untested_ feature.

src/main/java/io/anserini/search/SearchInvertedDenseVectors.java
```diff
@@ -16,6 +16,8 @@
```
lintool, line 16: Same as above. Remove?

Issue #2427, opened 2024-03-28
Can't run 2CR on pre-built indexes directly on fatjar - can't read YAML files


lintool:
We need to read the YAML files from the fatjar, not depend on a local file:

```
$ wget https://repo1.maven.org/maven2/io/anserini/anserini/0.25.0/anserini-0.25.0-fatjar.jar

$ java -cp anserini-0.25.0-fatjar.jar io.anserini.reproduce.RunMsMarco
Exception in thread "main" java.io.FileNotFoundException: src/main/java/io/anserini/reproduce/msmarco-v1-passage.yaml (No such file or directory)
	at java.base/java.io.FileInputStream.open0(Native Method)
	at java.base/java.io.FileInputStream.open(FileInputStream.java:219)
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:112)
	at io.anserini.reproduce.RunMsMarco.main(RunMsMarco.java:41)
```

16BitNarwhal:
working on this

ArthurChen189:
Sorry, I was prepping for final exams and working on hackathon stuff. I am happy to help @16BitNarwhal if there is any question :)

lintool:
@16BitNarwhal any progress on this?

16BitNarwhal:
I was able get RunMsMarco to read the YAML file from fatjar but the fatjar is still not completely self-contained as it can only be run from the root directory (it needs access to `bin/run.sh` which (I think) 
is running all fatjars from `target/`

I'm having a bit of trouble proceeding from here

lintool:
I think you need to read directly from a jar, something like this? https://stackoverflow.com/questions/20389255/reading-a-resource-file-from-within-jar

lintool:
#2469 was a first step, but we have another issue:

```
% java -cp `ls *-fatjar.jar` io.anserini.reproduce.RunMsMarco
# Running condition "bm25-default": BM25 (k1=0.9, b=0.4) 

  - topic_key: msmarco-v1-passage.dev

    Running retrieval command: bin/run.sh io.anserini.search.SearchCollection -threads 16 -index msmarco-v1-passage -topics msmarco-v1-passage.dev -output runs/run.msmarco-v1-passage.bm25-default.msmarco-v1-passage.dev.txt -hits 1000 -bm25
```

[...]

PR #2469, opened 2024-04-27
msmarco passage yaml read from fatjar resources

16BitNarwhal:
Resolves issue #2427
Added `msmarco-v1-passage.yaml` to resources directory
Reads the yaml resource within the fatjar when running `io.anserini.reproduce.RunMsMarco`

### DIFF
--- a/src/main/java/io/anserini/reproduce/RunMsMarco.java
+++ b/src/main/java/io/anserini/reproduce/RunMsMarco.java
@@ -35,10 +35,11 @@ public class RunMsMarco {
   private static final String FAIL = RED + "[FAIL]" + RESET;
 
   private static final String COLLECTION = "msmarco-v1-passage";
-
+ 
   public static void main(String[] args) throws Exception {
     final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
-    Config config = mapper.readValue(new FileInputStream("src/main/java/io/anserini/reproduce/msmarco-v1-passage.yaml"), Config.class);
+    InputStream inputStream = RunMsMarco.class.getClassLoader().getResourceAsStream("msmarco_v1/msmarco-v1-passage.yaml");
+    Config config = mapper.readValue(inputStream, Config.class);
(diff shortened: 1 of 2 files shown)


src/main/java/io/anserini/reproduce/RunMsMarco.java
```diff
@@ -35,10 +35,11 @@ public class RunMsMarco {
   private static final String FAIL = RED + "[FAIL]" + RESET;
 
   private static final String COLLECTION = "msmarco-v1-passage";
-
+ 
   public static void main(String[] args) throws Exception {
     final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
-    Config config = mapper.readValue(new FileInputStream("src/main/java/io/anserini/reproduce/msmarco-v1-passage.yaml"), Config.class);
+    InputStream inputStream = RunMsMarco.class.getClassLoader().getResourceAsStream("msmarco_v1/msmarco-v1-passage.yaml");
```
lintool, line 41: hygiene: wrap in a `try(...)` to make sure `inputStream` gets closed?

src/main/resources/msmarco_v1/msmarco-v1-passage.yaml
```diff
@@ -0,0 +1,300 @@
+conditions:
```
lintool, line 1: We already have `src/main/java/io/anserini/reproduce/msmarco-v1-passage.yaml` right? Can't we just use that?

lintool (changes requested):

src/main/resources/msmarco_v1/msmarco-v1-passage.yaml
16BitNarwhal: I believe all the files in `src/main/resources/` can be read within a fatjar. To read files from `src/main/java/io/anserini/reproduce/msmarco-v1-passage.yaml` we would need to specify that directory as a resource in `pom.xml`.

lintool (approved):

PR #2470, opened 2024-04-27
Adding support for trec_eval to take symbols representing standard qrels

### LINKED ISSUE #2391: Allow trec_eval to take symbols representing standard qrels (instead of full qrel files)
Currently, for `trec_eval`, we have to do something like:

```
target/appassembler/bin/trec_eval -c -M 10 -m recip_rank tools/topics-and-qrels/qrels.msmarco-passage.dev-subset.txt runs/run.msmarco-passage.bm25.txt
```

It would be great if we could do something like:

```
target/appassembler/bin/trec_eval -c -M 10 -m recip_rank msmarco-passage runs/run.msmarco-passage.bm25.txt
```

That is, take a symbol instead of full file.

Simple algorithm: take symbol, find the prefix that matches, e.g., so user can specify `msmarco-passage` instead of `msmarco-passage.dev-subset`; behind the scenes, we download to `~/.cache`.

Something like this already works with Pyserini, we need to backport to Pyserini.

DanielKohn1208:
Closes #2391. The behavior of `RelevanceJudgments.getQrelsPath(Path qrelPath)` is also slightly modified to allow for qrelPath to be a symbol.


src/main/java/trec_eval.java
```diff
@@ -140,12 +142,40 @@ public trec_eval() {
     //System.err.println(ourTE.toString());
```
lintool, line 142: remove dead code?

src/main/java/io/anserini/eval/Qrels.java
```diff
@@ -204,4 +204,25 @@ public static boolean contains(Path topicPath) {
     }
     return false;
   }
+
+  public static boolean containsSymbol(Path topicPath) {
+    for (Qrels c: Qrels.values()) {
```
lintool, line 209: looping seems inefficient?

Why not do it once (on init), and build a dictionary (e.g., Map) to support quick lookup?

lintool:
@DanielKohn1208 thanks for your contribution.

Test cases please ;)

DanielKohn1208:
> @DanielKohn1208 thanks for your contribution.
> 
> Test cases please ;)

I have found the test file for `RelevanceJudgments`. Do you know where the test file for `trec_eval.java` can be found?

lintool:
> I have found the test file for `RelevanceJudgments`. Do you know where the test file for `trec_eval.java` can be found?

`trec_eval` has no test case currently... it's e2e so a bit more difficult to test. (But feel free to create new test!)

However, you can test symbol bindings in isolation in `RelevanceJudgmentsTest`. (The tests are extensive there - probably don't need to test _everything_, but at least spot checking the symbol bindings?)

lintool (approved):

PR #2476, opened 2024-04-29
Locate and run fatjar in RunMSMarco

16BitNarwhal:
for issue #2427 

For `RunMsMarco.java`, rather than depending on `bin/run.sh` and `bin/trec_eval`, directly locate and run fatjar with Java

Issue #2473, opened 2024-04-28
Align RunMsMarco with Fatjar regression instructions


lintool:
Today, you can run regressions with the fatjar directly:
https://github.com/castorini/anserini/blob/master/docs/fatjar-regressions-v0.35.1.md

Today, you can also do this:
```
java -cp `ls target/*-fatjar.jar` io.anserini.reproduce.RunMsMarco
```

These aren't aligned. `RunMsMarco` only runs a subset of the cases in the fatjar instructions. We should bring the two into alignment - i.e., add the missing cases into `RunMsMarco`. It probably makes sense to run have then generate the same output to be consistent.

lintool:
Separately:

+ we have fatjar instructions for BEIR, but no `io.anserini.reproduce.RunBeir` or something similar.
+ we have no repro for V2.1 runs, e.g., add option to `io.anserini.reproduce.RunMsMarco`.

lintool:
Initial commit #2477 - some errors noted #2480

lintool:
I believe this issue has been resolved.

PR #2477, opened 2024-04-29
Added RunMsMarco V2.1, Added RunBeir

wu-ming233:
- Added command-line option for RunMsMarco repro to support V2.1 and V2.1 segmented indexes.
- Added a new repro for RunBeir with output format following the instructions in [fatjar-regressions-v0.35.1.md](https://github.com/castorini/anserini/blob/master/docs/fatjar-regressions-v0.35.1.md).
- Removed some minor dead code.


wu-ming233:
Refs #2473

src/main/java/io/anserini/reproduce/RunMsMarco.java
```diff
@@ -39,66 +33,81 @@ public class RunMsMarco {
...
+
   public static void main(String[] args) throws Exception {
-    final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
-    try (InputStream inputStream = RunMsMarco.class.getClassLoader()
-        .getResourceAsStream("reproduce/msmarco-v1-passage.yaml")) {
-      if (inputStream == null) {
-        throw new FileNotFoundException("Resource file not found. Please check the file path.");
+
+    // check for cmd option
+    if (args.length >= 2 && "-v".equals(args[0])) {
```
lintool, line 41: Can you parse args in a manner that's consistent with everything else in repo? E.g., https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/index/IndexCollection.java#L272

src/main/java/io/anserini/reproduce/RunMsMarco.java
```diff
@@ -39,66 +33,81 @@ public class RunMsMarco {
...
+      String collectionVersion = args[1];
+      switch (collectionVersion) {
+          case "1":
+              COLLECTION = "msmarco-v1-passage";
+              break;
+          case "2":
+              COLLECTION = "msmarco-v2.1-doc";
```
lintool, line 48: Ultimately, we want something like https://castorini.github.io/pyserini/2cr/msmarco-v2-doc.html

So it's okay with put doc and doc segmented together.

src/main/resources/reproduce/msmarco-v2.1-doc.yaml
```diff
@@ -0,0 +1,39 @@
+conditions:
```
lintool, line 1: Combine doc with doc-segmented?

lintool (changes requested):
Lots of code duplication... is there any low-hanging fruit in terms of abstracting out common code?

src/main/resources/reproduce/msmarco-v2.1-doc.yaml
wu-ming233: While coding I was thinking how RunBeir could just simply be another command line option in RunMsMarco, due to the uniform interface in the yaml files. Would it make sense if we combine them into one script, called "RunRepro" or something similar, and specify the different regressions in yaml files?

lintool: Currently, in Pyserini, each corpus is a separate file: https://github.com/castorini/pyserini/tree/master/pyserini/2cr

We're modelling after that repo as a start, but can probably further combine later?

For now, let's have MS MARCO and BEIR as separate classes, and then maybe circle back?

wu-ming233:
Refactored RunMsMarco and RunBeir; put common code in a new "RunRepro".
Note that src/test/java/io/anserini/doc/GenerateReproductionDocsTest.java was modified to use an updated variable name, though the semantics remained.

lintool:
Hi @wu-ming233 can you resolve conflict and merge in #2476 ? Thanks!

wu-ming233:
Merged with #2476 and updated the corresponding yaml's for other regressions as well, adhering to #2476.

lintool (approved):

Issue #2480, opened 2024-04-30
Errors with new MS MARCO v2.1 and BEIR regressions


lintool:
Running:

```
java -cp `ls target/*-fatjar.jar` io.anserini.reproduce.RunMsMarco -v 2
```

Getting some errors:

```
# Running condition "bm25-segmented": BM25 v2.1 Segmented Corpus (k1=0.9, b=0.4) 

  - topic_key: msmarco-v2-doc-dev

    Running retrieval command: java -cp /Users/jimmylin/workspace/anserini/target/anserini-0.36.1-SNAPSHOT-fatjar.jar io.anserini.search.SearchCollection -threads 16 -index msmarco-v2.1-doc-segmented -topics msmarco-v2-doc-dev -output runs/run.msmarco-v2.1-doc.bm25-segmented.msmarco-v2-doc-dev.txt -hits 1000 -bm25 -selectMaxPassage -selectMaxPassage.delimiter "#" -selectMaxPassage.hits 1000
    Run successfully completed!

     MRR@10: 0.0000 [FAIL] expected 0.1973

  - topic_key: msmarco-v2-doc-dev2

    Running retrieval command: java -cp /Users/jimmylin/workspace/anserini/target/anserini-0.36.1-SNAPSHOT-fatjar.jar io.anserini.search.SearchCollection -threads 16 -index msmarco-v2.1-doc-segmented -topics msmarco-v2-doc-dev2 -output runs/run.msmarco-v2.1-doc.bm25-segmented.msmarco-v2-doc-dev2.txt -hits 1000 -bm25 -selectMaxPassage -selectMaxPassage.delimiter "#" -selectMaxPassage.hits 1000
    Run successfully completed!

     MRR@10: 0.0000 [FAIL] expected 0.2000

  - topic_key: trec2021-dl

    Running retrieval command: java -cp /Users/jimmylin/workspace/anserini/target/anserini-0.36.1-SNAPSHOT-fatjar.jar io.anserini.search.SearchCollection -threads 16 -index msmarco-v2.1-doc-segmented -topics trec2021-dl -output runs/run.msmarco-v2.1-doc.bm25-segmented.trec2021-dl.txt -hits 1000 -bm25 -selectMaxPassage -selectMaxPassage.delimiter "#" -selectMaxPassage.hits 1000
    Run successfully completed!

        MAP: 0.0000 [FAIL] expected 0.2609

     MRR@10: 0.0000 [FAIL] expected 0.9026

[...]

PR #2483, opened 2024-05-02
Fix MS MARCO V2.1 repo experiments on segmented doc collection

lintool:
Proposed fix to https://github.com/castorini/anserini/issues/2480#issuecomment-2085063194

The issue is that '#' may get interpreted as a comment; WSL and Mac might also different on how interpolation, escaping, etc. is handled with `ProcessBuilder`. This works on my Mac.


MXueguang (approved):

lintool:
This works on both Linux and Mac, so merging in. We can get someone to check Windows via WSL later.

PR #2491, opened 2024-05-08

billycz8:
System setup:

OS: macOS Ventura 13.4
Memory: 16GB
Chip: Apple M1 Pro
Python Version: 3.7
Java Version: 22.0.1
Maven: 3.9.6

Everything worked perfectly. Suggestion: I think that adding a short addition to the description of MRR (only the rank of the "first relevant document" in the ranked list is used to compute a query score) could be helpful.


lintool:
> Everything worked perfectly. Suggestion: I think that adding a short addition to the description of MRR (only the rank of the "first relevant document" in the ranked list is used to compute a query score) could be helpful.

Hi @billycz8 - can you propose a wording change directly in this PR? Let's discuss...

lintool (approved):

billycz8:
Hi @lintool, I have added a possible rewording for the description of the MRR (we could still change it). I think adding it could be helpful as I first found it a little ambiguous when I did the calculation by hand (in the exercise later in the doc). It is just a suggestion

lintool:
lg, thanks!

Issue #2502, opened 2024-05-24
Discussion: REST API routes for different corpus/model combinations - how do we name?


lintool:
We have different corpus/model combinations. For example, we have `msmarco-v1-passage`, `msmarco-v1-passage.splade-pp-ed`, and `msmarco-v1-passage.cos-dpr-distil`. They are all on MS MARCO V1 Passage, but different models - the first for BM25, SPLADE, and cosDPR.

For querying different corpus/model combinations, how do we want to configure the REST API route?

Options I see are:

(1) corpus in route, model as query parameter, e.g.,

+ `/api/collection/msmarco-v1-passage/search?query={query}&model=bm25`
+ `/api/collection/msmarco-v1-passage/search?query={query}&model=splade-pp-ed`
+ `/api/collection/msmarco-v1-passage/search?query={query}&model=cos-dpr-distil`

(2) corpus/model (=index) combination directly in route, e.g.,

+ `/api/collection/msmarco-v1-passage/search?query={query}&model=bm25`
+ `/api/collection/msmarco-v1-passage.splade-pp-ed/search?query={query}`
+ `/api/collection/msmarco-v1-passage.cos-dpr-distil/search?query={query}`

The first feels conceptually cleaner, but requires an extra layer of indirection to map into our final indexes.

If we go with option (2), then "collection" is misnamed, but we can change to "index"?

16BitNarwhal:
I initially named the route to match the PrebuiltIndexHandler since all the indexes were from there so in that case option (2) changed to "index" would be an improvement (I'm not sure why I named 
it "collection" at the start). I also agree that (1) is cleaner, but I'm not sure if it works for BEIR (and correct me if I'm wrong) since I don't think .flat is a model

ronakice:
Using the first option, you can maintain clarity and separation between the corpus and the model, making it easier to add or change models in the future. It also simplifies API usage, as users will understand that the primary resource is the collection, and the model is a configurable parameter. So I side with it, but agreed too that it is a bit of work.

lintool:
> I also agree that (1) is cleaner, but I'm not sure if it works for BEIR (and correct me if I'm wrong) since I don't think .flat is a model

Yea, you'll need another layer of indirection.

I'll make an executive decision of going with (2) for now, with perhaps moving to (1) further down the road... it's probably too much work right now.

@16BitNarwhal can you rename `/api/collection/` to `/api/index/` and let's wrap this up for now?

[...]

PR #2525, opened 2024-06-16
Webapp update

16BitNarwhal:
following discussion of #2502 
- exposed two new routes, `/indexes` and `/indexes/{index}/status` from the server
- webapp dropdown list is generated from `/indexes` route
- dropdowns use index corpus + method instead of indexname field now
- dropdowns have indicator for whether they're already cached
- `/index/{index}/` is renamed to `/indexes/{index}`

image


src/main/java/io/anserini/server/Controller.java
```diff
@@ -52,4 +59,34 @@ public Map<String, Object> search(@PathVariable(value = "index", required = fals
     return queryMap;
   }
 
+  @RequestMapping(method = RequestMethod.GET, path = "/isCached/{index}")
```
lintool, line 62: See https://medium.com/@nadinCodeHat/rest-api-naming-conventions-and-best-practices-1c4e781eb6a5

What's the best practice for a "status endpoint"?

src/main/java/io/anserini/server/Controller.java
```diff
@@ -52,4 +59,34 @@ public Map<String, Object> search(@PathVariable(value = "index", required = fals
...
+    if (!IndexInfo.contains(index)) {
+      throw new IllegalArgumentException("Index name " + index + " not found!");
+    }
+    
+    PrebuiltIndexHandler handler = new PrebuiltIndexHandler(index);
+    handler.initialize();
+    return handler.checkIndexFileExist();
+  }
+
+  @RequestMapping(method = RequestMethod.GET, path = "/list")
```
lintool, line 73: Same comment... what are some alternative route names that might better conform to best practices?

lintool:
Also, can you attach here a screenshot so I have an idea of what it looks like now?

16BitNarwhal:
> Additional thoughts about API design: maybe the route should be `catalog`? and asking whether or not it's downloaded should be a param?

Whoops, saw this a bit late. I was thinking `/index/{index}/cached` to check status and `/indexes` to list all

lintool:
I think this is a good time to document the API in an MD, e.g., `docs/rest-api.md` or something like that?

Provide examples like this? https://github.com/castorini/anserini/blob/master/docs/fatjar-regressions/fatjar-regressions-v0.36.1.md#webapp-and-rest-api

lintool (approved):

PR #2535, opened 2024-06-20
Tweak regression scores due to DJL upgrade

lintool:
Recent upgrade to DJL v0.28.0 #2529 caused a bunch of score differences related to underlying use of hgf tokenizers.

Documented in code, but extracting here for visibility:

```
    // Note upgrading from djl v0.21.0 to v0.28.0 (June 2024)
    //
    // In theory, since we're just tokenizing, we shouldn't be constrained by the modelMaxLength.
    // Previously, at v0.21.0, we were able to tokenize arbitrarily long sequences.
    // However, the implementation seems to have changed.
    //
    // As of the v0.28.0 upgrade, if we put a large value, we get the warning:
    // "maxLength is greater then (sic) modelMaxLength, change to: 512"
    //
    // On the other hand, if we don't set this value, we get the warning:
    // "maxLength is not explicitly specified, use modelMaxLength: 512".
    //
    // In other words, the implementation forces truncation, even for our IR application, i.e., it
    // forces FirstP retrieval.
```

### DIFF
--- a/docs/regressions/regressions-dl19-doc-segmented.wp-ca.md
+++ b/docs/regressions/regressions-dl19-doc-segmented.wp-ca.md
@@ -76,13 +76,13 @@ With the above commands, you should be able to reproduce the following results:
 
 | **AP@100**                                                                                                   | **BM25 (default)**|
 |:-------------------------------------------------------------------------------------------------------------|-----------|
-| [DL19 (Doc)](https://trec.nist.gov/data/deep2019.html)                                                       | 0.2604    |
+| [DL19 (Doc)](https://trec.nist.gov/data/deep2019.html)                                                       | 0.2606    |
 | **nDCG@10**                                                                                                  | **BM25 (default)**|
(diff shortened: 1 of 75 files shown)


theyorubayesian:
This is interesting. Does this happen for tokenizers trained & stored locally and tokenizers attached to models on HF?


+ Tagging @ToluClassics for thoughts.

lintool:
> This is interesting. Does this happen for tokenizers trained & stored locally and tokenizers attached to models on HF?
> 
> * Tagging @ToluClassics for thoughts.

Not sure. In this PR are all the scores that changed. Everything else didn't.

theyorubayesian:
I took a look and tested it with the GPT-2 tokenizer. The DJL implementation doesn't use the modelMaxLength, even if it is set in the tokenizer_config.json.

lintool:
Supersede by #2536 which is the better solution.

PR #2537, opened 2024-06-25
Add Reproduction Log in start-here

FaizanFaisal25:
Reproduced the instructions from "start-here" doc.
  - macOS Sonoma 14.5 (M2)
  - Everything worked as expected


lintool:

lintool (approved):

PR #2538, opened 2024-06-25
Refactor tolerance settings for BEIR dense vector regressions

lintool:
+ tweak docs for flat indexes.
+ refactor tolerance values for HNSW indexes, calibrate wrt flat index scores.

PR #2541, opened 2024-06-29
Refactor tolerance settings for MS MARCO dense vector regressions

lintool:
Continuation of #2538 

+ tweak docs for flat indexes.
+ refactor tolerance values for HNSW indexes, calibrate wrt flat index scores.


lintool:
@MXueguang @tteofili I've verified all results, we should be good to go.

MXueguang (approved):

tteofili (approved):

PR #2556, opened 2024-07-31
documentation update for splade++ huggingface -> onnx

AndreSlavescu:
Updated documentation for up to date e2e splade conversion to onnx (huggingface -> onnx)

### DIFF
--- a/docs/onnx-conversion.md
+++ b/docs/onnx-conversion.md
@@ -2,53 +2,99 @@
 This MD file will describe steps to convert particular PyTorch models (i.e., [SPLADE++](https://doi.org/10.1145/3477495.3531857)) to ONNX models and options to further optimize compute graph for Transformer-based models. For more details on how does ONNX Conversion work and how to optimize the compute graph, please refer to [ONNX Tutorials](https://github.com/onnx/tutorials#services).
 
 ## Setups
-- pip requirements:
-    ```
-    onnx                          1.13.1
-    onnxoptimizer                 0.3.10
-    onnxruntime                   1.11.1
-    ```
+
 - SPLADE [Git repo](https://github.com/naver/splade):
    ```bash
    git clone https://github.com/naver/splade.git
    ```
 
+- pip requirements:
+    ```bash
+    pip install torch transformers onnx onnxruntime onnxoptimizer
+    ```
+
 ## Converting from PyTorch models to ONNX model
 The following sections will describe how to convert SPLADE++ model to ONNX model. The steps are as follows:  
 
 ### Load the corresponding HuggingFace model
+
 ```python
-from splade.models.transformer_rep import Splade
+from splade.splade.models.transformer_rep import Splade # this assumes you are writing the script in the same directory as where Splade was cloned
 import torch
 
-model_type_or_dir = "naver/splade-cocondenser-ensembledistil"
-model = Splade(model_type_or_dir, agg="max", fp16=True)
+device = "cuda" if torch.cuda.is_available() else "cpu"
+
+model_name = "naver/splade-cocondenser-ensembledistil"
+model = Splade(model_name, agg="max").to(device)
 model.eval()
 ```
 
-### Creat dummy input variables to be used for ONNX conversion
-```python
-input_ids = torch.randint(1,100, size=(1,50))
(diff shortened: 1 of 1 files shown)


lintool:
@cadurosar can you please take a look at this?

docs/onnx-conversion.md
```diff
@@ -1,75 +1,251 @@
...
-    onnxruntime                   1.11.1
-    ```
-- SPLADE [Git repo](https://github.com/naver/splade):
-   ```bash
-   git clone https://github.com/naver/splade.git
-   ```
+
+pip requirements:
+```bash
+pip install torch transformers onnx onnxruntime onnxoptimizer
```
carlos-lassance, line 15: Are we ok with installing any version or do we want to keep a defined version as it was before?

docs/onnx-conversion.md
```diff
@@ -1,75 +1,251 @@
...
+else:
+    return [f'output_{i}' for i in range(len(outputs))]
 ```
 
-### Creat dummy input variables to be used for ONNX conversion
+This is modularized in the function definition for ```get_model_output_names(model, test_input)```
+
+---
+
+Another important component is being able to specify the dimensions of the input and output tensors. This is achieved by the following code:
```
carlos-lassance, line 64: I would say more that we are giving the dynamic axis names here and not specifying the dimensions

docs/onnx-conversion.md
```diff
@@ -1,75 +1,251 @@
...
+
+### Getting Output Specificaton from the Model
+
+Since we want our script to be generic, and be able to load any huggingface model with simplicity, we make it possible to fetch all required information from the model output structure with the following code:
+
 ```python
-from splade.models.transformer_rep import Splade
-import torch
+with torch.no_grad():
+    outputs = model(**test_input) # We provide the text test input and the model object.
```
carlos-lassance, line 48: Where does model come from here? also test output?

docs/onnx-conversion.md
```diff
@@ -1,75 +1,251 @@
...
-splade.run(None , {"input_ids": inputs['input_ids'], 
-                   "token_type_ids": inputs['token_type_ids'], 
-                   "attention_mask": inputs['attention_mask']})
+To run the script for running inference, run the following sequence of commands:
+```bash
+# Begin by going to the appropriate directory
+cd src/main/python/onnx
+# Now run the script
+python3 run_onnx_model_inference.py --model_path models/splade-cocondenser-ensembledistil-optimized.onnx \
+                                    --model_name naver/splade-cocondenser-ensembledistil
```
carlos-lassance, line 250: maybe an explanation here that we get the model nome to use the tokenizer? 
Another possibility is using the model name to load the hf version and then do a comparison between both to check that things are ok (always reassuring to someone using this for the first time)

carlos-lassance (changes requested):
This looks very good, just a few comments

docs/onnx-conversion.md
AndreSlavescu: Any version should be okay, tested with it on fixed versions that were previously provided and opset 14 isnt supported for the provided onnx. Also I wasn't able to find onnxruntime 1.11.1

docs/onnx-conversion.md
AndreSlavescu: provided in the function definition. I guess I can just copy the full function definitions to make it easier.

docs/onnx-conversion.md
AndreSlavescu: Sounds good. I can add more details for this

docs/onnx-conversion.md
carlos-lassance: Maybe you can put the versions you were using on yours as a secondary one, something like 
`we tested with the following versions 'pip install... ' but we recommend trying first with the latest one from the previous command`

lintool:
@AndreSlavescu Can you also add a blurb on how to test the newly generated SPLADE++ ED model e2e? E.g., point to the repro, "swap in" the new model in `~/.cache/`, etc.

docs/onnx-conversion.md
AndreSlavescu: Yes, sounds good

lintool (approved):

PR #2563, opened 2024-08-06
Fix bugs in regressions involving dense vectors using flat indexes

lintool:
(1) fix incorrect root commands of regressions involving flat indexes
(2) fix incorrect regression scores (in some cases)


lintool:
Recently, I discovered that `flat.onnx` scores weren't accurately captured in the YAML regressions. This, in fact, was by design. All conditions captured the score of `flat.cached`, and I used score tolerances to account for divergences from this "reference". This simplified record keeping in the regressions.

Closing issue.

PR #2564, opened 2024-08-06
Add BEIR BGE prebuilt flat indexes and repro bindings

lintool:
Still work in progress, but ultimate goal is to make something like this work:

```bash
nohup java -cp target/anserini-0.36.2-SNAPSHOT-fatjar.jar io.anserini.reproduce.RunBeir >& log.flat &
```

"Work" = include BGE regressions using flat indexes.


lintool:
@UShivani3 @16BitNarwhal this is ready for CR now.

lintool:
Everything checks out on my end.

UShivani3 (approved):

Issue #2579, opened 2024-08-27
Refactor ThreadPoolExecutor to use try-with-resources


lintool:
We use ThreadPools in many places throughout the codebase, e.g.,
https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/HnswDenseSearcher.java

Best practices starting Java 19 I believe is to use try-with-resources blocks, e.g.,
https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/concurrent/ExecutorService.html

```
 try (ExecutorService e =  Executors.newWorkStealingPool()) {
   // submit or execute many tasks with e ...
 }
```

lintool:
Just start working on it?

PR #2584, opened 2024-08-31
Refactoring on the Anserini end to enable cleaner Pyserini bindings

lintool:
+ `batch_search` in dense searchers - changed method signature to take queries, then qids - to be consistent with `SimpleSearcher` and `SimpleImpactSearcher`
+ dense searchers: refactor `ThreadPoolExecutor` to use try-with-resources, see https://github.com/castorini/anserini/issues/2579


lintool:
@MXueguang this is ready for review. The code coverage check is pointing at lines that are difficult to test, so I'm going to punt on that now.

Running regressions - all seems good so far. Will circle back and confirm but CR can proceed in parallel.

MXueguang (approved):

PR #2590, opened 2024-09-06
Adds fusion feature for Anserini

Stefan824:
The design is following :
- One TrecRun class that handles TrecRun-related methods including file-read, rescore, clone...
- One main class TrecRunFuser that calls methods from TrecRun and actually does the fusion
- One entry point class FuseTrecRuns that is in charge of only arguments and dispatches

Did a number of tests on codes but not thorough enough. Will test more if the design is accepted.

Should close https://github.com/castorini/anserini/pull/2355 and the results align with pyserini.


pom.xml
```diff
@@ -516,5 +516,17 @@
...
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13.2</version>
+      <scope>test</scope>
+    </dependency> 
+    <dependency>
+        <groupId>org.junit.jupiter</groupId>
```
lintool, line 526: algin indentation with above

src/main/java/io/anserini/trectools/TrecRun.java
```diff
@@ -0,0 +1,264 @@
...
+  private Boolean reSort = false;
+  
+  // Constructor without reSort parameter
+  public TrecRun(Path filepath) throws IOException {
+    this(filepath, false);
+  }
+
+  // Constructor with reSort parameter
+  public TrecRun(Path filepath, Boolean reSort) throws IOException {
+      this.resetData();
```
lintool, line 54: consistent two space indentation?

src/main/java/io/anserini/trectools/TrecRun.java
```diff
@@ -0,0 +1,264 @@
...
+      this.readRun(filepath);
+  }
+
+  // Constructor without parameters
+  public TrecRun() {
+      this.resetData();
+  }
+
+  private void resetData() {
+      runData = new ArrayList<>();
```
lintool, line 66: same here

src/test/resources/simple_trec_run_fusion_1.txt
```diff
@@ -0,0 +1,3 @@
+1 Q0 pyeb86on 1 24 reciprocal_rank_fusion_k=60
```
lintool, line 1: put all these files under a `fusion/` directory please.

src/test/java/io/anserini/trectools/TrecRunTest.java
```diff
@@ -0,0 +1,76 @@
+package io.anserini.trectools;
```
lintool, line 1: add boilerplate to here also?

lintool (changes requested):
I don't think there's a need for a new package `src/main/java/io/anserini/trectools/` - just dump everything into `io.anserini.fusion` for now?

lintool:
Oh, I see, you're modeling the structure after pyserini... yea, let's do a from-ground-up redesign.

src/test/resources/simple_trec_run_fusion_1.txt
Stefan824: These test files were actually from Daniel early on; I didn't include my test files on the commit. Of course, I could move these to a `fusion/` directory or just delete them.

lintool: If these files are actually used, then please move. If they're not, then remove.

lintool (changes requested):
Can you please look more carefully at the test cases?

src/test/java/io/anserini/fusion/TrecRunTest.java
```diff
@@ -0,0 +1,92 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+public class TrecRunTest {
+  private TrecRun trecRun;
+  private Path sampleFilePath;
+
+  @Before
+  public void setUp() throws IOException {
+    sampleFilePath = Paths.get("runs/testlong/run.neuclir22-zh-en-splade.splade.topics.neuclir22-en.splade.original-desc_title.txt");
```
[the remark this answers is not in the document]
Stefan824: So sorry. The file seemed a bit too big for git. The plan was to bring this up when doing more tests after the code style/structure is accepted.

lintool (approved):
good enough for now...

PR #2608, opened 2024-09-21
Reproduction Log Update

andrewxucs:
Successfully reproduced steps 1 and 2 of onboarding

**Setup:**
OS: mac OS Sonoma 14.6.1
Chip: Apple M1
Memory: 8 GB

**Problems Encountered**
1.  the command "wget" is actually not pre-installed in my computer, so I just used "curl -o". It works just as well. Since this issue is easy to resolve I think it is not necessary to mention it in the guide
2.  threads aborted during indexing. I tried again after a system update (I didn't actually update because of this onboarding) and then it somehow worked
3.  mvn clean package worked but it took a very long time (approximately 40 minutes on my side). I think telling other hackers how long they should expect to wait is a good idea, but that may not improve readability

### DIFF
--- a/docs/experiments-msmarco-passage.md
+++ b/docs/experiments-msmarco-passage.md
@@ -526,3 +526,4 @@ The BM25 run with default parameters `k1=0.9`, `b=0.4` roughly corresponds to th
 + Results reproduced by [@Amirkia1998](https://github.com/Amirkia1998) on 2024-09-20 (commit [`9e0cd5b`](https://github.com/castorini/anserini/commit/204bc87ef66e689773549ff804377eae20d5d7ce))
 + Results reproduced by [@CCarolD](https://github.com/CCarolD) on 2024-09-20 (commit [`2cb5d13`](https://github.com/castorini/anserini/commit/2cb5d1377862d49f70fa60cc452e96b31d815b94))
 + Results reproduced by [@pjyi2147](https://github.com/pjyi2147) on 2024-09-20 (commit [`2cb5d13`](https://github.com/castorini/anserini/commit/2cb5d1377862d49f70fa60cc452e96b31d815b94))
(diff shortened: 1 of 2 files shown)


lintool (changes requested):
Please fix conflicts, also make sure you've updated `tools/`, there shouldn't be a diff for `topics-and-qrels/topics.beir-v1.0.0-bioasq.test.bge-base-en-v1.5.jsonl.gz`

docs/experiments-msmarco-passage.md
```diff
@@ -526,3 +526,4 @@ The BM25 run with default parameters `k1=0.9`, `b=0.4` roughly corresponds to th
 + Results reproduced by [@Amirkia1998](https://github.com/Amirkia1998) on 2024-09-20 (commit [`9e0cd5b`](https://github.com/castorini/anserini/commit/204bc87ef66e689773549ff804377eae20d5d7ce))
 + Results reproduced by [@CCarolD](https://github.com/CCarolD) on 2024-09-20 (commit [`2cb5d13`](https://github.com/castorini/anserini/commit/2cb5d1377862d49f70fa60cc452e96b31d815b94))
 + Results reproduced by [@pjyi2147](https://github.com/pjyi2147) on 2024-09-20 (commit [`2cb5d13`](https://github.com/castorini/anserini/commit/2cb5d1377862d49f70fa60cc452e96b31d815b94))
++ Results reproduced by [@andrewxucs](https://github.com/andrewxucs) on 2024-09-21 (commit [`f04321f`](https://github.com/castorini/anserini/commit/f04321f40b6eb64308ea90394749912b6199589d4))
```
lintool, line 529: https://github.com/castorini/anserini/commit/f04321f40b6eb64308ea90394749912b6199589d4 -> this URL 404s for me...

PR #2612, opened 2024-09-24
Align all parquet BEIR yaml configs with jsonl versions

lintool:
{flat, hnsw} x {fp32, int8}: aligned scores, tolerances, etc.


lintool:
cc @valamuri2020

MXueguang (approved):

PR #2614, opened 2024-09-26
Add to onboarding reproduction logs (#2614)

Hossein-Molaeian:
windows OS with conda environment. a bit of adjustment needed for indexing and retrieval (xms to 512mb and xmx to 4gb eventhough my machine runs on 24gb of ram, also 4 threads in the command instead of 9). I encountered more errors on Windows than on Linux, but most of them can be easily fixed with a quick online search. Other than these, everything worked just fine.


lintool:
hi @Hossein-Molaeian thanks for your PR. Given that we have relatively few Windows users, please remove your Windows commands on this PR - we'll leave it as a exercise to the user to figure out the equivalent commands. Thanks!

lintool (approved):

Issue #748, opened 2019-07-22
Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project anserini: Fatal error compiling: invalid target release: 11


rajarsheem:
how to resolve this issue?

emmileaf:
Anserini was recently upgraded to Java 11 (commit [`17b702d`](https://github.com/castorini/anserini/commit/17b702d9c3c0971e04eb8386ab83bf2fb2630714)). We recommend using Java 11 if you would like to compile the latest commits, but to compile with Java 8, you can also edit/revert the following lines in `pom.xml`: https://github.com/castorini/anserini/blob/33d630cec4d07e926a022731046a0553aa928fb5/pom.xml#L81-L84

rajarsheem:
Yeah, I solved it already. Forgot to close the issue. Thanks @emmileaf .

yongbowin:
The same error occurs when my executing, modify `11` to your version perform well. Thx.

rajarsheem:
@yongbowin I instead updated my Java to 11 ;)

lintool:
What error are you getting? Have you trying searching online for potential solutions?

techylife:
It is actually giving these much warnings and errors. 
![image](https://user-images.githubusercontent.com/59837849/110493785-f4220300-8118-11eb-9e46-a690a1553da3.png)

lintool:
Unfortunately, none of our team has any experience with Java 15, and your error trace doesn't seem to offer much info...

techylife:
ok

bajeelapv:
Same issue for me also. I am using java 11

HashirLabs:
For me it was working till jdk 15, i just updated it to jdk 16, and started getting the compilation error in intellij

Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project XXX: Fatal error compiling

lintool:
Unfortunately, Anserini is currently only supported on Java 11.

kasuohei:
The same error occurring for me. I am using Java 17. But this solution is not working for me. Would be much thankful if got some help 🙏

nikunj23190:
I update my target release from 11 to 8 and it worked,

Bhrema:
[...]

PR #2623, opened 2024-10-21

Samantha-Zhan:
Details on my setup (e.g., operating system, environment and configuration, etc.):
- Windows Ubuntu 24.04.1 LTS (WSL2)
- Java 21

The setup actually took me quite some time for a couple of reasons:
1. I jumped straight into [start-here](https://github.com/castorini/anserini/blob/master/docs/start-here.md) by clicking on links inside the [ura doc](https://github.com/castorini/onboarding/blob/master/ura.md) without reading [readme](https://github.com/castorini/anserini); therefore, I had the repo setup locally on my windows for the first task, only to redo everything on Ubuntu for the second task.
2. My Ubuntu had Java 11, and when I tried to build I got `Fatal error compiling: invalid target release: 21`. Referencing to [an internal post](https://github.com/castorini/anserini/issues/748), I was able to update Java version to 21, following [these steps](https://stackoverflow.com/questions/53365643/windows-subsystem-for-linux-not-recognizing-java-home-environmental-variable).
3. I also got an additional build failure, where build took 13 minutes and the error was: `Crashed tests: io.anserini.search.SimpleImpactSearcherTest`. I actually was not able to locate the exact cause of this issue, and instead continue building with `mvn clean package -DskipTests -Dmaven.javadoc.skip=true` instead of `mvn clean package`, since this is irrelevant to current tasks.
image
4. In addition, when I am trying to run `bin/run.sh` script to create inverted index, I got `java.lang.SecurityException` for `anserini-0.38.1-SNAPSHOT-fatjar.jar` file. I had to manually unzip the jar, remove the signatures, and zip it back up.
5. Following `experiments-msmarco-passage.md`, when created inverted index, I had to modify `run.sh` to change the -Xms option to 2GB and -Xmx to 6GB due to insufficient memory.

With that said, the actual theoretical knowledge and scripts in the tasks are very easy to follow and understand; I was also able to figure out unexplained terms and configurations through a quick google search.

### DIFF
--- a/docs/experiments-msmarco-passage.md
+++ b/docs/experiments-msmarco-passage.md
@@ -533,4 +533,5 @@ The BM25 run with default parameters `k1=0.9`, `b=0.4` roughly corresponds to th
 + Results reproduced by [@sisixili](https://github.com/sisixili) on 2024-10-01 (commit [`25523b4`](https://github.com/castorini/anserini/commit/25523b48597b4f061f7d016f888a124028b9b01f))
 + Results reproduced by [@Raghav0005](https://github.com/Raghav0005) on 2024-10-07 (commit [`ee97c1d`](https://github.com/castorini/anserini/commit/ee97c1d5deeb684748723179711128f67557832c))
 + Results reproduced by [@a-y-m-a-n-c-h](https://github.com/a-y-m-a-n-c-h) on 2024-10-16 (commit [`0346842`](https://github.com/castorini/anserini/commit/03468423c820e1c0c38c9f48dc25d1f2f315831c))
++ Results reproduced by [@Samantha-Zhan](https://github.com/Samantha-Zhan) on 2024-10-20 (commit [`daceb40`](https://github.com/castorini/anserini/commit/daceb4084c8e8103e3e86c81a8e0d597d409220e))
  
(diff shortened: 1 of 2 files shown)


lintool:
Hi @Samantha-Zhan 

+ re: (3) any more details on what happened?
+ re: (4) what security exception did you get?

This is the first I've heard/seen of these issues, so I'd like to dig into a bit more...

Samantha-Zhan:
- (3) any more details on what happened?
I simply ran `mvn clean package` after recursively pulling the repo following the [installation guide](https://github.com/castorini/anserini?tab=readme-ov-file#-installation). I attempted it multiple times and realized the exact test that crushed seemed to be pretty arbitrary, and time it took to crush ranging from 5 minutes to 10 minutes:
image
Looking deeper into the error message, `Process Exit Code: 137` means the process terminated due to out of memory. So I proceed to increase memory in[ pom.xml ](https://github.com/castorini/anserini/blob/master/pom.xml#L101)file from 8gb to 12gb, but it's still reporting code 137.
- (4) what security exception did you get?
```
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.SecurityException: Invalid signature file digest for Manifest main attributes
```
The same error in #2621 and #2615 ; I took the same approach as #2615 to resolve this.

lintool (approved):

lintool:
Hrm... really odd, on both accounts. Apparently, you're not the first one to report (4)... but it doesn't appear to be pervasive. We'll keep an eye on it.

PR #2630, opened 2024-11-21
Adds baselines for rag24.test with umbrela qrel

UShivani3:


src/main/resources/regression/rag24-doc-segmented-test.yaml
```diff
@@ -0,0 +1,101 @@
+---
```
lintool, line 1: should we name `...test-umbrela-all`?

src/main/resources/docgen/templates/rag24-doc-segmented-test.template
```diff
@@ -0,0 +1,52 @@
+# Anserini Regressions: TREC 2024 RAG Track Test Topics
+
+**Models**: various bag-of-words approaches on segmented documents
+
+This page describes experiments, integrated into Anserini's regression testing framework, on the "RAGgy dev topics" on the MS MARCO V2.1 _segmented_ document corpus.
```
lintool, line 5: "RAGgy dev topics" -> not accurate

lintool (changes requested):
after you run `bin/build.sh`, it'll create a doc in `docs/regressions`; check that in also.

lintool:
We should probably add the "doc" condition also?

src/main/resources/regression/rag24-doc-segmented-test.yaml
lintool: Overriding above; nope, this is fine.

lintool (approved):

PR #2634, opened 2024-11-23
Update start-here.md

Divyajyoti02:
Ran successfully on Macbook air, 256 GB disk, 16 GB RAM, M2 chip. I would like to know what the artifact column in the qrels file is, but I will look into it regardless. Good tutorial!


Divyajyoti02:
Merged to #2636

PR #2636, opened 2024-11-23
Update experiments-msmarco-passage.md

Divyajyoti02:
There were issues in building the Anserini package on the Macbook Air M2 chip with 16 GB RAM and 256 GB memory. On building the package with maven (mvn clean package), the following error happens:

[ERROR] Failures: 
[ERROR]   ControllerTest.testSearch:42 expected:<10> but was:<0>

The error went away when I passed the DskipTests flag. The results were then reproducible successfully.

Also, it would be better for beginners if it is explained why an inverted index is necessary.


lintool:
Can you please collapse this and #2634 into a single PR?

Also, please provide more details regarding what happened to the failed test... we should look into more deeply.

lintool:
Please fix conflicts.

Divyajyoti02:
Fixed conflicts

lintool (approved):

PR #2638, opened 2024-11-25
Install RAG24 baseline + update docs

lintool:
Note that rag2024 regression doesn't actually work yet, we're blocked on https://github.com/castorini/anserini/pull/2639

However, since @UShivani3 is currently away, I'd like to get his merged first and then circle back to fix.


README.md
```diff
@@ -291,12 +291,28 @@ Key:
 
 </details>
 <details>
-<summary>MS MARCO V2.1 Document Regressions</summary>
+<summary>MS MARCO V2.1 Passage Regressions (TREC RAG 24)</summary>
```
ronakice, line 294: Terminology here can be misleading, if we are doing segmented maybe we should say that instead of passage?

ronakice (changes requested):
One main point, I think besides this its good to go.

ronakice (approved):
lgtm

PR #2639, opened 2024-11-25
Fixes type reader

UShivani3:


lintool:
Needs more tweaking before merging. Hold.

lintool:
The PR fixes the reader issue, but is incomplete (I wanted more changes). However, since @UShivani3 is currently unavailable, we'll merge and then circle back to address.

lintool (approved):

PR #2640, opened 2024-11-27
Install MS MARCO Passage V1 Parquet regressions

lintool:
Performed Faiss to Parquet conversion, setting up regressions from Parquet encoding of dense vectors.


lintool:
Removed this statement from docs because it isn't true anymore...

> Note that here we are explicitly using Lucene's `NoMergePolicy` merge policy, which suppresses any merging of index segments.
> This is because merging index segments is a costly operation and not worthwhile given our query set.

ronakice (approved):

ronakice:
LGTM!

PR #2649, opened 2024-12-06
Remove 'NoMerge' statement in docs for BEIR

lintool:
This isn't true anymore... so removing:

> Note that here we are explicitly using Lucene's `NoMergePolicy` merge policy, which suppresses any merging of index segments. This is because merging index segments is a costly operation and not worthwhile given our query set.


ronakice:
lg!

ronakice (approved):

Issue #2654, opened 2024-12-15
ParquetDenseVectorCollection reads all records in memory before processing...


lintool:
This leads to unnecessarily high memory consumption.

At https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/collection/ParquetDenseVectorCollection.java#L124 -we can see that the `initializeParquetReader` method reads _all_ records into memory. It shouldn't do that. Instead, the class should be incrementally pulling records when `readNext()` is called...

PR #2655, opened 2024-12-16
Refactor `ParquetDenseVectorCollection.java` to use incremental streaming

b8zhong:
Related to https://github.com/castorini/anserini/issues/2654

Removed all preloaded lists (of vectors, ids, and contents) that were causing OOM.

Added incremental reading w/ `ParquetReader<Group>` object, and updated `ReadNext()` to access records from the reader


b8zhong:
Works to index MS MARCO

b8zhong:
Closing this as we're using https://github.com/castorini/anserini/pull/2658 instead

PR #2628, opened 2024-11-09
Onboarding: Successful Anserini Reproduction

Sinanmz:
Hi, 

I’m using a MacBook Pro with an M2 Pro chip, running:
- Python 3.9.20
- Maven 3.9.9
- Java 21.0.5

Everything was successful, but when running `mvn clean package`, 6 of the tests failed. I was able to proceed by using `mvn clean package -DskipTests` instead.

PR #2657, opened 2024-12-16
Reproducing Results of BM25 Baselines for MS MARCO

Alireza-Zwolf:
Macbook Air M3 (macOS 14.6)
Python 3.13.1
Java openJDK 23.0.1
Maven 3.9.9

All the results were reproduced without any problem. However during the build time using Maven I encountered errors on 5 test cases on these classes: 
- JsonVectorCollectionDocumentObjectTest, BM25StatTest, BigramFeaturesTest, LmDirTest, ExtractTopDfTermsTest

and the errors were like this:
- [ERROR]   JsonVectorCollectionDocumentObjectTest The test or suite printed 16688 bytes to stdout and stderr, even though the limit was set to 8192 bytes. Increase the limit with @limit, ignore it completely with @SuppressSysoutChecks or run with -Dtests.verbose=true

Basically the errors indicate that the tests are failing because the amount of output printed to stdout and stderr during the test execution exceeded the allowed limit of 8192 bytes. It can be solved by increasing the stdout buffer limit (e.g. adding @limit(bytes = 30000) on top of each mentioned classes) or simply build the project using Maven with -Dtests.verbose=true flag which ignore the buffer limit.


lintool:
Hi @Alireza-Zwolf - it's not clear where the exceptions are coming from...

Is it a JDK issue? I see you're using JDK 23. As stated here: https://github.com/castorini/anserini?tab=readme-ov-file#-installation

Can you please downgrade to JDK 21?

lintool:
Hrm, seems like #2628 outlines the same issues... @zdann15 @sherloc512 @nihalmenon you recently completed onboarding... did you encounter?

sherloc512:
@lintool, I encountered some errors on the test cases when building using maven using JDK 23, but then it was fixed after I switched to JDK 21

Alireza-Zwolf:
Hi @lintool, I downgraded my JDK 21 and the issue was fixed. Seems it was a JDK problem on Java 23.

PR #2658, opened 2024-12-16
Refactor ParquetDenseVectorCollection and added tests

zdann15:
Reduces memory consumption (issue #2654)


src/main/java/io/anserini/collection/ParquetDenseVectorCollection.java
```diff
@@ -164,19 +141,34 @@ private void initializeParquetReader(java.nio.file.Path path) throws IOException
...
+        reader.close();
+        readerInitialized = false;
         throw new NoSuchElementException("End of file reached");
       }
-
-      // Get the current document's ID, contents, and vector
-      String id = ids.get(currentIndex);
-      double[] vector = vectors.get(currentIndex);
+      
+    // Read each record from the Parquet file
```
lintool, line 155: fix indentation

lintool (changes requested):
minor nit

src/main/java/io/anserini/collection/ParquetDenseVectorCollection.java
zdann15: fixed

lintool (approved):

Issue #2662, opened 2024-12-18
Current Parquet files store each vector component as a double


lintool:
... do we really need doubles? what about just floats?

Try this - take the Parquet data for MS MARCO, regenerate a new copy using float instead of double, and see how much the size changes... report back here in this issue.

b8zhong:
@lintool , I did it for the first chunk, but I imagine it's exactly the same for the rest.

```############ Column(element) ############
name: element
path: vector.list.element
max_definition_level: 3
max_repetition_level: 1
physical_type: FLOAT
logical_type: None
converted_type (legacy): NONE
compression: SNAPPY (space_saved: 0%)```

It's 4.4 vs 2.9G, so like 34% less.

lintool:
wow, that's a big difference... we should probably rewrite all our parquet files to use floats.

can you modify the indexer so that it can take both floats and ints? I assume there's an introspection method that will tell you the type?

b8zhong:
@lintool, we can go ahead and close this now, right?

lintool:
Yup!

PR #2667, opened 2024-12-22
Dynamic type support for ParquetDenseVectorDocumentGenerator

b8zhong:
To solve issue https://github.com/castorini/anserini/issues/2662

Not done (nor do I have tests yet!). But let me know what you think of this @lintool


lintool:
No, I don't think you're doing it at the best place... I was thinking in `ParquetDenseVectorCollection` somewhere here:
https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/collection/ParquetDenseVectorCollection.java#L166

i.e., at the point you're reading in the data from the file.

b8zhong:
Ohh - got it that makes sense. Will update

lintool:
We have a test failure here?

We currently have `src/test/resources/sample_docs/parquet/msmarco-passage-bge-base-en-v1.5.parquet` - what about a test cases where we have a variant of the parquet file stored as floats, and we make sure both work?

vincent-4:
Updated the test format to use the same structure as previously– (in the two files below)
With new entries in:
```
src/test/java/io/anserini/index/IndexFlatDenseVectorsTest.java
src/test/java/io/anserini/index/IndexHnswDenseVectorsTest.java
```


And added a new, converted parquet file (floats) in the path (used by the new tests): ```src/test/resources/sample_docs/parquet/msmarco-passage-bge-base-en-v1.5.parquet-float/test1.parquet```

src/main/java/io/anserini/collection/ParquetDenseVectorCollection.java
```diff
@@ -152,19 +153,24 @@ protected synchronized void readNext() throws IOException, NoSuchElementExceptio
...
+      // We detect the type from the schema
+      Group firstElement = vectorGroup.getGroup(0, 0);
+      boolean isDouble = firstElement.getType().getFields().get(0).asPrimitiveType().getPrimitiveTypeName().equals(PrimitiveType.PrimitiveTypeName.DOUBLE);
+      
+      // Single-pass read with conditional cast if needed
       for (int i = 0; i < vectorSize; i++) {
-        Group listGroup = vectorGroup.getGroup(0, i); // Access the 'list' group
-        vector[i] = listGroup.getDouble("element", 0); // Get the double value from the 'element' field
+        Group listGroup = vectorGroup.getGroup(0, i);
+        vector[i] = isDouble ? (float) listGroup.getDouble("element", 0) : listGroup.getFloat("element", 0);
```
lintool, line 171: Here, you're assuming if it's not a double, it's a float, right? But what if it's an int or some other type?

How about we check explicitly for double and float, and throw an exception if it's neither?

lintool (changes requested):
Thoughts?

lintool (approved):

PR #2668, opened 2024-12-24
Prune Parquet Dependencies

vincent-4:
Pruned with https://github.com/strategicblue/parquet-floor, as well as added tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -550,7 +550,7 @@
     <dependency>
       <groupId>blue.strategic.parquet</groupId>
       <artifactId>parquet-floor</artifactId>
-      <version>1.49</version>
+      <version>1.36</version>
     </dependency>
   </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -42,6 +42,10 @@
 
   <repositories>
     <repository>
+      <id>jitpack.io</id>
+      <url>https://jitpack.io</url>
+    </repository>
+    <repository>
       <id>public</id>
       <url>https://repo1.maven.org/maven2</url>
     </repository>
@@ -50,8 +54,8 @@
       <url>https://raw.githubusercontent.com/lintool/AnseriniMaven/master/mvn-repo/</url>
     </repository>
     <repository>
-      <id>jitpack.io</id>
-      <url>https://jitpack.io</url>
+      <id>strategicblue</id>
+      <url>https://raw.githubusercontent.com/strategicblue/parquet-floor/mvn-repo/</url>
     </repository>
   </repositories>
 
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -50,8 +54,8 @@
       <url>https://raw.githubusercontent.com/lintool/AnseriniMaven/master/mvn-repo/</url>
     </repository>
     <repository>
-      <id>jitpack.io</id>
-      <url>https://jitpack.io</url>
+      <id>strategicblue</id>
```
lintool, line 57: I don't think we need this?
The artifact is published in Maven central:
https://central.sonatype.com/artifact/blue.strategic.parquet/parquet-floor

src/test/java/io/anserini/collection/ParquetDenseVectorCollectionTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class ParquetDenseVectorCollectionTest extends DocumentCollectionTest<ParquetDenseVectorCollection.Document> {
```
lintool, line 21: Move "Tests for the ParquetDenseVectorCollection class...." comment here.

src/test/java/io/anserini/collection/ParquetDenseVectorCollectionTest.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Tests for the ParquetDenseVectorCollection class which handles dense vector embeddings stored in Parquet format.
+ * This test suite verifies the collection's ability to read and process Parquet files containing vector embeddings
+ * using the parquet-floor library instead of Hadoop dependencies.
+ */
+package io.anserini.collection;
```
lintool, line 6: Add standard Apache 2 license boilerplate.

lintool (changes requested):
Nice!

What's the fatjar size, before and after?

vincent-4:
@lintool
See [3e85803](https://github.com/castorini/anserini/pull/2668/commits/3e858036a866c2bef5e96276056f69c61b695453) for the changes that address your feedback. I updated the Parquet format to version 1.49 in the same commit, but it caused issues. [d17afc8](https://github.com/castorini/anserini/pull/2668/commits/d17afc89b10ff9f7c67408c99b5854be2dbd121a) resolves this problem.

lintool:
🎉 Yay for smaller size!

```
    <repository>
      <id>jitpack.io</id>
      <url>https://jitpack.io</url>
    </repository>
```
Do we still need this though?

vincent-4:
I keep the jitpack.io repository as it's required for com.github.TREMA-UNH:trec-car-tools-java dependency.
See: https://github.com/TREMA-UNH/trec-car-tools-java/blob/master/pom.xml#L14-L16 
If I'm not wrong? I don't find it on Maven central.

lintool:
Ah, okay! I'll queue this up for regression testing (machines occupied right now) before I merge.

What tests have you run, BTW? Did the MS MARCO v1 tests pass?

lintool:
Need to run more regressions, but initial signs are 👍 - the following passed:

```
python src/main/python/run_regression.py --download --index --verify --search --regression msmarco-v1-passage.bge-base-en-v1.5.parquet.flat.onnx
```

vincent-4:
Any updates on the regressions?

lintool:
@vincent-4 why are we on v1.36 and not a later version?

vincent-4:
> @vincent-4 why are we on v1.36 and not a later version?

Without modifying the existing codebase, I tested `parquet-floor` versions between 1.36 and 1.49. The latest version that doesn't produce upgrade-related errors using `mvn clean verify` is 1.36.

I have two possible options:

1. Continue working on fixing the (several) errors related to switching to the latest version of `parquet-floor` (which would require more changes).
2. Implement all the changes now if your priority is just to use parquet-floor, which should be fine given the successful regressions and MSMARCO tests. If we choose this option, I could try upgrading to the latest version in another PR.

Let me know which approach you prefer?

lintool:
I ran all Parquet regressions:

```bash
cat src/main/python/regressions-batch0* | grep parquet > src/main/python/regressions-parquet.txt
nohup python src/main/python/run_regressions_with_load.py --file src/main/python/regressions-parquet.txt --load 64 --sleep 60 >& regressions.parquet.log &
```

Everything checks out. Ready to merge.

lintool (approved):

PR #2672, opened 2025-01-06
Rename prebuilt indexes: cos-dpr -> cosdpr

lintool:
Goal is to make prebuilt index names consistent with Pyserini. Lots of minor tweaks...


lintool:
Ready for review. I will cut a release after this.

MXueguang (approved):

Issue #2661, opened 2024-12-17
SourceDocument should provide method for vector


lintool:
Currently, `SourceDocument` only provides `contents()` which returns a String.

This means that for dense vectors, we first convert into a string and then parse back into a vector... which is silly.

We should add something like

```
float[] vector();
```

lintool:
@vincent-4 how is this going?

vincent-4:
I've done most of the implementations– just have to do a sanity check with the tests (existing, maybe add some new ones.)

lintool:
I've confirmed all regressions.

At Anserini commit [`df900e`](https://github.com/castorini/anserini/commit/df900e81c61a6bf0caa2eca13ac18d6b20317d4d) (2025/01/25).

On `tuna` with `pyserini-dev7`, `batch03` and `batch04`:

```bash
$ python src/main/python/analyze_regression_logs.py
Total regressions: 1050
 - Passed:         1026
 - OK:              24

Start time: 2025-01-20 19:59:25,109
End time:   2025-01-25 09:00:07,376

Duration: 4 days, 13:00:42.267000 ~109.0h
```

All good, closing issue. Thanks @vincent-4

PR #2676, opened 2025-01-12
Providingmethod sourcedocument

vincent-4:
Changes made:

1. Added `vector()` method to `SourceDocument` interface as an optional default method that returns `float[]` for direct vector access

2. Implemented `vector()` in collection classes:
   - `JsonDenseVectorCollection`: Returns parsed vector for array format
   - `JsonVectorCollection`: Returns vector for dense format, null for sparse

3. Modified document generators to use direct vector access:
   - Try `vector()` first
   - Fall back to string parsing for backward compatibility
   - Affects: JsonDenseVectorDocumentGenerator, JsonInvertedDenseVectorDocumentGenerator

4. Added vector validation in tests:
   - Added helper method `validateVectorIfPresent` in JsonVectorCollectionTest
   - Verifies correct vector parsing and null handling
   - Tests both array and non-array formats
  
Addresses: #2661

### DIFF
--- a/src/main/java/io/anserini/collection/JsonDenseVectorCollection.java
+++ b/src/main/java/io/anserini/collection/JsonDenseVectorCollection.java
@@ -24,7 +24,6 @@ import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Map;
 
-
 /**
  * A document collection for encoded dense vectors for ANN (HNSW) search.
  * The "vector" field are concatenated into the "contents" field for indexing.
@@ -75,14 +74,14 @@ public class JsonDenseVectorCollection extends DocumentCollection<JsonDenseVecto
       this.id = json.get("docid").asText();
       this.contents = json.get("vector").toString();
       JsonNode vectorNode = json.get("vector");
-        if (vectorNode != null && vectorNode.isArray()) {
-            vectorData = new float[vectorNode.size()];
-            for (int i = 0; i < vectorNode.size(); i++) {
-                vectorData[i] = (float) vectorNode.get(i).asDouble();
-            }
-        } else {
-            vectorData = null;
+      if (vectorNode != null && vectorNode.isArray()) {
+        vectorData = new float[vectorNode.size()];
+        for (int i = 0; i < vectorNode.size(); i++) {
+          vectorData[i] = (float) vectorNode.get(i).asDouble();
         }
+      } else {
+        vectorData = null;
+      }
       // We're not going to index any other fields, so just initialize an empty map.
       this.fields = new HashMap<>();
     }
--- a/src/main/java/io/anserini/collection/JsonDenseVectorCollection.java
+++ b/src/main/java/io/anserini/collection/JsonDenseVectorCollection.java
@@ -66,6 +66,7 @@ public class JsonDenseVectorCollection extends DocumentCollection<JsonDenseVecto
     private final String id;
     private final String contents;
     private final String raw;
+    private final float[] vectorData;
     private Map<String, String> fields;
 
     public Document(JsonNode json) {
@@ -73,6 +74,15 @@ public class JsonDenseVectorCollection extends DocumentCollection<JsonDenseVecto
       this.raw = json.toPrettyString();
       this.id = json.get("docid").asText();
       this.contents = json.get("vector").toString();
+      JsonNode vectorNode = json.get("vector");
+        if (vectorNode != null && vectorNode.isArray()) {
+            vectorData = new float[vectorNode.size()];
+            for (int i = 0; i < vectorNode.size(); i++) {
+                vectorData[i] = (float) vectorNode.get(i).asDouble();
+            }
+        } else {
+            vectorData = null;
+        }
       // We're not going to index any other fields, so just initialize an empty map.
       this.fields = new HashMap<>();
     }
@@ -101,6 +111,11 @@ public class JsonDenseVectorCollection extends DocumentCollection<JsonDenseVecto
     @Override
     public Map<String, String> fields() {
       return fields;
+    }
+
+    @Override
+    public float[] vector() {
+      return vectorData;
     }
   }
 }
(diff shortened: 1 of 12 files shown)


lintool:
@vincent-4 I think it's okay to deprecate the string parsing methods? I.e., remove completely. Once this is in a reasonable state I will verify regressions - there's no need/reason to keep the jank around...

lintool:
hey @vincent-4 this PR is still in draft... is it ready for me to look at?

vincent-4:
Yeah, I would appreciate it. 

Btw, unrelated, but I thought I'd send a message while you're looking at this too: but I left a message in slack about parquet-floor's issues: essentially the re-implementation of parquet's native stuff by parquet-floor breaks, and I can't get it working properly.

src/main/java/io/anserini/collection/JsonDenseVectorCollection.java
```diff
@@ -66,13 +66,23 @@ public static class Document extends JsonCollection.Document {
...
+    private final float[] vectorData;
     private Map<String, String> fields;
 
     public Document(JsonNode json) {
       super();
       this.raw = json.toPrettyString();
       this.id = json.get("docid").asText();
       this.contents = json.get("vector").toString();
+      JsonNode vectorNode = json.get("vector");
+        if (vectorNode != null && vectorNode.isArray()) {
```
lintool, line 78: indentation issue?

src/main/java/io/anserini/collection/JsonVectorCollection.java
```diff
@@ -23,7 +23,8 @@
 import java.nio.file.Path;
 
 /**
- * A JSON document collection where the user can specify directly the vector to be indexed.
+ * A JSON document collection where the user can specify directly the vector to
```
lintool, line 26: revert? keep on single line please.

src/main/java/io/anserini/collection/JsonVectorCollection.java
```diff
@@ -39,7 +40,8 @@ public FileSegment<JsonVectorCollection.Document> createFileSegment(Path p) thro
   }
 
   @Override
-  public FileSegment<JsonVectorCollection.Document> createFileSegment(BufferedReader bufferedReader) throws IOException {
+  public FileSegment<JsonVectorCollection.Document> createFileSegment(BufferedReader bufferedReader)
+      throws IOException {
```
lintool, line 44: revert? don't change things that aren't relevant?

src/main/java/io/anserini/collection/JsonVectorCollection.java
```diff
@@ -60,30 +62,47 @@ protected Document createNewDocument(JsonNode json) {
...
   public static class Document extends JsonCollection.Document {
     private final String contents;
+    private final float[] vectorData;
 
     public Document(JsonNode json) {
       super(json);
 
-      // We're going to take the map associated with "vector" and generate pseudo-document.
+      // We're going to take the map associated with "vector" and generate
+      // pseudo-document.
```
lintool, line 71: same

src/test/java/io/anserini/collection/JsonVectorCollectionTest.java
```diff
@@ -17,6 +17,8 @@
 package io.anserini.collection;
 
 import java.util.Map;
+import org.junit.Test;
+import static org.junit.Assert.*;
```
lintool, line 21: don't use * import please.

src/main/java/io/anserini/search/topicreader/JsonStringVectorTopicReader.java
```diff
@@ -29,11 +29,16 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 public class JsonStringVectorTopicReader extends TopicReader<String> {
+  private final Map<String, float[]> vectorCache = new HashMap<>();
```
lintool, line 32: can you explain why we need a cache? I don't think topics are repeated?

lintool (changes requested):

lintool:
hi @vincent-4 gave you some comments.

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -447,6 +447,31 @@ public static Map<String, Long> getDocumentVector(IndexReader reader, String doc
...
 
+  /**
+   * Returns the dense vector representation of a document, if available.
+   *
+   * @param reader index reader
+   * @param docid document id
+   * @return vector as float array, or null if not available
+   * @throws IOException if error encountered during access to index
+   */
+  public static float[] getDenseVector(IndexReader reader, String docid) throws IOException {
```
lintool, line 458: Do we need this method anymore? I don't think we ever parse from string in the current impl?

src/main/java/io/anserini/collection/JsonVectorCollection.java
```diff
@@ -60,30 +60,46 @@ protected Document createNewDocument(JsonNode json) {
...
-        for (int i=0; i<cnt; i++ ) {
-          sb.append(e.getKey()).append(" ");
+      if (vectorNode.isArray()) {
+        // Dense vector format - store directly
+        vectorData = new float[vectorNode.size()];
+        for (int i = 0; i < vectorNode.size(); i++) {
+          vectorData[i] = (float) vectorNode.get(i).asDouble();
         }
-      });
+        this.contents = vectorNode.toString();
```
lintool, line 77: I'm thinking here... what should be the behavior here? Returning `null` is another option if the document only has vector data? Thoughts?

lintool: If we return `null` here then we would never need `getDenseVector` right?

lintool (commented):
Thoughts?

src/main/java/io/anserini/collection/JsonVectorCollection.java
vincent-4: Yeah then we can omit it..
We can just have one rep. for the vector
Implementing!

lintool: Yup, I think that makes sense, thanks!

lintool:
lg for now. I'm going to run regressions and make sure everything still works before I merge.

lintool (approved):

PR #2678, opened 2025-01-12
Remove LTR code

lintool:
Completely ripping out LTR stuff, because the technology is defunct, and we're just dragging along dead code. Thoughts?


ronakice (approved):
Agreed, not sure who uses these features in Anserini.

PR #2627, opened 2024-11-08
Improved fusion regression

Stefan824:
Improved fusion regression by:

1. Modified fusion yaml files to include regression yamls and run-name
2. Adjusted fuse-regression script accordingly
3. Added error handling for missing run-files

If well, should address https://github.com/castorini/anserini/issues/2619

PR #2683, opened 2025-01-16
Tweak fusion regression

lintool:
Patch on top of #2627


wu-ming233 (approved):
lgtm!

lintool:
Confirmed that this sequence works:

```bash
python src/main/python/run_regression.py --index --verify --search --regression beir-v1.0.0-robust04.flat
python src/main/python/run_regression.py --index --verify --search --regression beir-v1.0.0-robust04.bge-base-en-v1.5.parquet.flat.cached

python src/main/python/run_fusion_regression.py --regression beir-v1.0.0-robust04
```

PR #2686, opened 2025-01-17
Update: ONNX conversion doc

b8zhong:
Update onnx + onnxruntime

Don't think `quantize_onnx_model.py` supports model name flag? Did not work w/ SPLADE++ Ensemble Distil earlier.

https://github.com/castorini/anserini/blob/master/src/main/python/onnx/quantize_onnx_model.py#L19-L24


lintool:
Actually, @b8zhong you know what would be helpful? A snippet of code that compares the encoded output of the hgf model vs. the ONNX model - and take the two vectors and compute l1 norm or something - to confirm that the conversion worked.

would you might adding this?

b8zhong:
@lintool any idea why such a large L1 diff?

```
L1 difference between PyTorch and ONNX outputs: 0.18276220560073853
```

Maybe `AutoModel` should actually be `AutoModelForMaskedLM`... ?

b8zhong:
Ty for the hint.. I don't think I would have ever thought of that 😛

There we go:
`L1 difference between PyTorch and ONNX outputs: 0.009487475268542767`

Acceptable I think?

lintool:
Yup, that's good.

The page as written is specific to SPLADE++, can you make it generic, _using_ SPLADE++ as an example?

b8zhong:
Thoughts?

docs/onnx-conversion.md
```diff
@@ -19,39 +19,59 @@ pip install torch transformers onnx onnxruntime onnxoptimizer
...
-onnx                         1.16.1
+onnx                         1.17.0
 onnxoptimizer                0.3.13
-onnxruntime                  1.18.1
+onnxruntime                  1.20.1
 ```
 
 ## Converting from PyTorch models to ONNX model
-The following sections will describe how to convert SPLADE++ model to ONNX model. The steps are as follows:  
+The following sections will describe how to convert a Transformer-based model to ONNX model. We use the SPLADE++ model as an example. The steps are as follows:
```
lintool, line 28: We use lowercase t for transformers.

lintool: We also use the convention of each sentence on its own line.
This makes diffs easier to read.

docs/onnx-conversion.md
```diff
@@ -360,9 +380,9 @@ quantize_dynamic(
 
 ## Concluding Remarks
 
-Now that we have successfully gone through a complete reproduction of converting SPLADE++ Ensemble Distil from PyTorch to ONNX, and ran inference with the optimized model, the scripts can be used to reproduce any model available on Huggingface.
+Now that we have successfully gone through a complete reproduction of converting SPLADE++ Ensemble Distil from PyTorch to ONNX, and ran inference with the optimized model, the scripts can be used to reproduce any model available on Hugging Face.
 
-### Reproducing SPLADE++ Ensemble Distil Regressions 
+### Reproducing Transformer-based Model Regressions
```
lintool, line 385: Transformer-Based -> capital B in this case.

lintool (changes requested):
minor nits, I'll fix and then I'll merge.

lintool (approved):

Issue #2688, opened 2025-01-19
Server app currently does not support dense retrieval indexes


lintool:
The current webapp/REST API https://github.com/castorini/anserini/blob/master/docs/rest-api.md does not support dense HNSW indexes.

For example, this works, on `msmarco-v1-passage`:

```
curl -X GET "http://localhost:8081/api/v1.0/indexes/msmarco-v1-passage/search?query=How%20does%20the%20process%20of%20digestion%20and%20metabolism%20of%20carbohydrates%20start" 
```

But this doesn't:

```
curl -X GET "http://localhost:8081/api/v1.0/indexes/msmarco-v1-passage.bge-base-en-v1.5.hnsw/search?query=How%20does%20the%20process%20of%20digestion%20and%20metabolism%20of%20carbohydrates%20start" 
```

That's because `msmarco-v1-passage.bge-base-en-v1.5.hnsw` uses an HNSW index, and https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/server/SearchService.java does not provide bindings to HNSW indexes.

We should start by fixing this.

@vincent-4

lintool:
@vincent-4 should we close this issue or more to do here?

vincent-4:
This is good for now, if we have more to work on wrt the interface, it can go in another issue.

PR #2691, opened 2025-01-23
Add HNSW support to the localhost API interface

vincent-4:
Addresses #2688 

- Keep BM25 as base case, treat HNSW as extension to maintain backward compatibility

Example usage:
`curl -X GET 'http://localhost:8080/api/v1.0/indexes/msmarco-v1-passage.bge-base-en-v1.5.hnsw/search?query=How%20does%20the%20process%20of%20digestion%20and%20metabolism%20of%20carbohydrates%20start&hits=10&efSearch=128&encoder=BgeBaseEn15&queryGenerator=VectorQueryGenerator'`

Although would a different API querying method be better? We specify the encoder and generator but without the additional params, (user specified encoder and querygen) it won't work. But the existing usage is not changed, eg
`curl -X GET "http://localhost:8081/api/v1.0/indexes/msmarco-v1-passage.bge-base-en-v1.5.hnsw/search?query=How%20does%20the%20process%20of%20digestion%20and%20metabolism%20of%20carbohydrates%20start"` is fine.


src/main/java/io/anserini/server/ControllerV1_0.java
```diff
@@ -61,10 +61,8 @@ public Map<String, Object> searchIndex(@PathVariable(value = "index", required =
       throw new IllegalArgumentException("Index " + index + " not found!");
     }
 
-    if (index.contains(".hnsw")) {
-      if (encoder == null || queryGenerator == null) {
-        throw new IllegalArgumentException("HNSW indexes require both 'encoder' and 'queryGenerator' parameters");
-      }
+    if (index.contains(".hnsw") && (encoder == null || queryGenerator == null)) {
```
lintool, line 64: Why don't we add a new field in the `Enum` in `IndexInfo` so we don't have to do this janky check?

lintool: Indexes need to be paired with generators and encoders, right? so put that in `IndexInfo`?

lintool:
Another thought: maybe have defaults hard-coded in `IndexInfo`, and then use `api/v1.0/indexes/msmarco-v1-passage.bge-base-en-v1.5.hnsw/setting?param1=value1&param2=value2` to set?

Something like that?

src/main/java/io/anserini/server/ControllerV1_0.java
vincent-4: > Why don't we add a new field in the `Enum` in `IndexInfo` so we don't have to do this janky check?

For each enum? Or something different

lintool: Sure, why not? Add another field in the enum denoting indexing type, the encoder, and the query generator?

vincent-4:
> Another thought: maybe have defaults hard-coded in `IndexInfo`, and then use `api/v1.0/indexes/msmarco-v1-passage.bge-base-en-v1.5.hnsw/setting?param1=value1&param2=value2` to set?
> 
> Something like that?

I guess we could have it per-request or make it index-level (persistent). Does the usual use case make it easier to work with the first or second approach? Or could add both possibilities but more complex

lintool:
>  per-request or make it index-level (persistent)

Should be persistent.

vincent-4:
@lintool  I took down getting the actual doc contents, since we would have to 'load twice' in a sense– want it back?

Also, updated IndexInfo mappings, but there's quite a few left to do (appreciate input on if I'm screwing up this approach).

src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -25,7 +25,14 @@ public enum IndexInfo {
       "BM25",
       new String[] {
           "https://github.com/castorini/anserini-data/raw/master/CACM/lucene-index.cacm.20221005.252b5e.tar.gz" },
+<<<<<<< Updated upstream
```
lintool, line 28: Resolve?

src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -25,7 +25,14 @@ public enum IndexInfo {
...
       "cfe14d543c6a27f4d742fb2d0099b8e0"),
+=======
+      "cfe14d543c6a27f4d742fb2d0099b8e0",
+      IndexType.bm25,
```
lintool, line 32: `IndexType.INVERTED` -> inverted index is what allows BM25, and tfidf, etc.

IIRC, all caps is the Java convention?

src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -46,7 +56,10 @@ public enum IndexInfo {
       "SPLADE++ EnsembleDistil",
       new String[] {
           "https://rgw.cs.uwaterloo.ca/pyserini/indexes/lucene/lucene-inverted.msmarco-v1-passage.splade-pp-ed.20230524.a59610.tar.gz" },
-      "2c008fc36131e27966a72292932358e6"),
+      "2c008fc36131e27966a72292932358e6",
+      IndexType.flat,
```
lintool, line 60: `IndexType.DENSE_FLAT`?

src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -56,7 +69,14 @@ public enum IndexInfo {
       "cosDPR-distil w/ HNSW fp32",
       new String[] {
           "https://rgw.cs.uwaterloo.ca/pyserini/indexes/lucene/lucene-hnsw.msmarco-v1-passage.cosdpr-distil.20240108.825148.tar.gz" },
+<<<<<<< Updated upstream
       "df4c60fa1f3804fa409499824d12d035"),
+=======
+      "df4c60fa1f3804fa409499824d12d035",
+      IndexType.hnsw,
```
lintool, line 76: `IndexType.DENSE_HNSW`?

lintool:
> @lintool I took down getting the actual doc contents, since we would have to 'load twice' in a sense– want it back?

Sorry, don't understand what this means?

lintool:
re: fetching documents - only the inverted indexes store documents.

If I try to fetch doc from an HNSW index, then it should just "re-route" to the inverted index. We can add "pairings" in `IndexInfo` - e.g., which inverted index goes with which HNSW index.

vincent-4:
> > @lintool I took down getting the actual doc contents, since we would have to 'load twice' in a sense– want it back?
> 
> Sorry, don't understand what this means?

When we make a request, we get this:

curl -X GET 'http://localhost:8080/api/v1.0/indexes/msmarco-v1-passage.bge-base-en-v1.5.hnsw/search?query=How%20does%20the%20process%20of%20digestion%20and%20metabolism%20of%20carbohydrates%20start&hits=10&efSearch=128&encoder=BgeBaseEn15&queryGenerator=VectorQueryGenerator'
{"query":{"text":"How does the process of digestion and metabolism of carbohydrates start","qid":""},"candidates":[{"docid":"4812922","score":0.9180973,"doc":null},{"docid":"5721972","score":0.9178153,"doc":null},{"docid":"4517519","score":0.91594195,"doc":null},{"docid":"5918340","score":0.9146468,"doc":null},{"docid":"2969775","score":0.91335136,"doc":null},{"docid":"7494976","score":0.91238195,"doc":null},{"docid":"5164081","score":0.9123745,"doc":null},{"docid":"3145583","score":0.91142595,"doc":null},{"docid":"8699089","score":0.9110645,"doc":null},{"docid":"4953055","score":0.9097354,"doc":null}]}

I omit returning "doc" (which normally has contents with non-HNSW). But I get your latest re:fetching documents comment. Thanks!

lintool:
```
curl -X GET 'http://localhost:8080/api/v1.0/indexes/msmarco-v1-passage.bge-base-en-v1.5.hnsw/search?query=How%20does%20the%20process%20of%20digestion%20and%20metabolism%20of%20carbohydrates%20start&hits=10&efSearch=128&encoder=BgeBaseEn15&queryGenerator=VectorQueryGenerator'
```

I'm thinking: `encoder=BgeBaseEn15&queryGenerator=VectorQueryGenerator` - these aren't needed? As in, they can't be anything else... if you have a bge index, you _must_ use the bge encoder, or you'll get gibberish?

vincent-4:
> ```
> curl -X GET 'http://localhost:8080/api/v1.0/indexes/msmarco-v1-passage.bge-base-en-v1.5.hnsw/search?query=How%20does%20the%20process%20of%20digestion%20and%20metabolism%20of%20carbohydrates%20start&hits=10&efSearch=128&encoder=BgeBaseEn15&queryGenerator=VectorQueryGenerator'
> ```
> 
> I'm thinking: `encoder=BgeBaseEn15&queryGenerator=VectorQueryGenerator` - these aren't needed? As in, they can't be anything else... if you have a bge index, you _must_ use the bge encoder, or you'll get gibberish?

Yeah, that was because:
Prior to adding the tagging, using it with no params is fine. I just had it specified because it wasn't defined the enums yet. But IndexInfo won't  compile just yet beacause of the WIP state. But not having to specify is the end goal.

vincent-4:
Few more interface Qs:
- If an inverted index is listed in the mapping but not found locally, should we automatically download it, or let the user handle that? I know running regressions does it so I guess it would align somewhat.
- 
- Is there a straightforward way where can I check the naming patterns/mappings? I’m only certain about the most commonly used ones (for instance, bge, since I've used it before) but I can't check my correctness on these

lintool:
> Q:
> 
> * Is there a straightforward way where can I check the naming patterns/mappings? I’m only certain about the most commonly used ones (for instance, bge, since I've used it before) but I can't check my correctness on these

Look in `IndexInfo`, which contains the ground truth metadata. We've tried to name consistently, but it's evolved over the years. The intention was for the UI to fetch and display the metadata from `IndexInfo` - to present in a more readable form (and allow user actions like download, etc.).

vincent-4:
Updated IndexInfo (not _really_ worth reviewing)– will update later w/ better search that uses the new enum fields

vincent-4:
On the latest test failure: Error:    PrebuiltIndexTest.testNumPrebuiltIndexes:63 expected:<166> but was:<169>
I reverted the changes though? Weird.

Otherwise this is ready for review, will add some more tests

src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -46,7 +54,11 @@ public enum IndexInfo {
       "SPLADE++ EnsembleDistil",
       new String[] {
           "https://rgw.cs.uwaterloo.ca/pyserini/indexes/lucene/lucene-inverted.msmarco-v1-passage.splade-pp-ed.20230524.a59610.tar.gz" },
-      "2c008fc36131e27966a72292932358e6"),
+      "2c008fc36131e27966a72292932358e6",
+      IndexType.SPLADE_PP_ED,
+      "SpladePlusPlusEnsembleDistil",
```
lintool, line 59: If we changed to `SpladePlusPlusEnsembleDistilEncoder.class` we can reference an actual class as opposed to string, then empty changes can be `null`s. I think I like that better?

(And same for everything below?)

src/main/java/io/anserini/server/ControllerV1_0.java
```diff
@@ -22,35 +22,48 @@
...
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.ExceptionHandler;
 
 @RestController
 @RequestMapping(path = "/api/v1.0")
```
lintool, line 38: Should we increment the API version to, say, v1.1?

(okay old routes are removed, but let's make it explicit?)

lintool (changes requested):
Thoughts?

src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -96,7 +124,11 @@ public enum IndexInfo {
       "cohere-embed-english-v3.0 w/ HNSW fp32",
       new String[] {
           "https://rgw.cs.uwaterloo.ca/pyserini/indexes/lucene/lucene-hnsw.msmarco-v1-passage.cohere-embed-english-v3.0.20240228.eacd13.tar.gz" },
-      "c7294ca988ae1b812d427362ffca1ee2"),
+      "c7294ca988ae1b812d427362ffca1ee2",
+      IndexType.DENSE_HNSW,
+      "CohereEmbedEnglishV30",
```
vincent-4, line 129: Btw, what should be its encoder? I take it literally from the name right above as:
>       "Lucene quantized (int8) HNSW index of the MS MARCO V1 passage corpus encoded by Cohere embed-english-v3.0.",

But we don't have anything like it? I looked in io/anserini/encoder

lintool: Oh, this doesn't have an encoder... you have to call the Cohere API... so should be `null` for now and probably not searchable.

src/main/java/io/anserini/index/IndexInfo.java
vincent-4: Yep

lintool (approved):
lgtm

Issue #2705, opened 2025-02-01
Discussion: Command-line for sharded indexes


lintool:
#2690 is a discussion about the REST API. I want to start a similar discussion about a command-line interface...

Currently, to search over shards, we have to issue _n_ different commands, using a script like this:

```
SHARDS=(00 01 02 03 04 05 06 07 08 09); for shard in "${SHARDS[@]}"
do
    bin/run.sh io.anserini.search.SearchHnswDenseVectors -index msmarco-v2.1-doc-segmented-shard${shard}.arctic-embed-l.hnsw-int8 -efSearch 1000 -topics rag24.test -topicReader TsvString -topicField title -encoder ArcticEmbedL -output runs/run.msmarco-v2.1-doc-segmented.arctic-l.rag24.test.shard${shard}.txt -hits 250 -threads 32 > logs/log.msmarco-v2.1-doc-segmented.arctic-l.rag24.test.shard${shard}.txt 2>&1
done
```

What if we implemented a `SearchShardedHnswDenseVectors` class that has something like the following behavior?

```
bin/run.sh io.anserini.search.SearchShardedHnswDenseVectors \
  -shards msmarco-v2.1-doc-segmented-shard00.arctic-embed-l.hnsw-int8 \
          msmarco-v2.1-doc-segmented-shard01.arctic-embed-l.hnsw-int8 \
          msmarco-v2.1-doc-segmented-shard02.arctic-embed-l.hnsw-int8 \
          msmarco-v2.1-doc-segmented-shard03.arctic-embed-l.hnsw-int8 \
          msmarco-v2.1-doc-segmented-shard04.arctic-embed-l.hnsw-int8 \
          msmarco-v2.1-doc-segmented-shard05.arctic-embed-l.hnsw-int8 \
          msmarco-v2.1-doc-segmented-shard06.arctic-embed-l.hnsw-int8 \
          msmarco-v2.1-doc-segmented-shard07.arctic-embed-l.hnsw-int8 \
          msmarco-v2.1-doc-segmented-shard08.arctic-embed-l.hnsw-int8 \
          msmarco-v2.1-doc-segmented-shard09.arctic-embed-l.hnsw-int8 \
  -efSearch 1000 -topics rag24.test -topicReader TsvString -topicField title -encoder ArcticEmbedL \
  -output runs/run.msmarco-v2.1-doc-segmented.arctic-l.rag24.test.txt -hits 250 -threads 32
```

[...]

PR #2710, opened 2025-02-07
feat: reorganize shards and implement parallel search

vincent-4:
Now, we'll have usage like
`/bin/run.sh .... -index "msmarco-v2.1-doc-segmented.arctic-embed-l.hnsw-int8" ...rest of the command"`

The nice thing is that, if I assume you'd always want to search across all shards, there's a way to find all shards: otherwise, radically reorganizing the original enum is not very organized. It'll automatically find the rest of the shards as defined in the new enum type in ShardInfo.

There is an issue, where performance is either 3.11 Queries/s or as low as 0.33 Queries/s on `orca`, but it might just have to do with local resource usage– in the middle of investigating.

Posting for posterity–
Review when you like.

I will attach the changes to sharded search service shortly (which is just another endpoint).

This addresses #2705


src/main/java/io/anserini/index/ShardInfo.java
```diff
@@ -0,0 +1,242 @@
...
+    this.queryGenerator = queryGenerator;
+    this.invertedIndex = invertedIndex;
+    this.shardId = shardId;
+  }
+
+  public enum ShardedCollection {
+    MSMARCO_V21_DOC_SEGMENTED_ARCTIC_EMBED_L_HNSW_INT8(
+      "msmarco-v2.1-doc-segmented.arctic-embed-l.hnsw-int8",
+      new ShardInfo[] {
+        MSMARCO_V21_DOC_SEGMENTED_SHARD00_ARCTIC_EMBED_L_HNSW_INT8,
```
lintool, line 191: Why not leave the existing shards as `IndexInfo`, and a shard is just an array of these?

src/main/java/io/anserini/search/SearchShardedHnswDenseVectors.java
```diff
@@ -0,0 +1,237 @@
...
+  private final List<String> queries = new ArrayList<>();
+  private final ShardInfo[] shards;
+  private final int threadsPerShard;
+
+  /*
+   * Constructor for sharded HNSW dense vector search.
+   * The caller must provide an identifier for the sharded index defined in ShardInfo.
+   * This identifier is used to retrieve the shards from ShardInfo.
+   */
+  public SearchShardedHnswDenseVectors(Args args) throws IOException {
```
lintool, line 89: What if we just re-packaged the params and dispatched to `SearchHnswDenseVectors` in parallel? Wouldn't this reduce code duplication?

lintool (changes requested):
Thoughts?

src/main/java/io/anserini/index/ShardInfo.java
vincent-4: Yeah, since if we have it use IndexInfo, it spares me from duplicating– makes sense 👍

vincent-4:
Output: something like this
https://gist.github.com/vincent-4/da59291ba394b00f6d724c11b9593fe8
Could be nicer, but I don't really see a nice way to stop each instance from outputting

lintool:
@vincent-4 ugly output is fine!

vincent-4:
This works now:
I'll clean up and send update tommorow
https://gist.github.com/vincent-4/956974cacfe390b0bc4000776249966a

src/main/java/io/anserini/search/SearchShardedHnswDenseVectors.java
```diff
@@ -0,0 +1,173 @@
...
+      }
+
+      return;
+    }
+    
+    try (SearchShardedHnswDenseVectors<?> searcher = new SearchShardedHnswDenseVectors<>(searchArgs)) {
+      searcher.run();
+    } catch (RuntimeException e) {
+      System.err.printf("Error: %s\n", e.getMessage());
+    }
```
lintool, line 171: Where are you assembling the results from each shard back together into a single file?

vincent-4: Whoops I left that out in the diff, I'll reattach

src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -351,7 +351,7 @@ public enum IndexInfo {
           "https://rgw.cs.uwaterloo.ca/pyserini/indexes/lucene/lucene-hnsw-int8.msmarco-v2.1-doc-segmented-shard09.arctic-embed-l.20250114.4884f5.tar.gz" },
       "5e580bb7eb9ee2bb6bfa492b3430c17d",
       IndexType.DENSE_HNSW,
-      "ArcticEmbedLEncoder.classEncoder.class",
+      "ArcticEmbedLEncoder.class",
```
lintool, line 354: here and elsewhere - it'd be a better idea to have the class itself, rather than a String of the class.

not really part of this PR, so you can save for later... or include it here.

vincent-4:
https://github.com/castorini/anserini/blob/00e29bbbf9c830bd5644866da6fec49baab35ee2/src/main/java/io/anserini/index/IndexInfo.java#L590C1-L590C49
https://github.com/castorini/anserini/blob/00e29bbbf9c830bd5644866da6fec49baab35ee2/src/main/java/io/anserini/search/HnswDenseSearcher.java#L123

Right now, these two conflict:
We can either have the full class name, but it requires changing the way `HnswDenseSearcher` is configured. So then, we have to change some tests too. Example: Something like `BgeBaseEn15.class` will work with the current usage, since it'll have Encoder appended to it, but something like `ArcticEmbedLEncoder.class` won't because it'll be doubly appended :(

When I named it previously, I didn't really notice how dense searcher used the class names, so kiinda a mix of both (which will be fixed depending on the solution we go for this). This would just involve removing the bit with Encoder from `HnswDenseSearcher`.

I think it'd be easy to not touch any tests (hopefully hnsw dense searcher usage is only limited to its test? But I'm not sure)– but since we're re-organizing anyways maybe we could go for it? LMK what you think @lintool

lintool:
I see. Yes, let's circle back to refactor this later. Can you file an issue against yourself to track?

vincent-4:
On tests for this... how to test w/o having multiple shards? Even one is 50GB and clearly we've run out of space above.
Maybe two or so shards only? @lintool

lintool:
@vincent-4 re: tests - create small test sharded indexes (e.g., 10 docs) that can be directly checked into the repo.

src/test/java/io/anserini/server/SearchServiceTest.java
```diff
@@ -116,4 +118,50 @@ public void testInvalidIndex() {
...
+  @Test
+  public void testShardedSearch() throws Exception {
+    String identifier = "msmarco-v2.1-doc-segmented.arctic-embed-l.hnsw-int8";
+    ShardInfo shardGroup = ShardInfo.fromIdentifier(identifier);
+
+    List<Map<String, Object>> results = SearchService.searchSharded(
+        shardGroup, "test query", 10, null, null, null,
+        index -> new SearchService(index));
+
+    assertNotNull(String.format("Expected non-null results for sharded search with identifier '%s'", identifier),
```
lintool, line 131: This and below... I'm fine with longish lines.
People have large monitors these days...

src/test/java/io/anserini/search/SearchShardedHnswDenseVectorsTest.java
```diff
@@ -0,0 +1,248 @@
...
+        "-hits", "250",
+        "-threads", "32",
+        "-selectMaxPassage",
+        "-selectMaxPassage.hits", "5"
+    };
+
+    SearchShardedHnswDenseVectors.main(searchArgs);
+
+    // TODO: Update with actual expected results
+    TestUtils.checkRunFileApproximate(runfile, new String[] {
```
lintool, line 242: Add some basic checks please?

lintool: This and above?

src/main/java/io/anserini/server/SearchService.java
```diff
@@ -233,6 +238,21 @@ private IndexInitializationResult initializeIndex(String prebuiltIndex) {
...
 
+  static List<Map<String, Object>> searchSharded(ShardInfo shardGroup, String query, int hits,
+    Integer efSearch, String encoder, String queryGenerator, Function<String, SearchService> serviceProvider) {
+
+    return Arrays.stream(shardGroup.getShards())
+      .parallel()
+      .map(shard -> {
+        SearchService service = serviceProvider.apply(shard.indexName);
+        return service.search(query, hits, efSearch, encoder, queryGenerator);
+      })
```
lintool, line 249: nit, but I think I like `})` on the previous line better so the method class align.

src/main/java/io/anserini/server/SearchService.java
```diff
@@ -58,8 +62,7 @@ public List<Map<String, Object>> search(String query, int hits) {
     return search(query, hits, null, null, null);
   }
 
-  public List<Map<String, Object>> search(String query, int hits,
-      Integer efSearch, String encoder, String queryGenerator) {
+  public List<Map<String, Object>> search(String query, int hits,Integer efSearch, String encoder, String queryGenerator) {
```
lintool, line 65: nit, space after hits.

lintool (changes requested):

lintool:
@vincent-4 please fix conflicts, and give me the updated command to run to test this feature?

src/main/java/io/anserini/index/ShardInfo.java
```diff
@@ -0,0 +1,56 @@
+package io.anserini.index;
```
lintool, line 1: Add our standard license boilerplate, etc.?

vincent-4:
It's doing the weird thing with the downstream failures:
If I remove the tests added here, it passes (as expected).
But if I make modifications downstream, none of can stop the _kill_ signal ?
I'll take another pass tomorrow morning...

lintool:
ok noted

lintool (approved):

vincent-4:
Fingers crossed on CI

PR #2714, opened 2025-02-09
Swap out parquet files: double -> float

lintool:
Follow up to #2711 but for BEIR.


b8zhong (approved):

lintool:
Thanks @b8zhong - running tests right now. Will merge when everything passes.

PR #2724, opened 2025-02-16
Move regression order to avoid conflicts

lintool:
Before, running the cached and onnx versions of the regressions was causing conflicts because they were both trying to write to the same index location. Instead, spread them out.

Also, add docs for lexlsh/fw regressions.


b8zhong (approved):

vincent-4 (approved):
Lg

lintool:
I'm going to override and merge, the code coverage check isn't correct since I didn't modify any actual code.

PR #2735, opened 2025-02-22
Fusion results on BeIR datasets

lilyjge:
Added fusion results in YAML format for the following BeIR datasets:
- trec-covid
- bioasq
- nfcorpus
- nq
- fiqa
- signal1m
- trec-news
- arguana
- webis-touche2020
- dbpedia-entity
- scidocs
- scifact

Added functions to run ranx fusion after to compare results and time. 

Had to remove integer parsing for query IDs when sorting the run data because query IDs are not always integers and it dies without any very useful error messages.


lintool (approved):

lintool:
lgtm. I'll merge when checks complete.

PR #2736, opened 2025-02-23
Adjust IndexInfo naming to align with CLI arguments

vincent-4:
I tried various refactors, etc. which all kinda grew too much in complexity.

Using the direct class names would be nice (which I tried),
but unfortunately it would break things in the following ways.

- If we only fix the naming, then we're done.
- If we change IndexInfo to use direct class names, {Class}.class, then we'd have to change how it is used at various places. Since these entries are the fallback if no encoder is specified (in the API), it would mean we have diverging ways of representing this information vs. command-line arguments. 
- If we support both, well, it would get messy. There would be different behaviour from the CLI and the fallback values overall, having IndexInfo match what the searchers and their CLI arguments (e,g, HnswDenseSearcher) expect is the least confusing option.

In the end, it's easiest to just fix some broken entries, which makes everything work nicely, we don't really benefit from changing it significantly at the moment.

### DIFF
--- a/src/main/java/io/anserini/index/IndexInfo.java
+++ b/src/main/java/io/anserini/index/IndexInfo.java
@@ -22,9 +22,6 @@ import io.anserini.search.query.VectorQueryGenerator;
 import io.anserini.encoder.dense.BgeBaseEn15Encoder;
 import io.anserini.search.query.InvertedDenseVectorQueryGenerator;
 
-
-
-
 public enum IndexInfo {
   CACM("cacm",
       "Lucene index of the CACM corpus.",
--- a/src/main/java/io/anserini/index/IndexInfo.java
+++ b/src/main/java/io/anserini/index/IndexInfo.java
@@ -56,7 +56,7 @@ public enum IndexInfo {
           "https://rgw.cs.uwaterloo.ca/pyserini/indexes/lucene/lucene-inverted.msmarco-v1-passage.splade-pp-ed.20230524.a59610.tar.gz" },
       "2c008fc36131e27966a72292932358e6",
       IndexType.SPLADE_PP_ED,
-      "SpladePlusPlusEnsembleDistil",
+      "SpladePlusPlusEnsembleDistilEncoder",
       "InvertedDenseVectorQueryGenerator",
       "msmarco-v1-passage"),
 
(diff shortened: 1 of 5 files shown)


lintool:
How about as a compromise, we at least use the full name of the class, e.g., "SpladePlusPlusEnsembleDistilEncoder", instead of just "SpladePlusPlusEnsembleDistil"?

In the CLI, support both - with and without the "Encoder" suffix, so we don't break existing code? (i.e., add a test case)

lintool:
@vincent-4 in `SearchHnswDenseVectorsTest` and elsewhere, can you add a test case to confirm that both XEncoder and X without encoder work?

vincent-4:
Yeah, doing rn

lintool:
@vincent-4 CI failed... any ideas?

vincent-4:
> @vincent-4 CI failed... any ideas?

https://stackoverflow.com/questions/77182923/apparently-failed-github-actions-workflow-step-shows-a-hexagon-with-exclamation
I was looking at this, I think it's a resource issue? It's giving error code 143 on the EmbedL inference (which requires the ONNX model, right)?.. So maybe we're pushing over the (free) runner resources?

lintool:
But CI appears to have been triggered for this PR, which was after yours: https://github.com/castorini/anserini/pull/2737

I kicked it again... let's see.

vincent-4:
I blame @b8zhong  for the latest failure. I'll run again

src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -16,6 +16,15 @@
 
 package io.anserini.index;
 
+import io.anserini.encoder.sparse.SpladePlusPlusEnsembleDistilEncoder;
+import io.anserini.encoder.dense.ArcticEmbedLEncoder;
+import io.anserini.search.query.VectorQueryGenerator;
+import io.anserini.encoder.dense.BgeBaseEn15Encoder;
+import io.anserini.search.query.InvertedDenseVectorQueryGenerator;
+
+
```
lintool, line 25: too many spaces here...

lintool (approved):
👍 from me... merge when you've sorted out the CI issues...

vincent-4:
We still have the same issue.
The only thing is added these two new tests but it shouldn't overwhelm runner resources? I'm gonna investigate

vincent-4:
For future reference, some reference / descriptions below about some issues encounterd while trying to make this PR work. (If there ends up being issues related to this).
```
We add support for the below in this change, but also acknowledge a strange ? failure happening downstream as a result of some attempted changes, which were ditched.

With this PR, two usages are supported: -encoder CosDprDistil OR -encoder CosDprDistilEncoder work. Previously it was the former usage.
However, SearchHnswDenseVectorsTest is missing a test for this behaviour, because it causes a downstream failure at ArcticEmbedLEncoder. Currently, we don't know why this happens, commits / CI in the original PR show various experiments with this.
```

PR #2740, opened 2025-02-23
Refactor tests

lintool:
Quick try to see if I can fix #2736


lintool:
@vincent-4 this appears to be the affect test case... any ideas?

vincent-4:
If I'm reading it right removing `testLength` fixes this? It's not really _needed_, so we're safe to remove, I think

lintool:
> If I'm reading it right removing `testLength` fixes this? It's not really _needed_, so we're safe to remove, I think

Yea, but any idea why that might be the issue?

lintool:
Moved the critical part:

```
assertEquals(1024, embeddings.length);
```

From the first 1st test case to the 2nd test case.

vincent-4 (approved):
Lg!

PR #2745, opened 2025-03-03
Completed fusion results for BeIR datasets with documentation

lilyjge:
Added fusion results in YAML format for the following BeIR datasets:
- climate-fever
- cqadupstack
- fever
- hotpotqa
- quora
and ms-marco as well.

Renamed output files and dependencies for all fusion YAML files for consistency with actual model used (originally named as ONNX, but actually used cached queries).

Added documentation recording fusion results and how to replicate runs and evaluations. The commands were tested and works with all the values matching, with the exception of robust04, whose YAML values are slightly different from what the commands produce. The result of the commands are recorded in the documentation, while the values from the original YAML were left unchanged.

Added logging in fusion to print out the number of queries merged for every hundred queries.

Added Python script to generate YAML files for fusion.


docs/experiments-beir-fusion.md
```diff
@@ -0,0 +1,115 @@
...
+
+For all experiments recorded here, the values k = 1000, depth = 1000, rrf_k = 60, and alpha = 0.5 were used.
+
+The runs of two models were fused: flat BM25 and flat bge-base-en-v1.5 with cached queries.
+
+Since there were only two runs fused, the average and interpolation methods produced the same results.
+
+The table below reports the effectiveness of the methods:
+
+| Corpus                     | rrf:nDCG@10 | rrf:R@100 | rrf:R@1000 | avg:nDCG@10 | avg:R@100 | avg:R@1000 | interp:nDCG@10 | interp:R@100 | interp:R@1000 |
```
lintool, line 22: It makes sense to separate into different tables for each metric? so one for nDCG@10, on for R@100, one for R@1000.

lilyjge: Done

lintool (approved):

PR #2768, opened 2025-04-04
Minor cleanup

lintool:
+ Added more sample commands to README
+ Added more MS MARCO bindings in Topics

### DIFF
--- a/src/main/java/io/anserini/search/topicreader/Topics.java
+++ b/src/main/java/io/anserini/search/topicreader/Topics.java
@@ -549,14 +549,14 @@ public enum Topics {
   NTCIR_8_EN_EVAL(TsvStringTopicReader.class,   "topics.ntcir8en.eval.txt");
 
   public final String path;
-  public final Class<? extends TopicReader> readerClass;
+  public final Class<? extends TopicReader<?>> readerClass;
 
-  Topics(Class<? extends TopicReader> c, String path) {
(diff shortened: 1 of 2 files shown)


vincent-4 (approved):

src/main/java/io/anserini/search/topicreader/Topics.java
```diff
@@ -575,6 +575,13 @@ private static Map<String, Topics> generateSymbolDictionary() {
     m.put("msmarco-passage-dev-bge-base-en-v1.5", MSMARCO_PASSAGE_DEV_SUBSET_BGE_BASE_EN_15);
     m.put("msmarco-passage-dev-cohere-embed-english-v3.0", MSMARCO_PASSAGE_DEV_SUBSET_COHERE_EMBED_ENGLISH_30);
 
+    m.put("msmarco-passage-dev.splade-pp-ed", MSMARCO_PASSAGE_DEV_SUBSET_SPLADE_PP_ED);
```
vincent-4, line 578: The separators (-dev, commas, dots) diverging looked a bit funny at first (eg vs. the command that was changed) but it seems to make sense; just an observation

lintool: yes, trying to harmonize across both anserini and pyserini - my strategy has been to over-generate so anything "reasonable" will work.

PR #2769, opened 2025-04-05
Remove redundant searcher.close() call in run() method, fixes #2759

### LINKED ISSUE #2759: SearchShardedHnswDenseVectors closes underlying searchers twice
We close the underlying searchers twice.

+ Here, once: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SearchShardedHnswDenseVectors.java#L142 in `run()`
+ Then again: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SearchShardedHnswDenseVectors.java#L112 in `close()`

Which makes more sense? Not sure... but let's discuss @vincent-4 ?

vincent-4:


src/main/java/io/anserini/search/SearchShardedHnswDenseVectors.java
```diff
@@ -139,9 +139,8 @@ public void run() {
       try {
         searcher.args.output = shardOutputPath;
         searcher.run();
-        searcher.close();
         LOG.info("Closed searcher for shard {}", i);
```
lintool, line 142: You should remove the LOG statement because it's no longer accurate.

src/main/java/io/anserini/search/SearchShardedHnswDenseVectors.java
```diff
@@ -139,9 +139,8 @@ public void run() {
...
-      } catch (IOException e) {
+      } catch (RuntimeException e) {
```
lintool, line 143: I don't think you need the catch block any more.

lintool (changes requested):

lintool (approved):

PR #2770, opened 2025-04-05
Clean up msmarco-v2-doc-segmented.unicoil-0shot

lintool:


MXueguang (approved):

lintool:
More details on this PR.

Back from a while ago, we had the following (and related regressions for DL):

```
msmarco-v2-doc-segmented.unicoil-0shot.cached.yaml
msmarco-v2-doc-segmented.unicoil-0shot-v2.cached.yaml
msmarco-v2-doc-segmented.unicoil-noexp-0shot.cached.yaml
msmarco-v2-doc-segmented.unicoil-noexp-0shot-v2.cached.yaml
```

This was mostly a bug, because for v1, we only encoded the segment, which we later fixed with v2, encoding both title and segment. Until now, we've kept regressions for both v1 and v2.

This PR removes the v1 variants, since there's no need to keep around anymore.

Issue #2620, opened 2024-10-09
Refactor fusion implementation to use ScoredDocs


lintool:
The current implementation of fusion is built around `TrecRun`, e.g., https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/fusion/TrecRunFuser.java

But we already have `ScoredDocs`:
https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/ScoredDocs.java

Should we refactor our implementation to use `ScoredDocs`?

Potential cons:
+ faster?
+ less code duplication

lilyjge:
Working on it!

lilyjge:
Running times for end to end fusion after ScoredDocs refactoring with run_fusion_regression.py. Ranx cache increases speed quite a bit, so for fairness I cleared its cache before each run. Times measured in seconds. 

| Corpus                  | Anserini | Ranx (no cache) |
| ----------------------- | -------- | --------------- |
| trec-covid              | 8.41     | 39.22           |
| bioasq                  | 27.66    | 44.14           |
| nfcorpus                | 16.95    | 41.11           |
| nq                      | 146.01   | 79.36           |
| hotpotqa                | 307.29   | 115.63          |
| fiqa                    | 31.15    | 44.33           |
| signal1m                | 10.83    | 41.21           |
| trec-news               | 8.71     | 41.79           |
| robust04                | 17.32    | 43.48           |
| arguana                 | 75.83    | 77.96           |
| webis-touche2020        | 8.12     | 41.77           |
| cqadupstack-android     | 32.50    | 44.89           |
| cqadupstack-english     | 67.82    | 53.68           |
| cqadupstack-gaming      | 68.07    | 53.83           |
| cqadupstack-gis         | 42.33    | 46.31           |
| cqadupstack-mathematica | 35.56    | 45.95           |
| cqadupstack-physics     | 47.28    | 47.58           |
| cqadupstack-programmers | 41.47    | 46.80           |
| cqadupstack-stats       | 31.84    | 44.48           |
| cqadupstack-tex         | 121.55   | 69.58           |
| cqadupstack-unix        | 49.15    | 50.12           |
| cqadupstack-webmasters  | 25.69    | 43.47           |
| cqadupstack-wordpress   | 27.55    | 43.62           |
| quora                   | 370.70   | 135.31          |
| dbpedia-entity          | 25.66    | 216.71          |
| scidocs                 | 58.29    | 64.45           |
[...]

PR #2774, opened 2025-04-06
Refactor fusion implementation to use ScoredDocs

lilyjge:
Addresses this issue (#2620 ). New running times recorded in issue.

Includes updated tests for refactored implementation. 

Updates documentation including running instructions and slightly different scores due to using float instead of double (also updated in YAMLs).

Renames fusion files and directory containing YAMLs for fusion regression.


docs/experiments-beir-fusion.md
```diff
@@ -26,15 +26,15 @@ The table below reports the effectiveness of the methods with the nDCG@10 metric
 |---------------------------|-------:|--------:|--------------:|----------:|-------:|-------:|
 | `trec-covid`              | 0.8041 |  0.6567 |     0.6567    |   0.7956  | 0.5947 | 0.7815 |
 | `bioasq`                  | 0.5278 |  0.5308 |     0.5308    |   0.5428  | 0.5225 | 0.4148 |
-| `nfcorpus`                | 0.3725 |  0.3416 |     0.3416    |   0.3657  | 0.3218 | 0.3735 |
+| `nfcorpus`                | 0.3725 |  0.3415 |     0.3415    |   0.3657  | 0.3218 | 0.3735 |
 | `nq`                      | 0.4831 |  0.3241 |     0.3241    |   0.5183  | 0.3055 | 0.5415 |
 | `hotpotqa`                | 0.7389 |  0.6497 |     0.6497    |   0.7658  | 0.6330 | 0.7259 |
 | `fiqa`                    | 0.3671 |  0.2470 |     0.2470    |   0.3942  | 0.2361 | 0.4065 |
-| `signal1m`                | 0.3533 |  0.3463 |     0.3463    |   0.3629  | 0.3304 | 0.2886 |
```
lintool, line 33: Any idea why these numbers changed?

src/main/java/io/anserini/fusion/ScoredDocsFuser.java
```diff
@@ -0,0 +1,270 @@
...
+        // Parse RANK as integer
+        int rankInt = Integer.parseInt(data[3]);
+        rank.add(rankInt);
+
+        // Parse SCORE as float
+        float scoreFloat = Float.parseFloat(data[4]);
+        scores.add(scoreFloat);
+      }
+
+      scoredDocs.lucene_documents = lucene_documents.toArray(new Document[0]);
```
lintool, line 77: We don't use this, right? I think we can just set to `null`?

src/main/java/io/anserini/fusion/ScoredDocsFuser.java
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  private static void normalizeScores(ScoredDocs scoredDocs) {
+    Map<String, List<Integer>> indicesForTopics = new HashMap<String, List<Integer>>(); // topic, list of indices for that topic
+    int length = scoredDocs.lucene_documents.length;
+    for(int i = 0; i < length; i++){
+        indicesForTopics.computeIfAbsent(scoredDocs.lucene_documents[i].get(TOPIC), k -> new ArrayList<>()).add(i);
+    }
+
+    for(List<Integer> topicIndices : indicesForTopics.values()){
```
lintool, line 131: nit, two space intent, space after `for`.

src/main/java/io/anserini/fusion/ScoredDocsFuser.java
```diff
@@ -0,0 +1,270 @@
...
+  /**
+   * Rescored given ScoredDocs using the specified method.
+   *
+   * @param method  Rescore method to be applied (e.g., RRF, SCALE, NORMALIZE).
+   * @param rrfK    Parameter k needed for reciprocal rank fusion.
+   * @param scale   Scaling factor needed for rescoring by scaling.
+   * @param scoredDocs ScoredDocs object to be rescored.
+   * @throws UnsupportedOperationException If an unsupported rescore method is provided.
+   */
+  public static void rescore(RescoreMethod method, int rrfK, double scale, ScoredDocs scoredDocs) {
```
lintool, line 99: Might need to think about how this method works, because some of the variables are applicable to only some of the methods... e.g., rrfK is only applicable to RRF.

src/main/java/io/anserini/fusion/ScoredDocsFuser.java
```diff
@@ -0,0 +1,270 @@
...
+
+    // for every topic, produce a map of docid to score, num of accumulated
+    HashMap<String, HashMap<String, AbstractMap.SimpleEntry<Float, Integer>>> docScores = new HashMap<>();
+    for(ScoredDocs run : runs){
+      for(int i = 0; i < run.lucene_documents.length; i++){
+        String query = run.lucene_documents[i].get(TOPIC);
+        String docid = run.docids[i];
+        Float score = run.scores[i];
+        docScores.computeIfAbsent(query, key -> new HashMap<>())
+          .merge(docid, new AbstractMap.SimpleEntry<>(score, 1),
```
lintool, line 172: I suspect it's parts like this that are slow... manipulating `HashMap`s creates lots of objects.

Nothing actionable, just noting for a future iteration.

src/main/java/io/anserini/fusion/ScoredDocsFuser.java
```diff
@@ -0,0 +1,270 @@
...
+   * @throws IllegalStateException If the ScoredDocs is empty.
+   */
+  public static void saveToTxt(Path outputPath, String tag, ScoredDocs run) throws IOException {
+    if (run.lucene_documents == null || run.lucene_documents.length == 0) {
+      throw new IllegalStateException("Nothing to save. ScoredDocs is empty");
+    }
+
+    ScoredDocsFuser.sortScoredDocs(run);
+    try (BufferedWriter writer = Files.newBufferedWriter(outputPath)) {
+        for (int i = 0; i < run.lucene_documents.length; i++) {
```
lintool, line 264: two space indent.

lintool (changes requested):
Based on a quick look. Will provide more detailed feedback later.

docs/experiments-beir-fusion.md
lilyjge: ScoredDocs uses float for scores while TrecRun used double, so I suspect loss in accuracy

src/main/java/io/anserini/fusion/ScoredDocsFuser.java
lilyjge: I'm using it to hold the topic/query ids.

src/main/java/io/anserini/fusion/ScoredDocsFuser.java
```diff
@@ -0,0 +1,271 @@
...
+      case RRF -> ScoredDocsFuser.rescoreRRF(rrfK, scoredDocs);
+      case SCALE -> ScoredDocsFuser.rescoreScale(scale, scoredDocs);
+      case NORMALIZE -> ScoredDocsFuser.normalizeScores(scoredDocs);
+      default -> throw new UnsupportedOperationException("Unknown rescore method: " + method);
+    }
+  }
+
+  private static void rescoreRRF(int rrfK, ScoredDocs scoredDocs) {
+    int length = scoredDocs.lucene_documents.length;
+    for(int i = 0; i < length; i++){
```
lintool, line 110: nit, spaces ^
```suggestion
    for (int i = 0; i < length; i++) {
```

change for all instances please.

lintool:
Hi @lilyjge looking at your code, I probably want refactoring of how the class is called, but we can deal with that later.

lintool:
Hi @lilyjge I have some suggestions on how we should rework the interface, but we can circle back on that later.

Before we merge, can you please update: https://github.com/castorini/anserini/issues/2620 ?

I'd like to have some confidence that this refactoring _actually_ made things faster... so we can close that issue.

lilyjge:
A few of the datasets aren't done yet, but I'll add what I have right now and update as they finish.

lintool (approved):

PR #2776, opened 2025-04-08
Shard Merge tool

JMMackenzie:
This adds a tool for merging an index with multiple shards into an index with a single shard.

Help is required to test (write tests) and verify the tool.

### DIFF
--- a/src/main/java/io/anserini/index/MergeShards.java
+++ b/src/main/java/io/anserini/index/MergeShards.java
@@ -1,0 +1,35 @@
+package io.anserini.index;
+
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.store.FSDirectory;
+import org.apache.lucene.analysis.standard.StandardAnalyzer;
+
+import java.nio.file.Paths;
+
+public class MergeShards {
+
+    public static void main(String[] args) throws Exception {
+        if (args.length < 2) {
+            System.err.println("Usage: MergeShards <output_dir> <shard_dir1> [<shard_dir2> ...]");
+            System.exit(1);
+        }
+
+        String outputDir = args[0];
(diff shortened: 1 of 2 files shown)


vincent-4:
@JMMackenzie I will take a look and test to get this merged soon. @b8zhong has added some tests and I will correct, update as necessary. Thanks for this again 👍 
What indexes are you using with it? It might also help to have some reference.

JMMackenzie:
I usually use Anserini to build CIFF indexes that I import to other tools like PISA, but I almost always forget to use `-optimize` when I index. So this tool fixes that.

You could try it on MSMARCO-v1 and MSMARCO-v2 (passages) if you wanted something large, not sure how long it would take for v2 though.

Thanks for your help!

src/main/java/io/anserini/index/MergeShards.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.store.FSDirectory;
+import org.apache.lucene.analysis.standard.StandardAnalyzer;
+
+import java.nio.file.Paths;
+
+public class MergeShards {
+
```
lintool, line 27: two-space indent to match house style, please.

JMMackenzie: Resolved -- please double check!

lintool (approved):

lintool:
Actually, I approved before noticing broken tests. Should probably fix before merging...

JMMackenzie:
I'm not sure why these tests fail - I can't seem to generate any meaningful output. It seems the VM is crashing (I get "The forked VM terminated without properly saying goodbye" from surefire). Someone with more experience may be able to find a solution.

PR #2778, opened 2025-04-19
Update experiments-msmarco-passage2.md

JJGreen0:
Student Linux Environment 

Exceeded the CPU time limit during retrieval run. Got to about 6000 queries processed before being killed.  

Chatgpt told me this: 
You’re hitting a hard CPU‐seconds quota on your student account: the cluster is limiting the sum of CPU time across all threads, not just wall‑clock time. By spinning up 4 threads, your 6000‑query run burned through that quota and got killed. You’re not permanently stuck—you just need to throttle back or split the work. Here are three easy workarounds: 

Reduce parallelism/threads 
 
I tried running with 1 thread and got “CPU time limit exceeded” at 6800 queries 

I tried running it again with literally no changes and managed to just finish the 6900 queries.  

Additionally, I’m not sure why we named our run outputs for the last guide to end with .trec but for this one they end with .txt (despite the qrels still being .trec). 

Otherwise, everything worked.


lintool (approved):
haha, thanks for your valiant effort!

PR #2783, opened 2025-04-28
Fix small typo on fusion experiments

carlos-lassance:
While I was checking fusion numbers I found a small typo that said it was going to run normalize fusion, but actually performed interpolation.


lintool:
hi @carlos-lassance this code I believe was added _after_ the SIGIR paper, but if it repros the paper figures, all the better. cc @lilyjge

lilyjge:
It is indeed a typo. Thanks for catching it!

lintool (approved):

PR #2801, opened 2025-05-23
Fix PrebuiltIndexTest by closing connections

### LINKED ISSUE #2800: PrebuiltIndexTest server rate limit
PrebuiltIndexTest#testUrls is failing because it tries to open and test the connection for every prebuilt index's URLs but the server is sending us 503 after about 65 requests. We have 169 prebuilt indexes and possibly more URLs. 

I've tried adding a 1.5 second sleep between each request, sleeping for a minute if it encounters a 503 and trying again, and both at the same time. Still getting the same thing.

lilyjge:
Fixes #2800 by closing the URL connection after each URL is tested so it doesn't reach the rate limit.


lintool:
@lilyjge +1 thanks for taking care of this. Merge after tests pass please.

Issue #2793, opened 2025-05-13
Upload ONNX models to hgf


lintool:
Currently, all our ONNX models are stored at UWaterloo. What if we stored on hgf so we can keep track of downloads?

Maybe we can start with #2792

lintool:
While we're at it, maybe just move all ONNX models to hgf? i.e., all models in Anserini: https://github.com/castorini/anserini/tree/master/src/main/java/io/anserini/encoder

lintool:
Closed by #2817

Issue #2792, opened 2025-05-13
Add SPLADEv3 impl in ONNX


lintool:
Convert SPLADEv3 to ONNX + Java bindings

Follow this guide here:
https://github.com/castorini/anserini/blob/master/docs/onnx-conversion.md

First reproduce existing SPLADE++ ED/SD model conversion to confirm that you've done it correctly, then convert SPLADEv3.

lilyjge:
Converted and uploaded to HuggingFace here: https://huggingface.co/castorini/splade-v3-onnx.

PR #2810, opened 2025-05-27
Add SpladeV3 ONNX implementation 

lilyjge:
Addresses #2792. 

While the guide mentioned in the issue does convert a given model to ONNX and can be reproduced, for SPLADE (I tried SPLADE++ ED and SPLADE v3), the converted models don't fit with the existing system i.e. if you convert SPLADE++ ED with the script in the guide, the converted model fails the tests. This is because the converted model doesn't output what the tests expect, and some additional logic is needed like the [encode function here](https://github.com/castorini/pyserini/blob/master/pyserini/encode/_splade.py). 
To address this, I 'wrapped' the SPLADE v3 by adding on an extra layer with that logic before exporting it to ONNX. This allows the model to fit with the current Anserini system for SPLADE encoders. I think it's fine to do this as opposed to implementing the output conversion logic in Anserini because the SPLADE++ behaves similarly and the ONNX models probably had to do something similar in order to get the current effect. I've included the script I used for future reference. 

With the resulting SPLADE v3 ONNX model sharing the same logic as the SPLADE++ ONNX models, I renamed SpladePlusPlusEncoder to SpladeEncoder so it makes sense for SpladeV3Encoder to inherit from it.

I copied over the test inputs from the SPLADE++ models and encoded with the Pyserini encoder linked above to get the expected outputs. The ONNX model's outputs match exactly except for a single token where the weight is 9 instead of 10, which I attribute to rounding error and is probably negligible. I tested the converted model on nfcorpus and got the same score as [here](https://github.com/castorini/pyserini/blob/master/docs/conceptual-framework3.md).

Addressing #2793, I've uploaded the models to HuggingFace [here](https://huggingface.co/castorini/splade-v3-onnx). HuggingFace counts downloads by the number of times the model's config file is downloaded, so I uploaded a config file and added constructors to pass in a config URL to the base class, so it will be downloaded every time the model is downloaded. The number of downloads did update since I did that, but it's not live and takes a while.


src/main/java/io/anserini/encoder/OnnxEncoder.java
```diff
@@ -46,19 +48,21 @@ public abstract class OnnxEncoder<T> implements AutoCloseable {
...
+  private final String configUrl;
 
   protected final BertFullTokenizer tokenizer;
   protected final DefaultVocabulary vocab;
 
   protected final OrtEnvironment environment;
   protected final OrtSession session;
 
-  public OnnxEncoder(@NotNull String modelName, @NotNull String modelUrl, @NotNull String vocabName, @NotNull String vocabUrl)
+  public OnnxEncoder(@NotNull String modelName, @NotNull String modelUrl, @NotNull String vocabName, @NotNull String vocabUrl, String configUrl)
```
lintool, line 59: Since all other params have annotations, also add `@NotNull`?

src/main/java/io/anserini/encoder/sparse/SparseEncoder.java
```diff
@@ -46,14 +46,21 @@ public static String flatten(Map<String, Integer> intWeights) {
   }
 
   public SparseEncoder(int weightRange, int quantRange,
-                       @NotNull String vocabName, @NotNull String vocabUrl,
-                       @NotNull String modelName, @NotNull String modelUrl)
+                       @NotNull String modelName, @NotNull String modelUrl,
+                       @NotNull String vocabName, @NotNull String vocabUrl, String configUrl)
```
lintool, line 50: same

src/main/java/io/anserini/encoder/sparse/SpladeEncoder.java
```diff
@@ -41,12 +41,18 @@ public abstract class SpladePlusPlusEncoder extends SparseEncoder {
...
 
-  protected SpladePlusPlusEncoder(@NotNull String modelName, @NotNull String modelUrl,
+  protected SpladeEncoder(@NotNull String modelName, @NotNull String modelUrl,
                                   @NotNull String vocabName, @NotNull String vocabUrl)
       throws IOException, OrtException, URISyntaxException {
     super(WEIGHT_RANGE, QUANT_RANGE, modelName, modelUrl, vocabName, vocabUrl);
   }
 
+  protected SpladeEncoder(@NotNull String modelName, @NotNull String modelUrl,
+                                  @NotNull String vocabName, @NotNull String vocabUrl, String configUrl)
```
lintool, line 51: nit, spacing

src/test/java/io/anserini/encoder/sparse/SpladeV3EncoderTest.java
```diff
@@ -0,0 +1,448 @@
...
+import java.util.HashMap;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertTrue;
+
+public class SpladeV3EncoderTest extends BaseSparseEncoderInferenceTest {
+
+  private static final SparseExampleOutputPair[] EXAMPLES = {
+      new SparseExampleOutputPair("what is paula deen's brother",
+          new HashMap<>() {
```
lintool, line 33: but, I think you want `{{` on same line?

Same below...

lintool (approved):
Awesome! Minor nits... please tweak and merge!

src/main/java/io/anserini/encoder/OnnxEncoder.java
lilyjge: Sometimes configUrl is null because not all models are uploaded to HuggingFace with a config yet, so the other Splade models call the constructor under this without a config url, and that constructor calls this one with config url being null. I can probably edit SparseEncoder though.

src/main/java/io/anserini/encoder/sparse/SparseEncoder.java
lilyjge: Similarly here as above, with constructor delegation, configUrl is null for the other SPLADE models.

src/test/java/io/anserini/encoder/sparse/SpladeV3EncoderTest.java
lilyjge: The other splade tests do this too, should I change them as well?

lintool: ah, didn't realize the inconsistency... yes please.

Poking around online, this appear so to be the preferred formatting:

```
List<String> myList = new ArrayList<String>() {{
    add("item1");
    add("item2");
}};
```

Issue #2820, opened 2025-05-31
Storing prebuilt indexes on hgf


lintool:
Currently, uploading prebuilt indexes is bottlenecked on me. It would be nice if we could move the prebuilt indexes on huggingface, just like the ONNX models... we gotta think through how to organize them though?

For example, a separate repo for each collection, each model would be too much?

What about gathering all variants of the same collection together? E.g., `msmarco-v1-passage` is a repo, underneath it we could have:

```
+ lucene-inverted
  - base
  - splade
  - ...
+ lucene-hnsw
  - bge
  - ...
+ lucene-flat
  - bge
  - ...
```

Issue #2816, opened 2025-05-30
Install ONNX regressions for SPLADEv3 (for BEIR)


lintool:
There are lots of places where this will require changes to the current repo, but let's start here:
https://github.com/castorini/anserini/blob/master/docs/fatjar-regressions/fatjar-regressions-v1.0.0.md#beir

clides:
SPLADEv3 comparison with Splade++ and results from SPLADEv3 paper: https://docs.google.com/spreadsheets/d/1d9jLY-f9QYfnCg3RaM3kvw6ExvsDXfGKXgiO9zvA_6k/edit?usp=sharing

lilyjge:
Not sure ```fatjar-regressions-v1.0.0.md``` is the best place to put them right now since SpladeV3 ONNX isn't actually included in v1.0.0?

lintool:
Correct - create a new file called `fatjar-regressions-v1.0.1-SNAPSHOT.md` and put details there? This will be the beginning of the instructions for the next version.

lintool:
hi @lilyjge the next release is going to be v1.1.0 - https://github.com/orgs/castorini/projects/2 so just create `fatjar-regressions-v1.1.0.md` please.

lintool:
Doing a pass on `fatjar-regressions-v1.0.0.md` here #2855

lintool:
#2855 merged, closing here.

PR #2836, opened 2025-06-11
Add SPLADEv3 BeIR prebuilt indexes

lilyjge:
SPLADEv3 BeIR prebuilt indexes have been uploaded to Hugging Face and can be found [here](https://huggingface.co/datasets/castorini/prebuilt-indexes-beir/tree/main/lucene-inverted/splade-v3) along with its README. 
This is the last of Anserini's prebuilt indexes currently and completes #2820; also a step closer to allowing the commands for #2816 to actually run. Still needs topics to be mapped though: https://github.com/castorini/anserini/blob/1e11fc9500ad67f563025d3180b5ba8ac22ce4d7/src/main/java/io/anserini/search/topicreader/Topics.java
Maybe we should update Pyserini prebuilt indexes as well sometime?


lintool (approved):
👍 
🎉

Issue #2852, opened 2025-06-23
Metadata alignment with Pyserini prebuilt indexes


lilyjge:
Pyserini's prebuilt index definitions contain more metadata than Anserini, specifically the following fields: size compressed (bytes), total terms, documents, unique terms, downloaded. Anserini's prebuilt indexes don't have these fields. Given that Pyserini ports over some prebuilt indexes from Anserini directly, this means the indexes ported over are missing these fields compared to the 'native' Pyserini indexes. Alignment, i.e., adding these metadata fields to Anserini, would be nice.

PR #2853, opened 2025-06-23
Add metadata for prebuilt indexes for Pyserini alignment

lilyjge:
Addresses #2852.  This adds all the metadata removed by castorini/pyserini#2159 to Anserinin's prebuilt indexes. However, some Anserini indexes such as MSMARCO v2 shard don't exist in Pyserini, some are new like SPLADEv3 on BEIR, and some were already ported over from Anserini like BEIR BGE, so the metadata doesn't exist in Pyserini. Does it exist anywhere? Specificially: file size (compressed), number of total terms, number of documents, and number of unique terms.


lintool:
@lilyjge thanks for working on this - I think this is good for now, for v1.1.0 - let's circle back and do more reconciliation later.

lintool (approved):

PR #2854, opened 2025-06-24
Fix QueryGenerator typo in prebuilt indexes

lilyjge:
BM25 flat and multifield + SPLADE prebuilt indexes had InvertedDenseVectorQueryGenerator as the QueryGenerator instead of BagOfWordsQueryGenerator, which is a typo.


src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -21,7 +21,7 @@
 import io.anserini.encoder.dense.ArcticEmbedLEncoder;
 import io.anserini.search.query.VectorQueryGenerator;
 import io.anserini.encoder.dense.BgeBaseEn15Encoder;
```
lintool, line 20: are these in alphabetical order?

src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -247,7 +247,7 @@ public enum IndexInfo {
       "aab3f8e9aa0563bd0f875584784a0845",
       IndexType.DENSE_HNSW,
       ArcticEmbedLEncoder.class.getSimpleName(),
-      InvertedDenseVectorQueryGenerator.class.getSimpleName(),
```
lintool, line 250: I think for HNSW indexes the generator is `VectorQueryGenerator`?

Issue #2856, opened 2025-06-24
Implement -outputRerankerRequests for SearchHnswDenseVectors


lintool:
Looking at this:
https://github.com/castorini/anserini/blob/master/docs/fatjar-regressions/fatjar-regressions-v1.0.0.md

For `SearchCollection` we have the option `-outputRerankerRequests` that generates output that `rank_llm` can use.

I think we need this for `SearchHnswDenseVectors` as well?

@lilyjge I think we should implement this for v1.1.0?

lilyjge:
I have a 'working' branch [here](https://github.com/lilyjge/anserini/tree/rerank), but there's some problems/limitations that makes it not very useful:
1. The dense queries in tools only have the vector embeddings, not the query text, so the content of [this](https://github.com/castorini/anserini/blob/master/docs/fatjar-regressions/fatjar-regressions-v1.0.0.md?plain=1#L168) is a vector, which rankllm can't use
2. The parquet files on orca in `/store/collections/beir-v1.0.0/bge-base-en-v1.5/{dataset}.parquet` don't seem to contain the document field or contents, so I didn't implement the option for `ParquetDenseVectorCollection` yet, which would probably be necessary for large corpora
3. I implemented the option for `JsonDenseVectorCollection`. The jsonl versions of the corpora in `/store/collections/beir-v1.0.0/bge-base-en-v1.5/{dataset}` do have the "contents" field, but they don't actually have the real document contents, only placeholders like "contents MED-10"
4. My output follows the same format as [this](https://github.com/castorini/anserini/blob/master/docs/fatjar-regressions/fatjar-regressions-v1.0.0.md?plain=1#L174), that is, it dumps the raw documents in the reranker requests file, which in this case are large vectors essentially useless for reranking that took up 500mb for top 100 retrieval on nfcorpus 
6. Prebuilt indexes would not work with the option because they don't contain the raw Lucene documents

TLDR, putting aside the implementation, in order to make this work, one would need to munge the corpus and queries to contain both the original contents and the vectors, and reindex with `--storeRaw`. While doable, this just doesn't seem very convenient. 

[...]

PR #2859, opened 2025-06-26
Implement -outputRerankerRequests for SearchHnswDenseVectors

lilyjge:
Addresses #2856 with the problems mentioned.


lilyjge:
Instead of putting the option in the searcher, OutputRerankerRequests is a separate module that generates the reranker requests from a run in TREC format, a Lucene index containing the raw documents, and topics. The index can be a local path or prebuilt. If the index is prebuilt, its "inverted index" version mapped in `IndexInfo.java` will be used. Topics can also be a local path or a name mapped in `Topics.java`. If it's a local path, a `TopicReader` can be specified. If it's a name, it will go through regex to try to strip it down to its "base" form e.g. `beir-nfcorpus.bge-base-en-v1.5` -> `beir-nfcorpus`. 

While implementing reading the index, I noticed that the searchers had a chunk of repeated code to get the index path by looking locally or downloading. I needed to use that as well, so I pulled it out into `IndexReaderUtils.java`. Less code duplication.

Added tests. In order to test the actual request generation, I needed a prebuilt index with raw documents, so I made one off of `src/test/resources/sample_docs/beir/collection1/`. 

Example usage: 
```bash
bin/run.sh io.anserini.rerank.OutputRerankerRequests \
-index indexes/lucene-inverted.beir-v1.0.0-nfcorpus.flat \
-run runs/run.beir-v1.0.0-nfcorpus.bge-base-en-v1.5.bge-hnsw-cached.topics.beir-v1.0.0-nfcorpus.test.bge-base-en-v1.5.jsonl.txt \
-topics beir-nfcorpus -output runs/rerank_nfcorpus.jsonl
```

src/main/java/io/anserini/rerank/OutputRerankerRequests.java
```diff
@@ -0,0 +1,247 @@
+/*
```
lintool: how about `GenerateRerankerRequests`?

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -168,14 +172,25 @@ public long getTotalTF() {
...
    * @param path index path
    * @return index reader
    * @throws IOException if any errors are encountered
    */
   public static IndexReader getReader(String path) throws IOException {
-    Directory dir = FSDirectory.open(Paths.get(path));
+    return getReader(Paths.get(path));
+  }
+
+    /**
```
lintool:
```suggestion
  /**
```

src/main/java/io/anserini/rerank/OutputRerankerRequests.java
```diff
@@ -0,0 +1,247 @@
...
+      throw new IllegalArgumentException("Query ID not found in the list of topics: " + qid);
+    }
+    Map<String, Object> queryMap = new LinkedHashMap<>();
+    queryMap.put("query", new LinkedHashMap<>(Map.of("qid", qid, "text", queries.get(index))));
+    queryMap.put("candidates", candidates);
+    output.println(mapper.writeValueAsString(queryMap));
+    candidates = new ArrayList<>();
+  }
+
+  public void fromScoredDoc(K qid, ScoredDoc[] results) throws JsonProcessingException, IOException {
```
lintool: can you rename these methods? `fromX` are usually static methods that instantiate an instance of the object in question.

src/main/java/io/anserini/index/IndexReaderUtils.java
```diff
@@ -837,6 +852,30 @@ public static Map<String, String>  getFieldInfoDescription(IndexReader reader) {
     return description;
   }
 
+  public static Path getIndex(String index) {
```
lintool, line 855: quick explanation of why we need this? it would seem like we should already have this functionality?

src/main/java/io/anserini/rerank/OutputRerankerRequests.java
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+
+    try {
+      topics.forEach((qid, topic) -> {
+        String query = topic.get(TOPIC_FIELD);
+        assert query != null;
+        qids.add(qid);
+        queries.add(query);
+    });
```
lintool:
```suggestion
      });
```

src/main/java/io/anserini/rerank/OutputRerankerRequests.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+  }
+
+  public IndexReader getIndexReader(String index) {
+    Path indexPath = Path.of(index);
+    if (!Files.exists(indexPath)) { // get inverted index if index doesn't exist locally
+      IndexInfo currentIndex = IndexInfo.get(index);
+      indexPath = IndexReaderUtils.getIndex(currentIndex.invertedIndex);
+    }
+    System.out.println("Generating outputRerankerRequests with raw documents from index: " + indexPath.toString());
```
lintool: change this and other instances to logging instead?

src/main/java/io/anserini/search/FlatDenseSearcher.java
```diff
@@ -79,26 +76,7 @@ public static class Args extends BaseSearchArgs {
...
-      } catch (IOException e) {
-        throw new RuntimeException("MD5 checksum does not match!");
-      } catch (Exception e) {
-        throw new IllegalArgumentException(String.format("\"%s\" does not appear to be a valid index.", args.index));
-      }
-    } else {
-      // if it exists locally, we use it
-      indexPath = Paths.get(args.index);
-    }
+    Path indexPath = IndexReaderUtils.getIndex(args.index);
```
lintool, line 79: Ah, I get it now, nm.

src/main/java/io/anserini/search/topicreader/Topics.java
```diff
@@ -874,7 +874,7 @@ private static Map<String, Topics> generateSymbolDictionary() {
     return m;
   }
 
-  static public Topics getByName(String name) {
+  public static Topics getByName(String name) {
```
lintool, line 877: can you look elsewhere to see if we have static or public first? just be consistent.

src/main/java/io/anserini/search/topicreader/Topics.java
```diff
@@ -885,4 +885,10 @@ static public Topics getByName(String name) {
       return null;
     }
   }
+
+  public static Topics getBaseTopics(String name) {
```
lintool, line 889: same

src/main/java/io/anserini/search/topicreader/Topics.java
```diff
@@ -885,4 +885,10 @@ static public Topics getByName(String name) {
...
+    name = name.replaceFirst("^topics\\.", "");
+    String regex = "^(.*?)(?:[.-](bge|cohere|splade|unicoil|cosdpr|txt|tsv|v\\d+|v\\d+\\.\\d+)).*";
```
lintool: quick comment what the regex is doing.

src/main/java/io/anserini/search/topicreader/Topics.java
lilyjge: Almost everywhere is public static. [Here ](https://github.com/castorini/anserini/blob/017841f33794508f760d26f58dedb5c770c1cbfc/src/main/java/io/anserini/analysis/AnalyzerUtils.java#L39)is static public, should I change it as well?

lintool: Yes, please fix while you're at it.

Indeed, should be `public static`: https://stackoverflow.com/questions/239961/which-is-recommended-static-public-or-public-static

lintool (approved):

PR #2860, opened 2025-06-26
Tweak score tolerance in repro: add [OK*]

lintool:
Added better handling of cases when scores differ.
Additional minor tweaks.


lintool:
hi @lilyjge I think the tiny drop in codecov coverage is okay - once you +1, I can override and merge.

lilyjge (approved):

PR #2862, opened 2025-06-27
Add repro entry

suraj-subrahmanyan:
This PR adds my repro entry.

**System:** Xubuntu 22.04.5 LTS (running in VM on Windows host via VirtualBox)

**VM Resources:** 4060MB RAM + 2CPUs

**Python Version:** Python 3.10.12

Note: `python3` had to be used instead of `python` as it was not recognized for the commands that ran python scripts. 
Otherwise, reproduced exactly.


suraj-subrahmanyan:
Gotcha, will do!

suraj-subrahmanyan:
Hi @lintool, I was sorting out some git issues haha, please don't mind the force-push I made. As per your instructions, I have consolidated the repro entries for Anserini in one PR, and would like to know if this is what you were looking for. Thanks!

lintool (approved):

Issue #2864, opened 2025-06-29
Document SearchShardedHnswDenseVectors for TREC RAG


lintool:
Hi @lilyjge we have `SearchShardedHnswDenseVectors`, but probably needs a bit of refactoring?

We probably should document in `fatjar-regressions-v1.1.0.md`?

Right now, the instructions search each shard and then concat together... should be supplemented by `SearchShardedHnswDenseVectors`?

I.e., here's how you do each shard (docs we have now) - but here's the command to do it all at once.

wdyt?

lintool:
Final check in #2871 - done!

PR #2866, opened 2025-06-30
Add to doc for searching shards

lilyjge:
Addresses #2864.


lintool (changes requested):
hi @lilyjge in this PR, can you add the corresponding command to fetch the docs using `io.anserini.rerank.OutputRerankerRequests`? I.e., move that part down so it applies not only to BM25, but this also.

PR #2875, opened 2025-07-01
Remove IndexReaderUtils getReader with path

lilyjge:
Addresses castorini/pyserini#2166.


lilyjge:
Jar passes all Pyserini tests.

PR #2880, opened 2025-07-05
Add -dryRun and -printCommands to io.anserini.reproduce.RunRepro

lintool:
As a result, the fatjar docs don't need to explicitly render commands. So we can do something like:

```
bin/run.sh io.anserini.reproduce.RunMsMarco -collection msmarco-v2.1-doc -dryRun -printCommands
```


lintool:
@wu-ming233 I added tests, but it's not practical to cover the non-dry-run branches in order to get test coverage up.

PR #2882, opened 2025-07-06
Add BRIGHT prebuilt indexes

lilyjge:


lintool (approved):
I assume you remembered to `-optimize` the prebuilt indexes down to a single segment?

lilyjge:
Yep! I'll add the command I used to Pyserini metadata next. Speaking of which, I made a typo with the name of the readme for these indexes, fixed with the latest commit.

PR #2883, opened 2025-07-06
Replace ThreadPoolExecutor with newWorkStealingPool

Adefioye:
The goal of this work is to resolve this [issue](https://github.com/castorini/anserini/issues/2579)

In this work, all instances of __ThreadPoolExecutor__ have been replaced with __newWorkStealingPool__  in the codebase with the benefit of maximizing CPU utilization. In addition, `executor.invokeAll(tasks)` has been used in place of different instances of `executor.awaitTermination(), executor.shutdown() and executor.shutdownNow()` necessary for ensuring complete execution of multiple tasks before moving unto other code executions in a file.

I have also added a `monitor` in `BenchmarkCollectionReader, AbstractIndexer, and IndexCollection` class to maintain the existing INFO logs in the process of refactor.

Finally, I have also left in `processSegments(executor, segmentpaths)` for backward compatibility as I am not sure if `anserini` is being consumed in another library that I am not aware of.

__All tests are currently passing locally__.

### DIFF
--- a/src/main/java/io/anserini/index/AbstractIndexer.java
+++ b/src/main/java/io/anserini/index/AbstractIndexer.java
@@ -39,11 +39,14 @@ import java.lang.reflect.InvocationTargetException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.Callable;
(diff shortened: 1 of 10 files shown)


Adefioye:
Hi @lintool 

Please do you have any objections/feedback on the implementation. Also, I guess I will also need to perform regression tests for the PR to be acceptable?

Issue #2884, opened 2025-07-07
BRIGHT regressions


lilyjge:
From-scratch regressions for BRIGHT like https://github.com/castorini/anserini/tree/master/src/main/resources/regression.

PR #2900, opened 2025-07-17
Add BRIGHT regression documentation

lilyjge:
Links added in #2899 should now work. 
Addresses #2884.


lilyjge:
I agree for both pretty names and groupings. I'll update.

lilyjge:
@lintool does this look good? Alternatively, we could remove the grouping headings but keep the order? wdyt?

lintool (approved):
I like this!

Can you add:
+ link to original paper
+ link to website + leaderboard
+ link to our post-processed corpus
+ link to the post-processed script

lintool (approved):

PR #2903, opened 2025-07-18
Update reproduction log :)

MindlessTruffle:


lintool:
Please add some details about what you did, specs of your system, what happened, etc.

lintool:
hi @MindlessTruffle can you please fix conflicts?

MindlessTruffle:
Using the developer installation of anserini - and downloading the raw fatjar, I was able to use all bash commands in the onboarding without error.

I am running WSL: Ubuntu on Windows with (48/2) =24 GB of ram available to it, within VS Code.

I also found it helpful to alias a lot of commands within alternate syntax to the onboarding in WSL (Ubuntu, Windows 11) so that commands in the explanation docs run without error.

PR #2936, opened 2025-08-21
Add BRIGHT BGE-large-en-v1.5 Lucene flat prebuilt indexes, topic bindings, ONNX model

lilyjge:
Lucene flat indexes achieve same scores as Pyserini Faiss flat indexes and what the BRIGHT paper reported. Also adds query ids exclusion for search.

Includes ONNX conversion scripts. We have to convert it ourselves instead of using the one provided with the model on Hugging Face for integration with existing Anserini code. ONNX scores are close to cached ones, so the conversion should be good.


lilyjge:
Yes, I'll do that before merging. Or should I sent a separate PR?

lintool:
> Yes, I'll do that before merging. Or should I sent a separate PR?

nah, just do it before merging.

lintool:
@lilyjge I assume you'll add `io.anserini.reproduce.RunBright` bindings next?

PR #2939, opened 2025-08-21
FeatureVector no longer uses fastutil Hash Map

JJGreen0:
Speed Comparison:

```
- [HashMap] Adds: 1,000,000 ops in 0.265 s (3.8 Mops/s)
- [HashMap] L2 scale: n=198,675 in 0.083 s
- [HashMap] L1 scale: n=198,675 in 0.086 s
- [HashMap] Prune to k=200: in 0.119 s
- [fastutil] Adds: 1,000,000 ops in 0.157 s (6.4 Mops/s)
- [fastutil] L2 scale: n=198,675 in 0.038 s
- [fastutil] L1 scale: n=198,675 in 0.053 s
- [fastutil] Prune to k=200: in 0.150 s
```

- Adds: time to apply many addFeatureValue(term, +1.0f) updates.
- L2/L1 scale: time to normalize all feature weights to unit L2/L1 norm.
- Prune: time to keep the top k largest-weight features.

### DIFF
--- a/src/main/java/io/anserini/util/FeatureVector.java
+++ b/src/main/java/io/anserini/util/FeatureVector.java
@@ -16,18 +16,18 @@
 
 package io.anserini.util;
 
-import it.unimi.dsi.fastutil.objects.Object2FloatOpenHashMap;
-
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 public class FeatureVector {
-  private Object2FloatOpenHashMap<String> features = new Object2FloatOpenHashMap<String>();
+  private Map<String, Float> features = new HashMap<>();
 
   public enum Order {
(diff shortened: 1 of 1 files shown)


JJGreen0:
Yes, this class is used in relevance feedback, via the the RM3 and Rocchio rerankers. 

- RM3: src/main/java/io/anserini/rerank/lib/Rm3Reranker.java
    - Core calls: FeatureVector.fromTerms(...), addFeatureValue(...), pruneToSize(...), scaleToUnitL1Norm(), interpolate(...),
iteration over features to build the Lucene query.
- Rocchio: src/main/java/io/anserini/rerank/lib/RocchioReranker.java
    - Core calls: same FeatureVector methods as RM3, plus repeated getValue(...) lookups in the feedback aggregation.

First-pass retrieval is not affected
Indexing and the Lucene scorer are unaffected

The end-to-end slow downs shouldn't be too large because the affected methods (add, L2/L1 scale, Prune) aren't the most dominant costs compared to sorting and reading term info. Also, the expected range of map operations per query is much smaller than the amount of operations in the tests, and translates to sub-millisecond differences per query for the FeatureVector portion.

PR #2949, opened 2025-08-24
Add evaluation_bm25_full.py for BM25 metric evaluation on dev queries

CereNova:
This PR adds a new script scripts/evaluation_bm25_full.py to evaluate BM25 retrieval results on the MS MARCO passage dev set.

**Features:**
Computes MRR@10, MAP, and Recall@1000.
Uses paths relative to the project root for better portability.
Can be added directly to the repository without submodule issues.
BM25 is run using anserini-0.24.1-fatjar.jar to match the official Anserini version.

**Usage:**

`python scripts/evaluation_bm25_full.py
`

This script provides a quick and reproducible way to evaluate the BM25 baseline on all dev queries.


lintool:
Hi @CereNova - Thanks for your contribution, but I don't think it's necessary to add your script to the repo.

PR #2959, opened 2025-09-04
Add query-side BM25 and BRIGHT regressions

lilyjge:


lintool:
hi @lilyjge in the paper https://arxiv.org/abs/2509.02558 we call it "query-side BM25" - can we have some alignment with the classes, but without awkward names?

> This page documents BM25 regression experiments for [BRIGHT &mdash; AoPS](https://brightbenchmark.github.io/), with query vectors generated by BM25 scoring.

Explicitly reference the paper and use the phrase "query-side BM25" somewhere?

`BM25QueryGenerator` -> `QuerySideBm25QueryGenerator`? A mouthful, but more accurate? Also `Bm25` instead of `BM25` to conform to standard naming conventions?

src/main/java/io/anserini/search/SearchCollection.java
```diff
@@ -270,6 +271,12 @@ public static class Args extends BaseSearchArgs {
     @Option(name = "-bm25.b", handler = StringArrayOptionHandler.class, usage = "BM25: b parameter")
     public String[] bm25_b = new String[]{"0.4"};
 
+    // -------------------
+    // whether to apply BM25 on the query
+    // -------------------
+    @Option(name = "-bm25query", usage = "boolean switch to apply BM25 on the query")
```
lintool: does it make sense to rename to `-bm25.querySide`?

src/main/java/io/anserini/search/query/QuerySideBm25QueryGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+  private final float avgLength;
+
+  public QuerySideBm25QueryGenerator(float k1, float b, IndexReader indexReader) {
+    this.k1 = k1;
+    this.b = b;
+    this.indexReader = indexReader;
+    try{
+      this.avgLength = indexReader.getSumTotalTermFreq(Constants.CONTENTS) / (float)indexReader.getDocCount(Constants.CONTENTS);
+    }
+    catch (Exception e) {
```
lintool, line 53: I think "house style" is the `catch` on the same line the previous `{`

docs/regressions/regressions-bright-aops.bm25q.md
```diff
@@ -0,0 +1,77 @@
...
+bin/trec_eval -c -m ndcg_cut.10 tools/topics-and-qrels/qrels.bright-aops.txt runs/run.bright-aops.bm25query.topics.bright-aops.txt
+bin/trec_eval -c -m recall.100 tools/topics-and-qrels/qrels.bright-aops.txt runs/run.bright-aops.bm25query.topics.bright-aops.txt
+bin/trec_eval -c -m recall.1000 tools/topics-and-qrels/qrels.bright-aops.txt runs/run.bright-aops.bm25query.topics.bright-aops.txt
+```
+
+## Effectiveness
+
+With the above commands, you should be able to reproduce the following results:
+
+| **nDCG@10**                                                                                                  | **query-side-BM25**|
```
lintool: minor nit, can you change the table so that the columns align in MD?

lilyjge: wdym? they look like this right now, seems to align with [others](https://github.com/castorini/anserini/blob/master/docs/regressions/regressions-beir-v1.0.0-arguana.bge-base-en-v1.5.parquet.flat-int8.cached.md)
image
also, just noticed these docs are leftovers with the old naming. I will delete them. please look at the ones named with `bm25qs` instead.

lintool: I meant alignment in the markdown.

Not aligned:

```
| header1 | header2 |
|---------|---------|
| value | value |
```

Aligned:

```
| header1 | header2 |
|---------|---------|
| value   | value   |
```

Just a nit...

lilyjge: Since the docs are auto generated, I need to change the doc auto generation code, which affects a lot of regression docs (700+). I will send another PR for that.

PR #2963, opened 2025-09-06
Fix metadata for msmarco-v2-passage in IndexInfo and other minor fixes

lintool:
+ Fixed correct tarball for `msmarco-v2-passage`.
+ Other minor fixes.


lilyjge (approved):

lilyjge:
For BRIGHT lucene indexes, `QuerySideBm25QueryGenerator` could also work, but I guess default bag of words generator for consistency is better?

PR #2974, opened 2025-09-16
Remove fastutil dependency

lintool:
Thanks to @JJGreen0 #2939 and @suraj-subrahmanyan #2945 - we've eliminated the fastutil dependency!

Fatjar size drops from 176mb to 155mb, yay!


ronakice (approved):
LG

PR #2977, opened 2025-09-19
Add Anserini onboarding reproduction log entries

mahdi-behnam:
- OS: Windows 11 (used WSL)
- CPU: Ryzen 7 5800H
- RAM: 16GB
- Storage: SSD
- Python: 3.12.2
- Java: OpenJDK 21.0.7

Results:
- I had to change `bin/run.sh` to use `-Xmx8G` because I ran out of memory. Even so, everything worked.

Suggestions: 
- I think it is better to use a regex-anchored grep to ensure the ID is in the first column, e.g., `grep -P '^999391\t' runs/run.msmarco-passage.dev.bm25.tsv | head` since plain `grep 999391 ... | head` can also match IDs appearing in later columns.


lintool:
hi @mahdi-behnam please fix conflicts.

mahdi-behnam:
Hi @lintool, I have resolved the conflicts.

PR #2978, opened 2025-09-24
Add to onboarding reproduction logs (#2978)

AniruddhThakur:
Operating System: macOs Sequoia 15.3.1

Working Environment: Visual Studio Code Version: 1.104.0

Successfully reproduced all the results on start-here Anserini page and also practiced mapping extra queries to relevance judgments and then to documents to familiarise myself with the concepts.

Potential typo: In the line "This means that we have only about one relevance judgments per query." I believe it should be "relevance judgement" instead, as we're talking about a singular relevance judgment, but I'm not sure as this is my first time interacting with this field.


lintool:

Thanks for catching the typo. Please send a fix in your PR.

PR #2984, opened 2025-09-28
Fix index size typo

lilyjge:


lintool:
@lilyjge since this wouldn't take effect until I publish a new anserini artifact, can you send a PR in pyserini to hardcode and fix this for now, with a TODO to fix once we've pushed out the next anserini release?

PR #2993, opened 2025-10-13
Add index stats

lilyjge:
Fixes castorini/pyserini#2291 by adding SPLADE-v3 index stats for BEIR and BRIGHT


src/main/java/io/anserini/index/IndexInfo.java
```diff
@@ -2472,7 +2472,11 @@ public enum IndexInfo {
       IndexType.SPARSE_IMPACT,
       SpladeV3Encoder.class.getSimpleName(),
       BagOfWordsQueryGenerator.class.getSimpleName(),
-      "beir-v1.0.0-trec-covid.flat"),
+      "beir-v1.0.0-trec-covid.flat",
+      -1,
```
lintool: why is size -1?

lilyjge: placeholder, fixed now

PR #3007, opened 2025-10-21
Add reproduction log 

LiHuua258:
All works well. I think that Command for Windows can be added, I search it on my own to finish.


lintool:

PR #3008, opened 2025-10-21
Add reproduction log entry for MSMARCO passage conversion (Windows)

RichHene:
Reproduced MSMARCO passage conversion on Windows 11 using Python 3.10 and Java 17.
Everything worked after adjusting PowerShell syntax (backticks instead of backslashes).
Commit: [7302ce9]


lintool:
Superseded by #3011

PR #3032, opened 2025-11-11
Various cleanup + refactoring

lintool:
Cleanup of various warnings inside VS code, with some help from Codex. Increased test coverage by a bit.
Slop is minimal.

@ronakice @lilyjge PTAL?

I'm running from-scratch regressions, so just a sanity check is fine.

### DIFF
--- a/src/main/java/io/anserini/util/BenchmarkCollectionReader.java
+++ b/src/main/java/io/anserini/util/BenchmarkCollectionReader.java
@@ -163,7 +163,7 @@ public final class BenchmarkCollectionReader {
     } catch (CmdLineException e) {
       System.err.println(e.getMessage());
       parser.printUsage(System.err);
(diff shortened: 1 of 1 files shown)


src/main/java/io/anserini/util/BenchmarkCollectionReader.java
```diff
@@ -164,11 +163,10 @@ public static void main(String[] args) throws Exception {
     } catch (CmdLineException e) {
       System.err.println(e.getMessage());
       parser.printUsage(System.err);
-      System.err.println("Example: "+ BenchmarkCollectionReader.class.getSimpleName() +
-          parser.printExample(OptionHandlerFilter.REQUIRED));
+      System.err.println("Example: "+ BenchmarkCollectionReader.class.getSimpleName() + parser.printExample(OptionHandlerFilter.REQUIRED));
```
ronakice: does String.format not work here? just checking

PR #3043, opened 2025-11-19
More cleanup + test cases

lintool:
+ Clean up warnings from VS Code
+ Add test cases with help from Codex

### DIFF
--- a/src/main/java/io/anserini/collection/CoreCollection.java
+++ b/src/main/java/io/anserini/collection/CoreCollection.java
@@ -84,7 +84,9 @@ public class CoreCollection extends DocumentCollection<CoreCollection.Document> 
         InputStreamReader inReader = new InputStreamReader(xzStream);
         bufferedReader = new BufferedReader(inReader);
       } else {
-        bufferedReader = new BufferedReader(new InputStreamReader(new FileInputStream(path.toString())));
+        FileInputStream fIn = new FileInputStream(path.toString());
+        InputStreamReader inReader = new InputStreamReader(fIn);
+        bufferedReader = new BufferedReader(inReader);
       }
 
       ObjectMapper mapper = new ObjectMapper();
--- a/src/main/java/io/anserini/rerank/lib/AxiomReranker.java
+++ b/src/main/java/io/anserini/rerank/lib/AxiomReranker.java
@@ -112,7 +112,6 @@ public class AxiomReranker<T> implements Reranker<T> {
   private final String field; // from which field we look for the expansion terms, e.g. "body"
   private final boolean deterministic;  // whether the expansion terms are deterministically picked
   private final long seed;
-  //private final String originalIndexPath;
   private final String externalIndexPath;  // Axiomatic reranking can opt to use
   // external sources for searching the expansion
   // terms. Typically, we build another index
@@ -146,7 +145,6 @@ public class AxiomReranker<T> implements Reranker<T> {
     this.N = n;
     this.M = top;
     this.beta = beta;
-    //this.originalIndexPath = originalIndexPath;
     this.externalIndexPath = externalIndexPath;
     this.outputQuery = outputQuery;
     this.searchTweets = searchTweets;


src/main/java/io/anserini/collection/CoreCollection.java
```diff
@@ -75,15 +75,16 @@ public static class Segment extends FileSegment<CoreCollection.Document> {
...
-        bufferedReader = new BufferedReader(new InputStreamReader(
-          new XZInputStream(new FileInputStream(path.toString()))));
+        FileInputStream fIn = new FileInputStream(path.toString());
+        XZInputStream xzStream = new XZInputStream(fIn);
+        InputStreamReader inReader = new InputStreamReader(xzStream);
+        bufferedReader = new BufferedReader(inReader);
       } else {
-        bufferedReader = new BufferedReader(new InputStreamReader(
-          new FileInputStream(path.toString())));
+        bufferedReader = new BufferedReader(new InputStreamReader(new FileInputStream(path.toString())));
```
sahel-sh: Nit: seems like this can also benefit from one new per line.
`FileInputStream(path.toString()` can also be moved to outside of if else block .

src/main/java/io/anserini/rerank/lib/AxiomReranker.java
```diff
@@ -111,7 +112,7 @@ public class AxiomReranker<T> implements Reranker<T> {
   private final String field; // from which field we look for the expansion terms, e.g. "body"
   private final boolean deterministic;  // whether the expansion terms are deterministically picked
   private final long seed;
-  private final String originalIndexPath;
+  //private final String originalIndexPath;
```
sahel-sh: nit: meant to remove?

src/main/java/io/anserini/rerank/lib/AxiomReranker.java
```diff
@@ -145,7 +146,7 @@ public AxiomReranker(Analyzer analyzer, Class parser, String originalIndexPath,
     this.N = n;
     this.M = top;
     this.beta = beta;
-    this.originalIndexPath = originalIndexPath;
+    //this.originalIndexPath = originalIndexPath;
```
sahel-sh: ditto

src/main/java/io/anserini/search/SimpleSearcher.java
```diff
@@ -296,21 +297,21 @@ public void set_rm3(String collectionClass, int fbTerms, int fbDocs, float origi
    * @param outputQuery flag to print original and expanded queries
    * @param filterTerms whether to filter terms to be English only
    */
+  @SuppressWarnings("unchecked")
```
sahel-sh, line 300: [untested] according to gemini the following gets rids of this warning:
```
public void set_rocchio(String collectionClass, int topFbTerms, int topFbDocs, int bottomFbTerms, int bottomFbDocs, float alpha, float beta, float gamma, boolean outputQuery, boolean useNegative) {
    // Update 1: Use the raw type or wildcard for the variable to match asSubclass output easily
    Class<? extends DocumentCollection> clazz = null;

    try {
        if (collectionClass != null) {
            Class<?> c = Class.forName("io.anserini.collection." + collectionClass);
            // Update 2: Use asSubclass for a safe runtime check
            clazz = c.asSubclass(DocumentCollection.class);
        }
    } catch (ClassNotFoundException e) {
        LOG.error("collectionClass: " + collectionClass + " not found!");
    } catch (ClassCastException e) {
        // Update 3: Handle the case where the class exists but isn't a DocumentCollection
        LOG.error("collectionClass: " + collectionClass + " is not a valid DocumentCollection!");
    }

    useRocchio = true;
    cascade = new RerankerCascade<String>("rocchio");
    
    // Note: Ensure RocchioReranker constructor accepts Class<? extends DocumentCollection>
    cascade.add(new RocchioReranker<String>(this.analyzer, clazz, Constants.CONTENTS, topFbTerms, topFbDocs, bottomFbTerms, bottomFbDocs, alpha, beta, gamma, outputQuery, useNegative));
    cascade.add(new ScoreTiesAdjusterReranker<String>());
}
```

lintool: nah, rather not mess with this on this round of review... will circle back and carefully check slop.

sahel-sh (approved):
LGTM

PR #3042, opened 2025-11-19
Refactor test cases that redirect and save stderr/stdout

lintool:
+ There are bunch of test cases that redirect and save stderr/stdout (and then check them); reorganizing them into common classes.
+ Suppress warnings for missing javadoc elements.

PR #3048, opened 2025-11-21
Refactor test cases to extract common helper function

lintool:
Addresses https://github.com/castorini/anserini/pull/3042#discussion_r2548058564

Previous implementation had `StdOutStdErrRedirectableLuceneTestCase` and `StdOutStdErrRedirectableTestCase`, one of which extends `LuceneTestCase`. All but two cases can extend the Lucene version, but for the two exceptions (due to weird concurrency bugs), I had to duplicate the code that save/restores stdout/stdrr. It was either that or have two separate classes.

### DIFF
--- a/src/test/java/io/anserini/analysis/AutoCompositeAnalyzerTest.java
+++ b/src/test/java/io/anserini/analysis/AutoCompositeAnalyzerTest.java
@@ -53,6 +53,7 @@ public class AutoCompositeAnalyzerTest extends StdOutStdErrRedirectableLuceneTes
   public void tearDown() throws Exception {
     restoreStdOut();
     restoreStdErr();
+    super.tearDown();
   }
 
   private static final Map<String, Object[][]> examples = new HashMap<>() {
--- a/src/test/java/io/anserini/collection/ParquetDenseVectorCollectionTest.java
+++ b/src/test/java/io/anserini/collection/ParquetDenseVectorCollectionTest.java
@@ -46,6 +46,7 @@ public class ParquetDenseVectorCollectionTest extends DocumentCollectionTest<Par
   @After
   public void tearDown() throws Exception {
     restoreStdErr();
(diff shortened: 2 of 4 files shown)


src/test/java/io/anserini/StdOutStdErrRedirectableLuceneTestCase.java
```diff
@@ -26,6 +26,16 @@ public abstract class StdOutStdErrRedirectableLuceneTestCase extends LuceneTestC
   protected PrintStream saveOut;
   protected PrintStream saveErr;
 
+  protected static void suppressJvmLogging() {
```
sahel-sh: Nit: Ideally this should be part of LuceneTestCase since now many of the test classes that don't need redirect are extending this class to supress logs...
Hpw about a two level class:SuppresedLoggingLuceneTestCase extends LuceneTestCase and StdOut... extends Suppressed....

src/test/java/io/anserini/index/IndexCollectionInovcationsTest.java
```diff
@@ -37,12 +38,14 @@ public static void setupClass() {
   public void setUp() throws Exception {
     redirectStdOut();
     redirectStdErr();
+    super.setUp();
   }
 
   @After
   public void cleanUp() throws Exception {
```
sahel-sh: nit: calling this function tearDown(), is more consistent with calling super.tearDown(), better yet if you decide on the two level class hierarchy, and only classes that need redirect inherit from StdOut... class then `restoreStdOut();
    restoreStdErr();` could be called in the parent class function rather than in every child. ditto for setUP

sahel-sh (approved):

lintool:
Addressed comments, thanks!
+ implemented two-level hierarchy as suggested
+  `cleanUp` -> `tearDown`

PR #3068, opened 2025-12-16
Add reproduction log entry for MS MARCO passage

anjanpa:
Step1: 
Reproduced on Ubuntu 22.04 (WSL2), Intel Core i5 11th gen, 16GB RAM. 
Encountered memory exhaustion on native Ubuntu dual boot.
 
Step2:
Initially, I encountered an issue with an outdated Java version.
I was using WSL with 8 GB of memory, which caused memory exhaustion during indexing.
Later, I modified the memory settings in bin/run.sh (changing them from 512 MB to a range of 4–6 GB), after which the indexing completed successfully.

Step3:
During dense retrieval, downloading the 26 GB index via the command line repeatedly failed due to process termination. As a workaround, the index was manually downloaded using a higher-bandwidth office network (24 Mbps) and then extracted within WSL, since both environments resided on the same drive. Additionally, the ext4.vhdx file had to be manually shrunk because WSL continued to occupy excessive disk space even after file deletion. Despite these adjustments, the final retrieval process required approximately 3 hours and 20 minutes to complete, which is unexpectedly slow given the system’s hardware specifications.


lintool:

PR #3093, opened 2026-01-16
Remove backward compatibility w/ Lucene 8

lintool:
Now that we've upgraded to Lucene 10, we can't read Lucene 8 indexes anyways, so removing hacks for backward compatibility w/ Lucene 8.

Note that I'm merging against `lucene10`, not `master`.

@zdann15 PTAL?


zdann15:
LGTM, removing some no longer useful tests and other cleanup

PR #3102, opened 2026-01-25
Start of refactoring index symbols in IndexInfo: replacement with PrebuiltInvertedIndex

lintool:
Current, prebuilt index metadata comprises enums in `IndexInfo`, hard-coded in Java: it's difficult to maintain/modify/etc.

Proposed solution is to load metadata externally, in json. This is `PrebuiltInvertedIndex` in `io.anserini.index.prebuilt`.

Since we have a lot of metadata in `IndexInfo`, it's probably difficult to do a full swap all in one go; better to do incrementally: `PrebuiltIndexHandler` refactored to us new metadata if available, but default to `IndexInfo` for everything else.

My plan is to get this reviewed, merged, and then move away from `IndexInfo` incrementally.

Aside - initial version of `PrebuiltInvertedIndex` generated by GPT Codex, but I had to do quite a bit of refactoring. Codex appears to produce code with unnecessary layers of indirection.

@lilyjge wdyt?

### DIFF
--- a/src/main/java/io/anserini/index/prebuilt/PrebuiltInvertedIndex.java
+++ b/src/main/java/io/anserini/index/prebuilt/PrebuiltInvertedIndex.java
@@ -40,7 +40,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.json.JsonMapper;
 
 public class PrebuiltInvertedIndex {
-  private static final String RESOURCE_DIR = "prebuilt";
+  private static final String RESOURCE_DIR = "prebuilt-indexes";
   private static final String RESOURCE_SUFFIX = ".json";
   private static final TypeReference<List<Entry>> ENTRY_LIST_TYPE = new TypeReference<List<Entry>>() {};
   private static final ObjectMapper MAPPER = JsonMapper.builder()
@@ -55,6 +55,9 @@ public class PrebuiltInvertedIndex {
   public static class Entry {
     @JsonProperty("name")
     public String name;
+
+    @JsonProperty("corpus_index")
+    public String corpusIndex;
 
     @JsonProperty("description")
     public String description;
--- a/src/main/java/io/anserini/util/PrebuiltIndexHandler.java
+++ b/src/main/java/io/anserini/util/PrebuiltIndexHandler.java
@@ -47,7 +47,7 @@ public class PrebuiltIndexHandler {
   private static final int CONNECT_TIMEOUT_MS = 60_000;
   private static final int READ_TIMEOUT_MS = 120_000;
 
-  private final String indexName;
+  private final String name;
   private final String filename;
   private final String md5;
   private final String[] urls;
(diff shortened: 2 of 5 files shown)


src/main/java/io/anserini/index/prebuilt/PrebuiltInvertedIndex.java
```diff
@@ -0,0 +1,147 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.json.JsonReadFeature;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.json.JsonMapper;
+
+public class PrebuiltInvertedIndex {
```
lilyjge, line 42: When the move happens to more types of indexes, we should probably have some kind of inheritance relationship, eg, inverted indexes and dense indexes share most of the metadata fields so they should inherit them from a common parent, but inverted indexes have the addition of index statistics. The instantiation code checking for the desired file would go in the parent as well, and classes for specific index types like this, eg, has `Inverted` in the name, would just have the additional metadata fields specific to that type of index. Not needed for this PR, just something I thought of for the future.

src/main/java/io/anserini/index/prebuilt/PrebuiltInvertedIndex.java
```diff
@@ -0,0 +1,147 @@
...
+
+    @JsonProperty("total_terms")
+    public long totalTerms;
+
+    @JsonProperty("documents")
+    public int documents;
+
+    @JsonProperty("unique_terms")
+    public long uniqueTerms;
+  }
```
lilyjge, line 88: The `invertedIndex` metadata field is missing, which in this case of inverted indexes would point back to the index itself. This field is currently used for generating reranker requests to easily get the corresponding inverted index for any index to get the raw documents https://github.com/castorini/anserini/blob/0171f5942d733e1135fd5dd09aea5879f5b82ebf/src/main/java/io/anserini/rerank/GenerateRerankerRequests.java#L222
We could add the field back in here or add a check in `GenerateRerankerRequests`, I think either would work.

src/main/java/io/anserini/util/PrebuiltIndexHandler.java
```diff
@@ -46,51 +47,49 @@ public class PrebuiltIndexHandler {
...
 
   private String cacheDirectory;
   private Path downloadFilePath;
   private Path indexFolderPath;
 
   /**
-   * Creates a <tt>PrebuiltIndexHandler</tt>.
-   * By default, the default cache directory is at <tt>~/.cache/pyserini/indexes</tt>.
-   * Alternatively, a custom, user-specified cache directory can be specified via the environment variable <tt>$ANSERINI_INDEX_CACHE</tt>
-   * or the system property <tt>anserini.index.cache</tt>.
```
lilyjge, line 60: these comments for the cache directory seem useful to keep?

src/main/java/io/anserini/util/PrebuiltIndexHandler.java
```diff
@@ -46,51 +47,49 @@ public class PrebuiltIndexHandler {
   private static final int CONNECT_TIMEOUT_MS = 60_000;
   private static final int READ_TIMEOUT_MS = 120_000;
 
-  private String indexName;
-  private IndexInfo indexInfo;
+  private final String indexName;
+  private final String filename;
```
lilyjge, line 51: should `filename` be capitalized like `indexName`?

src/main/resources/prebuilt-indexes/bright-inverted.json
```diff
@@ -0,0 +1,170 @@
+[
```
lilyjge, line 1: might be clearer to name the subfolder in resources `prebuilt-indexes` or `indexes` instead of just `prebuilt`?
Also, would it make more sense to put the metadata in the `tools` repo? Then Anserini and Pyserini can access them at the same time, and Pyserini doesn't have to wait for new Anserini releases for more indexes.

src/main/java/io/anserini/index/prebuilt/PrebuiltInvertedIndex.java
lintool: Agreed.

src/main/resources/prebuilt-indexes/bright-inverted.json
lintool: +1 `prebuilt-indexes`

But not sure if these belong in `tools`. But your broader point is taken. I think metadata for all Lucene indexes should go in Anserini, but that doesn't prevent Pyserini from reading them before an official Anserini release. Pyserini can read a public URL to fetch metadata descriptions?

src/main/java/io/anserini/index/prebuilt/PrebuiltInvertedIndex.java
lintool: Separately surfaced as #3108

src/main/java/io/anserini/util/PrebuiltIndexHandler.java
lintool: Actually, better to rename `indexName` to `name`, since it's a field of `PrebuiltIndexHandler`, so "index" is superfluous.

lilyjge (approved):

lilyjge:
LGTM!

Issue #3104, opened 2026-01-26
Add test case that tests fetching documents using prebuilt index metadata field


lintool:
> The invertedIndex metadata field is missing, which in this case of inverted indexes would point back to the index itself. This field is currently used for generating reranker requests to easily get the corresponding inverted index for any index to get the raw documents

from https://github.com/castorini/anserini/pull/3102#discussion_r2725999926

Referring to this above - @lilyjge can you please write a test case that will allow me to test this feature in my PR?

lintool:
Closed with #3107

PR #3107, opened 2026-01-27
Add test case for getting index with raw documents from JSON metatdata

lilyjge:
After some digging, I found that `GenerateRerankerRequests` uses `IndexInfo` directly to get the metadata field that specifies the inverted index with the raw documents. This means that if we want, #3102 can be **merged without a problem right now** since it does not affect the direct access of `IndexInfo`. 

However, if we want to slightly refactor `GenerateRerankerRequests` to use the pattern introduced in #3102 to  "use new metadata if available, but default to IndexInfo for everything else", then we have a problem since we don't have a field to access the raw inverted index from the JSON metadata entries. This PR is what the slight refactor to `GenerateRerankerRequests` could look like, where it currently errors because I try to access a field on PrebuiltInvertedIndex.Entry that doesn't exist. To test this further, you can use `GenerateRerankerRequestsTest#testPrebuilt`. 

#3104

### DIFF
--- a/src/main/java/io/anserini/rerank/GenerateRerankerRequests.java
+++ b/src/main/java/io/anserini/rerank/GenerateRerankerRequests.java
@@ -222,8 +222,7 @@ public class GenerateRerankerRequests<K extends Comparable<K>> implements Closea
       PrebuiltInvertedIndex.Entry entry = PrebuiltInvertedIndex.get(index);
       if (entry != null) {
         resolvedIndex = IndexReaderUtils.getIndex(entry.invertedIndex).toString();
-      }
-      else {
+      } else {
         IndexInfo currentIndex = IndexInfo.get(index);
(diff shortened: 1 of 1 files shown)


src/main/java/io/anserini/rerank/GenerateRerankerRequests.java
```diff
@@ -218,8 +219,14 @@ public IndexReader getIndexReader(String index) {
     }
 
     if (isPrebuiltLabel) {
-      IndexInfo currentIndex = IndexInfo.get(index);
-      resolvedIndex = IndexReaderUtils.getIndex(currentIndex.invertedIndex).toString();
+      PrebuiltInvertedIndex.Entry entry = PrebuiltInvertedIndex.get(index);
+      if (entry != null) {
+        resolvedIndex = IndexReaderUtils.getIndex(entry.invertedIndex).toString();
+      }
```
lintool: move else on same line? i.e.

> `} else {`

PR #3109, opened 2026-01-28
Add MMEB visdoc topics and qrels

sahel-sh:


sahel-sh:
I did run the tests locally, all of them including the newly added tests passed.
This seems to be an unrelated issue with the MSMARCO passage something. Rerunning the job

lilyjge:
Yep, I also took a look at the earlier failed run before approving, and can confirm it's an unrelated temporary download problem.

Issue #3108, opened 2026-01-28
PrebuiltInvertedIndex should inherit from PrebuiltIndex


lintool:
Build out an inheritance hierarchy to DRY. Follow up to #3102

PR #3112, opened 2026-01-29
Add new features to prebuilt index metadata

lintool:
+ Implemented `PrebuiltImpactIndex`
+ `PrebuiltImpactIndex` and `PrebuiltInvertedIndex` both inherit from ``PrebuiltIndex`
+ Added metadata for BRIGHT

@lilyjge started implementing some of your suggestions, e.g., #3108

Code mostly written by Codex, prompted through multiple steps. Minimal edits from me.

### DIFF
--- a/src/main/java/io/anserini/index/prebuilt/PrebuiltImpactIndex.java
+++ b/src/main/java/io/anserini/index/prebuilt/PrebuiltImpactIndex.java
@@ -35,7 +35,7 @@ public class PrebuiltImpactIndex extends PrebuiltIndex {
   private final Map<String, Entry> byName;
 
   private PrebuiltImpactIndex() {
-    List<Entry> loadedEntries = loadEntries("impact", ENTRY_LIST_TYPE, PrebuiltImpactIndex.class);
+    List<Entry> loadedEntries = loadEntries(PrebuiltIndex.Type.IMPACT, ENTRY_LIST_TYPE, PrebuiltImpactIndex.class);
     entries = Collections.unmodifiableList(loadedEntries);
 
     Map<String, Entry> map = new HashMap<>(Math.max(16, entries.size() * 2));
--- a/src/main/java/io/anserini/index/prebuilt/PrebuiltIndex.java
+++ b/src/main/java/io/anserini/index/prebuilt/PrebuiltIndex.java
@@ -47,6 +47,17 @@ public class PrebuiltIndex {
       .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
       .build();
 
+  public enum Type {
+    IMPACT("impact"),
+    INVERTED("inverted");
+
+    private final String id;
+
+    Type(String id) {
+      this.id = id;
+    }
+  }
+
   @JsonIgnoreProperties(ignoreUnknown = true)
   public static class Entry {
     @JsonProperty("name")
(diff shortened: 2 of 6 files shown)


src/main/java/io/anserini/index/prebuilt/PrebuiltImpactIndex.java
```diff
@@ -0,0 +1,70 @@
...
+    List<Entry> loadedEntries = loadEntries("impact", ENTRY_LIST_TYPE, PrebuiltImpactIndex.class);
+    entries = Collections.unmodifiableList(loadedEntries);
+
+    Map<String, Entry> map = new HashMap<>(Math.max(16, entries.size() * 2));
+    for (Entry entry : entries) {
+      if (entry != null && entry.name != null) {
+        map.put(entry.name, entry);
+      }
+    }
+
```
lilyjge, line 47: This seems repeated with PrebuiltInvertedIndex, maybe it can go in the parent class?

src/main/java/io/anserini/index/prebuilt/PrebuiltImpactIndex.java
```diff
@@ -0,0 +1,70 @@
...
+import com.fasterxml.jackson.core.type.TypeReference;
+
+public class PrebuiltImpactIndex extends PrebuiltIndex {
+  private static final TypeReference<List<Entry>> ENTRY_LIST_TYPE = new TypeReference<List<Entry>>() {};
+
+  // This is the singleton instance of this class.
+  private static PrebuiltImpactIndex INSTANCE;
+
+  public static class Entry extends PrebuiltIndex.Entry {}
```
lilyjge, line 32: does this need to be its own class or can we just use the parent's version?

src/main/java/io/anserini/index/prebuilt/PrebuiltIndex.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @JsonProperty("documents")
+    public int documents;
+
+    @JsonProperty("unique_terms")
+    public long uniqueTerms;
+  }
+
+  protected static <T extends PrebuiltIndex.Entry> List<T> loadEntries(
+      String type,
```
lilyjge: comment or enum for all possible values of type?

lintool:
> @lilyjge the decrease in test coverage comes from not testing the jar protocol for reading JSON... that's a bit challenging to test, so I propose we just eat the decrease and override? (after you've reviewed, ofc)

Well, apparently, Codex can do it... after some nudging: https://x.com/lintool/status/2016944837887479952

src/main/java/io/anserini/index/prebuilt/PrebuiltImpactIndex.java
lintool: Unfortunately, in Java, methods can only return one value. If it could return multiple, as in Python, we could avoid code duplication. Alternatively, we could return a "pair", but that introduces a new class.

src/main/java/io/anserini/index/prebuilt/PrebuiltImpactIndex.java
lintool: This allows for distinct types, which I think is nice to have...

src/main/java/io/anserini/index/prebuilt/PrebuiltIndex.java
lintool: Good suggestion.

PR #3115, opened 2026-01-30
Add new features to prebuilt index metadata

lintool:
+ Implement `PrebuiltFlatIndex`
+ Remove BRIGHT indexes from `IndexInfo`
+ Use holder pattern in `PrebuiltIndex` classes for thread safe initialization
+ Add tests

### DIFF
--- a/src/test/java/io/anserini/index/PrebuiltIndexHandlerTest.java
+++ b/src/test/java/io/anserini/index/PrebuiltIndexHandlerTest.java
@@ -89,8 +89,20 @@ public class PrebuiltIndexHandlerTest {
   }
 
   @Test
-  public void testGet() throws Exception {
+  public void testGetInverted() throws Exception {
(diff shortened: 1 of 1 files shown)


docs/prebuilt-indexes.md
```diff
@@ -200,59 +200,6 @@ Therefore, do not modify this page directly; modify the test case instead.
 </dd>
 </dl>
 </details>
```
lilyjge, line 202: It looks like that since the prebuilt indexes doc is currently generated from reading IndexInfo, removing the BRIGHT indexes from IndexInfo also removed them from the doc...

lintool: Yea, I'll circle around and add back in once we fully decommission `IndexInfo`. Is that okay?

lilyjge: Yeah sounds good!

PR #3117, opened 2026-01-31
Port of run_regression into Java

lintool:
The is a completely Codex-generated port of https://github.com/castorini/anserini/blob/master/src/main/python/run_regression.py

Will follow up and test...

### DIFF
--- a/src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
+++ b/src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
@@ -62,8 +62,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
-public class RunRegressionFromCorpus {
-  private static final Logger LOG = LogManager.getLogger(RunRegressionFromCorpus.class);
+public class RunRegressionsFromCorpus {
+  private static final Logger LOG = LogManager.getLogger(RunRegressionsFromCorpus.class);
 
   private static final String[] CORPUS_ROOTS = new String[] {
       "./",
--- a/src/test/java/io/anserini/reproduce/RunRegressionsFromCorpusTest.java
+++ b/src/test/java/io/anserini/reproduce/RunRegressionsFromCorpusTest.java
@@ -39,12 +39,12 @@ import io.anserini.search.SearchCollection;
 import io.anserini.search.topicreader.TopicReader;
 import io.anserini.search.topicreader.Topics;
 
-public class RunRegressionFromCorpusTest extends StdOutStdErrRedirectableLuceneTestCase {
+public class RunRegressionsFromCorpusTest extends StdOutStdErrRedirectableLuceneTestCase {
   @BeforeClass
   public static void setupClass() {
     suppressJvmLogging();
 
-    Configurator.setLevel(RunRegressionFromCorpus.class.getName(), Level.ERROR);
+    Configurator.setLevel(RunRegressionsFromCorpus.class.getName(), Level.ERROR);
     Configurator.setLevel(AbstractIndexer.class.getName(), Level.ERROR);
     Configurator.setLevel(IndexCollection.class.getName(), Level.ERROR);
     Configurator.setLevel(SearchCollection.class.getName(), Level.ERROR);
@@ -69,7 +69,7 @@ public class RunRegressionFromCorpusTest extends StdOutStdErrRedirectableLuceneT
     SortedMap<Integer, Map<String, String>> topics = TopicReader.getTopics(Topics.CACM);
     assertNotNull(topics);
 
-    RunRegressionFromCorpus.main(new String[] {
+    RunRegressionsFromCorpus.main(new String[] {
         "--regression", "cacm",
         "--index",
         "--search",
@@ -82,7 +82,7 @@ public class RunRegressionFromCorpusTest extends StdOutStdErrRedirectableLuceneT
     SortedMap<Integer, Map<String, String>> topics = TopicReader.getTopics(Topics.CACM);
     assertNotNull(topics);
 
-    RunRegressionFromCorpus.main(new String[] {
+    RunRegressionsFromCorpus.main(new String[] {
         "--regression", "cacm",
         "--index",
         "--search",
@@ -134,7 +134,7 @@ public class RunRegressionFromCorpusTest extends StdOutStdErrRedirectableLuceneT
     SortedMap<Integer, Map<String, String>> topics = TopicReader.getTopics(Topics.CACM);
     assertNotNull(topics);
 
-    RunRegressionFromCorpus.main(new String[] {
+    RunRegressionsFromCorpus.main(new String[] {
         "--regression", "cacm-download",
         "--index",
         "--search",


src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
```diff
@@ -0,0 +1,772 @@
...
+  private static final String RED = "\u001B[91m";
+  private static final String BLUE = "\u001B[94m";
+  private static final String RESET = "\u001B[0m";
+
+  private static final String FAIL = RED + "[FAIL]" + RESET + " ";
+  private static final String OK = "   [OK] ";
+  private static final String OKISH = "  " + BLUE + "[OK*]" + RESET + " ";
+
+  public static class Args {
+    @Option(name = "--regression", required = true, usage = "Name of the regression test.")
```
sahel-sh, line 99: This support a single test a time, why not a list of tests?

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
```diff
@@ -0,0 +1,772 @@
...
+import java.util.Objects;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+public class RunRegressionFromCorpus {
+  private static final Logger LOG = LogManager.getLogger(RunRegressionFromCorpus.class);
+
+  private static final String[] CORPUS_ROOTS = new String[] {
```
sahel-sh, line 68: Would be nice to be able to add to this default roots

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
```diff
@@ -0,0 +1,772 @@
...
+  };
+
+  private static final String INDEX_COMMAND = "bin/run.sh io.anserini.index.IndexCollection";
+  private static final String INDEX_FLAT_DENSE_COMMAND = "bin/run.sh io.anserini.index.IndexFlatDenseVectors";
+  private static final String INDEX_HNSW_DENSE_COMMAND = "bin/run.sh io.anserini.index.IndexHnswDenseVectors";
+  private static final String INDEX_INVERTED_DENSE_COMMAND = "bin/run.sh io.anserini.index.IndexInvertedDenseVectors";
+  private static final String INDEX_STATS_COMMAND = "bin/run.sh io.anserini.index.IndexReaderUtils";
+  private static final String SEARCH_COMMAND = "bin/run.sh io.anserini.search.SearchCollection";
+  private static final String SEARCH_FLAT_DENSE_COMMAND = "bin/run.sh io.anserini.search.SearchFlatDenseVectors";
+  private static final String SEARCH_HNSW_DENSE_COMMAND = "bin/run.sh io.anserini.search.SearchHnswDenseVectors";
```
sahel-sh, line 87: bin/run.sh is repeated in all, some sort of template based string formatting would be less error prone, easier to maintain, etc

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
```diff
@@ -0,0 +1,772 @@
...
+  private static final String SEARCH_INVERTED_DENSE_COMMAND = "bin/run.sh io.anserini.search.SearchInvertedDenseVectors";
+
+  private static final String RED = "\u001B[91m";
+  private static final String BLUE = "\u001B[94m";
+  private static final String RESET = "\u001B[0m";
+
+  private static final String FAIL = RED + "[FAIL]" + RESET + " ";
+  private static final String OK = "   [OK] ";
```
sahel-sh, line 95: hm, do you want to use green for ok?

src/main/resources/regression/cacm-download.yaml
```diff
@@ -0,0 +1,84 @@
+---
+corpus: cacm
+corpus_path: collections/cacm/raw/
+
+download_url: https://github.com/castorini/anserini-data/raw/refs/heads/master/CACM/corpus/raw/cacm-in-folder.tar.gz
```
lilyjge, line 5: why does cacm specifically have this file with the download url? the other regression yamls don't have this?

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
sahel-sh: spoke too soon, I guess all flags are designed for a single regression set

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
lintool: Agreed. Will circle back on this.

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
lintool: Agreed. Once this is merged, I can refactor to pull out common pieces.

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
lintool: Okay, good idea. And will pull out common: https://github.com/castorini/anserini/pull/3117#discussion_r2805467371

src/main/resources/regression/cacm-download.yaml
lilyjge: I'm still confused on why there's a separate yaml for download that's almost completely the same as cacm.yaml, but follow up on these fields specifically...

I found that some yamls have the following, but most don't. 
```
download_url
download_checksum
download_corpus
```
the trend seems to be that this is present for learned sparse expanded corpus, eg msmarco_v2.1_doc_segmented_splade-v3. we should probably consolidate/make consistent eg separate yamls/metadata info for collections

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
```diff
@@ -0,0 +1,772 @@
...
+        corpusPath = args.corpusPath;
+      }
+    } else {
+      String yamlCorpusPath = Objects.requireNonNull(yaml.get("corpus_path")).asText();
+      for (String root : CORPUS_ROOTS) {
+        Path candidate = Paths.get(root, yamlCorpusPath);
+        if (Files.exists(candidate)) {
+          corpusPath = candidate.toString();
+          break;
+        }
```
lilyjge, line 329: should we do this CORPUS_ROOT check when trying to download the corpus? if it exists, we can just copy it over

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
```diff
@@ -0,0 +1,772 @@
...
+    JsonNode yaml = mapper.readTree(Files.newInputStream(yamlPath));
+
+    long start = System.nanoTime();
+
+    if (parsed.download) {
+      // TODO: If collection already exists, skip.
+      LOG.info("========== Downloading Corpus ==========");
+      JsonNode downloadUrl = yaml.get("download_url");
+      if (downloadUrl == null || downloadUrl.asText().isEmpty()) {
+        throw new IllegalArgumentException("Corpus download URL unknown!");
```
lilyjge, line 155: related: https://github.com/castorini/anserini/pull/3117/changes#r2805513773
I think this is going to error for most regressions

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
```diff
@@ -0,0 +1,772 @@
...
+        if (Files.exists(candidate)) {
+          corpusPath = candidate.toString();
+          break;
+        }
+      }
+    }
+
+    if (corpusPath == null) {
+      throw new RuntimeException(String.format("Unable to find the corpus '%s' at %s: looked in %s",
+          yaml.get("corpus").asText(), yaml.get("corpus_path").asText(), Arrays.toString(CORPUS_ROOTS)));
```
lilyjge, line 335: the code also looks in args.corpusPath if it is supplied

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
```diff
@@ -0,0 +1,772 @@
...
+      LOG.info("Corpus path is {}", path);
+      parsed.corpusPath = path.toString();
+    }
+
+    if (parsed.index) {
+      LOG.info("========== Indexing ==========");
+      String command = constructIndexingCommand(yaml, parsed);
+      if (!parsed.dryRun) {
+        runCommand(command);
+      }
```
lilyjge, line 185: should we log the command if it is a dryrun?

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
```diff
@@ -0,0 +1,772 @@
...
+        LOG.info("Skipping verification step for flat dense indexes.");
+      } else {
+        StringBuilder cmd = new StringBuilder();
+        cmd.append(INDEX_STATS_COMMAND)
+            .append(" -index ").append(constructIndexPath(yaml))
+            .append(" -stats");
+        if (indexType != null && "inverted-dense".equals(indexType.asText())) {
+          cmd.append(" -field vector");
+        }
+        String output = checkOutput(cmd.toString());
```
lilyjge, line 203: do we not worry about dryruns for verification?

src/main/java/io/anserini/reproduce/RunRegressionsFromCorpus.java
```diff
@@ -0,0 +1,772 @@
...
+      try {
+        future.get();
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+    }
+  }
+
+  private static String downloadUrl(String url, Path saveDir, String localFilename, String md5, boolean force, boolean verbose)
+      throws IOException, NoSuchAlgorithmException {
```
lilyjge, line 661: nit: I think we have some similar code for downloading prebuilt indexes that we can probably reuse

src/main/resources/regression/cacm-download.yaml
lintool: For `cacm.yaml`, the corpus is already checked into the repo. To test the download feature, I created `cacm-download.yaml`, which downloads the corpus. But yes, otherwise they are the same.

As to which configs have download and which don't... it's a bit ad hoc, since the feature was added rather late. There's also an issue of usage rights - e.g., for example, we can't provide downloads for corpus that aren't open. However, the learned representations are fine, even if the original corpus isn't open, we can publicly host embeddings.

@lilyjge does this answer your question?

lilyjge: ah I see, that makes sense!

PR #3133, opened 2026-02-22
Add reproduction logs for 'Start Here' and 'MS MARCO BM25 baseline'

Sina-rokna:
Successfully reproduced:

1. Start Here
2. Anserini: BM25 Baselines for MS MARCO Passage Ranking

Environment:
- OS: WSL2 Ubuntu 22.04
- RAM: 8GB
- Java: 21
- Maven: 3.x
- Storage: SSD (NTFS mounted via WSL)

Results:
MRR@10: 0.1874  
MAP: 0.1957  
Recall@1000: 0.8573  

Observations and issues encountered:

1. Line ending issues:
   `bin/run.sh` and `bin/trec_eval` initially failed due to CRLF line endings when working inside WSL. This was resolved using `dos2unix`.

2. Memory constraints:
   Indexing with higher thread counts ( 9 threads) resulted in the process being killed by the system. Reducing the number of threads (to 2) resolved the issue.

3. Index size difference:
   The documented index size is approximately 4.3G (as you reported), whereas the my produced index in this environment was ~4.1G.
   Potential reasons:
   - Differences in Lucene minor version or codec implementation
   - Filesystem differences 
   - OS-level compression and block allocation behavior

   However Retrieval results were identical despite the size difference.

All evaluation metrics matched the documented baseline values.


lintool:
Having seen no further activity, closing.

PR #3217, opened 2026-04-25
Add NanoKnow v1.0 regressions

lingwei-gu:
Add BM25 release regressions for NanoKnow v1.0 over the Karpathy FineWeb-Edu 100B index, using the official supported topics and qrels merged in anserini-tools.

The regressions track R@20 for NQ-Open and SQuAD. Since the qrels are derived from BM25 top-100 over this same index, R@100 is effectively fixed by construction; R@20 provides a stable but non-trivial signal for ranking regressions across Anserini releases.


nourj98 (approved):
Matches the results I got, looks good! I wonder if it's worth adding the Recall@100 as sanity, but @lintool would know better

lintool (approved):

PR #3230, opened 2026-04-30
[codex] Add cache directory resolver

lintool:
## Summary
- add `CacheDirectoryResolver` for shared cache path resolution
- route prebuilt indexes, topics/qrels, and encoder cache users through the resolver
- add resolver tests and update qrels path expectations

## Test Plan
- `mvn -Dtest=CacheDirectoryResolverTest test`
- `mvn -Dtest=PrebuiltIndexHandlerTest,RelevanceJudgmentsTest,TopicsRegistryTest test`
- `bin/build.sh` (fails in `PrebuiltInvertedIndexTest.testUrls` due timeout connecting to `https://rgw.cs.uwaterloo.ca/pyserini/indexes/lucene/lucene-inverted.msmarco-v2-doc-segmented.unicoil-0shot.20220808.4d6d2a.tar.gz`)

### DIFF
--- a/docs/prebuilt-indexes.md
+++ b/docs/prebuilt-indexes.md
@@ -36,7 +36,9 @@ Nope, that's not a bug.
 
 Downloaded indexes are by default stored in `~/.cache/pyserini/indexes/`.
 (Yes, `pyserini`, that's not a bug &mdash; this is so prebuilt indexes can be shared between Pyserini and Anserini.)
-You can specify a custom cache directory by setting the environment variable `$ANSERINI_INDEX_CACHE` or the system property `anserini.index.cache`.
+If an `index/` directory exists in the current working directory, downloaded indexes are stored there.
+Otherwise, you can specify a custom base cache directory by setting the environment variable `$ANSERINI_CACHE` or the system property `anserini.cache`.
+If no custom base cache directory is specified and a `.cache/` directory exists in the current working directory, downloaded indexes are stored in `.cache/pyserini/indexes/` under the current working directory.
 
 Another helpful tip is to download and manage the indexes by hand.
 As an example, from the metadata in [`msmarco-v1-passage-inverted.json`](https://github.com/castorini/anserini/blob/master/src/main/resources/prebuilt-indexes/msmarco-v1-passage-inverted.json), you can see that `msmarco-v1-passage` can be downloaded from:
--- a/docs/prebuilt-indexes.md
+++ b/docs/prebuilt-indexes.md
@@ -34,11 +34,14 @@ Nope, that's not a bug.
 
 ## Managing Indexes
 
-Downloaded indexes are by default stored in `~/.cache/pyserini/indexes/`.
-(Yes, `pyserini`, that's not a bug &mdash; this is so prebuilt indexes can be shared between Pyserini and Anserini.)
-If an `index/` directory exists in the current working directory, downloaded indexes are stored there.
-Otherwise, you can specify a custom base cache directory by setting the environment variable `$ANSERINI_CACHE` or the system property `anserini.cache`.
-If no custom base cache directory is specified and a `.cache/` directory exists in the current working directory, downloaded indexes are stored in `.cache/pyserini/indexes/` under the current working directory.
+Downloaded indexes are stored in the first matching cache location:
+
+1. If the system property `pyserini.cache` is set, downloaded indexes are stored in `indexes/` under that base cache directory.
+2. Otherwise, if the environment variable `$PYSERINI_CACHE` is set, downloaded indexes are stored in `indexes/` under that base cache directory.
+3. Otherwise, if a `.cache/` directory exists in the current working directory, downloaded indexes are stored in `.cache/pyserini/indexes/` under the current working directory.
+4. Otherwise, downloaded indexes are stored in `~/.cache/pyserini/indexes/`.
+
+Yes, `pyserini`, that's not a bug &mdash; this is so prebuilt indexes can be shared between Pyserini and Anserini.
 
 Another helpful tip is to download and manage the indexes by hand.
 As an example, from the metadata in [`msmarco-v1-passage-inverted.json`](https://github.com/castorini/anserini/blob/master/src/main/resources/prebuilt-indexes/msmarco-v1-passage-inverted.json), you can see that `msmarco-v1-passage` can be downloaded from:
--- a/src/main/java/io/anserini/encoder/OnnxEncoder.java
+++ b/src/main/java/io/anserini/encoder/OnnxEncoder.java
@@ -80,7 +80,7 @@ public abstract class OnnxEncoder<T> implements AutoCloseable {
   }
 
   public Path getVocabPath() throws URISyntaxException, IOException {
-    File vocabFile = new File(getCacheDir(), vocabName);
+    File vocabFile = CacheDirectoryResolver.getEncodersCachePath().resolve(vocabName).toFile();
     if (!vocabFile.exists()) {
       FileUtils.copyURLToFile(new URI(vocabUrl).toURL(), vocabFile);
     }
@@ -88,22 +88,13 @@ public abstract class OnnxEncoder<T> implements AutoCloseable {
     return vocabFile.toPath();
   }
 
-  @SuppressWarnings("ResultOfMethodCallIgnored")
-  public String getCacheDir() {
-    File cacheDir = CacheDirectoryResolver.getEncodersCachePath().toFile();
-
-    if (!cacheDir.exists()) {
-      cacheDir.mkdirs();
-    }
-    return cacheDir.getPath();
-  }
-
   public Path getModelPath() throws IOException, URISyntaxException {
-    File modelFile = new File(getCacheDir(), modelName);
+    Path cachePath = CacheDirectoryResolver.getEncodersCachePath();
+    File modelFile = cachePath.resolve(modelName).toFile();
     if (!modelFile.exists()) {
       FileUtils.copyURLToFile(new URI(modelUrl).toURL(), modelFile);
       if (configUrl != null) {
-        File configFile = new File(getCacheDir(), this.getClass().getSimpleName() + '-' + BASE_CONFIG_NAME);
+        File configFile = cachePath.resolve(this.getClass().getSimpleName() + '-' + BASE_CONFIG_NAME).toFile();
         FileUtils.copyURLToFile(new URI(configUrl).toURL(), configFile);
       }
     }
--- a/src/main/java/io/anserini/encoder/OnnxEncoder.java
+++ b/src/main/java/io/anserini/encoder/OnnxEncoder.java
@@ -26,6 +26,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 
+import io.anserini.util.CacheDirectoryResolver;
+
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
@@ -35,8 +37,6 @@ import java.util.List;
 
 public abstract class OnnxEncoder<T> implements AutoCloseable {
   private static final Logger LOG = LogManager.getLogger(OnnxEncoder.class);
-
-  private static final String CACHE_DIR = Path.of(System.getProperty("user.home"), ".cache", "pyserini", "encoders").toString();
 
   private static final String BASE_CONFIG_NAME = "config.json";
 
@@ -90,10 +90,10 @@ public abstract class OnnxEncoder<T> implements AutoCloseable {
 
   @SuppressWarnings("ResultOfMethodCallIgnored")
   public String getCacheDir() {
-    File cacheDir = new File(CACHE_DIR);
+    File cacheDir = CacheDirectoryResolver.getEncodersCachePath().toFile();
 
     if (!cacheDir.exists()) {
-      cacheDir.mkdir();
+      cacheDir.mkdirs();
     }
     return cacheDir.getPath();
   }
--- a/src/main/java/io/anserini/eval/RelevanceJudgments.java
+++ b/src/main/java/io/anserini/eval/RelevanceJudgments.java
@@ -17,14 +17,12 @@
 package io.anserini.eval;
 
 import java.io.BufferedReader;
-import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
@@ -117,14 +115,6 @@ public class RelevanceJudgments {
     }
   }
 
-  private static String getCacheDir() {
-    File cacheDir = CacheDirectoryResolver.getTopicsAndQrelsCachePath().toFile();
-    if (!cacheDir.exists()) {
-      cacheDir.mkdirs();
-    }
-    return cacheDir.getPath();
-  }
-
   /**
    * Method will return the qrels file as a string
    * 
@@ -161,7 +151,7 @@ public class RelevanceJudgments {
     }
     if (!isContained && !isContainedSymbol) {
       // If the topic file is not in the list of known topics, we assume it is a local file.
-      Path tempPath = Paths.get(getCacheDir(), qrelsPath.getFileName().toString());
+      Path tempPath = CacheDirectoryResolver.getTopicsAndQrelsCachePath().resolve(qrelsPath.getFileName());
       if (Files.exists(tempPath)) {
         // if it is a unregistred topic in the Topics Enum, but it is in the cache, we use it.
         return tempPath;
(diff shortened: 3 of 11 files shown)


src/test/java/io/anserini/util/CacheDirectoryResolverTest.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  @Test
+  public void testLocalCacheSubpaths() throws IOException {
+    Path cwd = tempFolder.newFolder("local-subpaths").toPath();
+    Files.createDirectory(cwd.resolve(".cache"));
+    System.setProperty("user.dir", cwd.toString());
+
+    assertEquals(cwd.resolve("indexes"), CacheDirectoryResolver.getIndexCachePath());
+    assertEquals(cwd.resolve(".cache").resolve("topics-and-qrels"),
```
lintool: put on a single line.

src/test/java/io/anserini/util/CacheDirectoryResolverTest.java
```diff
@@ -0,0 +1,107 @@
...
+      System.setProperty(CacheDirectoryResolver.INDEX_CACHE_PROPERTY, indexCacheProperty);
+    }
+  }
+
+  @Test
+  public void testDefaultBasePath() throws IOException {
+    Path cwd = tempFolder.newFolder("no-local-cache").toPath();
+    System.setProperty("user.dir", cwd.toString());
+
+    assertEquals(Path.of(System.getProperty("user.home"), ".cache", "pyserini"),
```
lintool: put on a single line.

src/main/java/io/anserini/util/CacheDirectoryResolver.java
```diff
@@ -0,0 +1,74 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.anserini.util;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public final class CacheDirectoryResolver {
```
lintool, line 32: Document here the resolution rules as javadoc.

src/main/java/io/anserini/util/CacheDirectoryResolver.java
```diff
@@ -0,0 +1,74 @@
...
+  private static final String TOPICS_AND_QRELS_DIR = "topics-and-qrels";
+  private static final String ENCODERS_DIR = "encoders";
+
+  private CacheDirectoryResolver() {
+  }
+
+  public static Path getBasePath() {
+    Path localCachePath = getLocalCachePath();
+    if (Files.isDirectory(localCachePath)) {
+      return localCachePath;
```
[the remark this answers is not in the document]
lintool: If the users cwd is the user's home directory, then use the `~/.cache/pyserini/...` resolution.

src/main/java/io/anserini/util/CacheDirectoryResolver.java
```diff
@@ -0,0 +1,74 @@
...
+    if (cacheDirectory == null || cacheDirectory.isEmpty()) {
+      cacheDirectory = System.getenv(INDEX_CACHE_ENV);
+    }
+
+    if (cacheDirectory != null && !cacheDirectory.isEmpty()) {
+      return Path.of(cacheDirectory);
+    }
+
+    if (Files.isDirectory(getLocalCachePath())) {
+      return Path.of(System.getProperty("user.dir"), INDEXES_DIR);
```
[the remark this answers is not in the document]
lintool: If the users cwd is the user's home directory, then use the `~/.cache/pyserini/...` resolution.

lintool (commented):
More comments.

docs/prebuilt-indexes.md
```diff
@@ -36,6 +36,7 @@ Nope, that's not a bug.
 
 Downloaded indexes are by default stored in `~/.cache/pyserini/indexes/`.
 (Yes, `pyserini`, that's not a bug &mdash; this is so prebuilt indexes can be shared between Pyserini and Anserini.)
+If a `.cache/` directory exists in the current working directory and the current working directory is not your home directory, downloaded indexes are stored in `indexes/` under the current working directory.
```
lilyjge: it's a little ambiguous what this means. I can either understand this as if my CWD is `~/mydir/` and `~/mydir/.cache/` exists, then downloaded indexes are stored at 
- `~/mydir/.cache/indexes/` or
- `~/mydir/indexes/` 
and why is `current working directory is not your home directory` a condition?

docs/prebuilt-indexes.md
```diff
@@ -36,6 +36,7 @@ Nope, that's not a bug.
...
 You can specify a custom cache directory by setting the environment variable `$ANSERINI_INDEX_CACHE` or the system property `anserini.index.cache`.
```
lilyjge: for future: Pyserini's cache env is `PYSERINI_CACHE` https://github.com/castorini/pyserini/blob/master/docs/installation.md?plain=1#L225 but Anserini here is `ANSERINI_INDEX_CACHE`
we probably want it to be consistent?

src/main/java/io/anserini/util/CacheDirectoryResolver.java
```diff
@@ -0,0 +1,93 @@
...
+import java.nio.file.Path;
+
+/**
+ * Resolves cache directories used by Anserini. If a {@code .cache} directory exists in the current working directory
+ * and the current working directory is not the user's home directory, the base cache path is {@code <cwd>/.cache};
+ * otherwise, the base cache path is {@code ~/.cache/pyserini}. Index caches first honor the
+ * {@code anserini.index.cache} system property, then the {@code ANSERINI_INDEX_CACHE} environment variable. If neither
+ * override is set and local cache mode is active, indexes are stored in {@code <cwd>/indexes}; otherwise, indexes are
+ * stored in {@code <base>/indexes}. Topics, qrels, and encoders are stored under {@code <base>/topics-and-qrels} and
+ * {@code <base>/encoders}.
```
lilyjge: I guess this is the root of the inconsistency with Pyserini: design wise, why does the custom cache directory only apply to indexes and not encoders, topics, and qrels? I believe in Pyserini, all of them respect the custom cache dir

lintool:
Hi @lilyjge yupp, it might be a bit confusing, but I've thought through the scenarios...

With the help of Codex:

| Scenario | Base cache | Indexes | Topics/qrels | Encoders |
|---|---:|---:|---:|---:|
| 1. Dev installation, default | `~/.cache/pyserini` | `~/.cache/pyserini/indexes` | `~/.cache/pyserini/topics-and-qrels` | `~/.cache/pyserini/encoders` |
| 2. Dev installation, user creates `<repo>/.cache/` | `<repo>/.cache` | `<repo>/indexes` | `<repo>/.cache/topics-and-qrels` | `<repo>/.cache/encoders` |
| 3. Fatjar installation, default | `~/.cache/pyserini` | `~/.cache/pyserini/indexes` | `~/.cache/pyserini/topics-and-qrels` | `~/.cache/pyserini/encoders` |
| 4. Fatjar installation, user creates `<cwd>/.cache/` | `<cwd>/.cache` | `<cwd>/indexes` | `<cwd>/.cache/topics-and-qrels` | `<cwd>/.cache/encoders` |

Two important caveats:

+ If `cwd` is the user’s home directory, then `~/.cache/` does not trigger local mode. It stays under `~/.cache/pyserini/...`.
+ `anserini.index.cache` and `ANSERINI_INDEX_CACHE` override only the index cache path. They do not affect topics/qrels or encoders.

---

Basically, by default, Anserini uses `~/.cache/` - which signals that the user is okay using their home directory and having multiple clones share caches.

However, if the user explicitly creates `~/.cache/` - that's read as a signal "I want to keep everything within this directory". This is a signal that the user wants all downloaded files isolated to a single directory and not scattered about.

In this case, I want to preserve the repo layout as much as possible, since `index/` is already present in the repo.

Similarly for fatjar. The corner case is when users download the fatjar to their home directory.

---

wdyt @sahel-sh ? I think Pyserini and RankLLM should have the same logic?

cc @nourj98

docs/prebuilt-indexes.md
lintool: hi @lilyjge can you look at my comments below and see if it answers your question?

docs/prebuilt-indexes.md
lintool: Noted.

src/main/java/io/anserini/util/CacheDirectoryResolver.java
lintool: See explanation below...

lilyjge:
> `anserini.index.cache` and `ANSERINI_INDEX_CACHE` override only the index cache path. They do not affect topics/qrels or encoders.

Why though? Wouldn't it make more sense to allow the user's specified cache directory to apply to everything like it does in Pyserini?

> 4. Fatjar installation, user creates `<cwd>/.cache/`

This is the case I find a bit confusing. For example, the user is in their own project directory and has created `<cwd>/.cache/` for their own project or for Anserini topics/qrels/encoders. They might not necessarily want `indexes/` in their `cwd/`. So the preserving the repo layout only applies to Anserini/Pyserini, and not users in this case.

Just some design questions I had. If you like the current design, I can just +1.

lintool:
> > `anserini.index.cache` and `ANSERINI_INDEX_CACHE` override only the index cache path. They do not affect topics/qrels or encoders.
> 
> Why though? Wouldn't it make more sense to allow the user's specified cache directory to apply to everything like it does in Pyserini?

I see, _not_ have an override for INDEX specifically. Allow the user to specify the cache in general, not fine-grained suppress index.

I like this, 👍 - I think that's pretty clean.

> > 4. Fatjar installation, user creates `<cwd>/.cache/`
> 
> This is the case I find a bit confusing. For example, the user is in their own project directory and has created `<cwd>/.cache/` for their own project or for Anserini topics/qrels/encoders. They might not necessarily want `indexes/` in their `cwd/`. So the preserving the repo layout only applies to Anserini/Pyserini, and not users in this case.

That's a good point. In `<cwd>`, just us `.cache/`.

> Just some design questions I had. If you like the current design, I can just +1.

No, this is good, thanks @lilyjge - definitely welcome. Good to discuss.

lintool:
Hi @lilyjge here are the updated resolution rules:

Base:
1. Use `anserini.cache` if set.
2. Else use `ANSERINI_CACHE` if set.
3. Else, if `<cwd>/.cache/` exists, use `<cwd>/.cache/pyserini`.
4. Else use `~/.cache/pyserini`.

Indexes:
1. If `<cwd>/index/` exists, use `<cwd>/index/`.
2. Else use `<base>/indexes`.

Topics/qrels:
Use `<base>/topics-and-qrels`.

Encoders:
Use `<base>/encoders`.

wdyt?

src/main/java/io/anserini/encoder/OnnxEncoder.java
```diff
@@ -90,10 +90,10 @@ public Path getVocabPath() throws URISyntaxException, IOException {
 
   @SuppressWarnings("ResultOfMethodCallIgnored")
   public String getCacheDir() {
-    File cacheDir = new File(CACHE_DIR);
```
lintool, line 93: Refactor to get rid of this method? Why not call CacheDirectoryResolver directly?

src/main/java/io/anserini/eval/RelevanceJudgments.java
```diff
@@ -117,9 +118,9 @@ public Map<String, Integer> getDocMap(String qid) {
   }
 
   private static String getCacheDir() {
```
lintool: Refactor to get rid of this method? Why not call CacheDirectoryResolver directly?

src/main/java/io/anserini/reproduce/ReproduceFromPrebuiltIndexes.java
```diff
@@ -397,14 +398,7 @@ private static Path resolveSingleSymlinkChild(Path p) throws IOException {
   }
 
   private static String getCacheRoot() {
```
lintool: Refactor to get rid of this method? Why not call CacheDirectoryResolver directly?

src/main/java/io/anserini/search/topicreader/TopicReader.java
```diff
@@ -211,9 +211,9 @@ public static Map<String, Map<String, String>> getTopicsWithStringIdsFromFileWit
   }
 
   private static String getCacheDir() {
```
lintool: Refactor to get rid of this method? Why not call CacheDirectoryResolver directly?

lintool (commented):
Simplify?

docs/prebuilt-indexes.md
```diff
@@ -36,7 +36,9 @@ Nope, that's not a bug.
 
 Downloaded indexes are by default stored in `~/.cache/pyserini/indexes/`.
 (Yes, `pyserini`, that's not a bug &mdash; this is so prebuilt indexes can be shared between Pyserini and Anserini.)
-You can specify a custom cache directory by setting the environment variable `$ANSERINI_INDEX_CACHE` or the system property `anserini.index.cache`.
+If an `index/` directory exists in the current working directory, downloaded indexes are stored there.
```
lilyjge: the directory name in anserini is `indexes`

src/main/java/io/anserini/util/CacheDirectoryResolver.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+public final class CacheDirectoryResolver {
+  public static final String CACHE_PROPERTY = "anserini.cache";
+  public static final String CACHE_ENV = "ANSERINI_CACHE";
+
+  private static final String LOCAL_CACHE_DIR = ".cache";
+  private static final String DEFAULT_CACHE_PARENT = ".cache";
+  private static final String DEFAULT_CACHE_NAME = "pyserini";
+  private static final String LOCAL_INDEX_DIR = "index";
+  private static final String INDEXES_DIR = "indexes";
```
lilyjge: make consistent? as mentioned above Anserini uses `indexes`
actually, do we even need two variables? similarly with `LOCAL_CACHE_DIR` and `DEFAULT_CACHE_PARENT`

lilyjge (commented):
yes, I think that makes more sense! a few more caveats I can think of though:
- if I'm working on both Anserini and Pyserini (assuming that we carry this behaviour over), both repositories have the `indexes` directory, so indexes will be downloaded there, but what if I want to use the same indexes across packages? then they would be downloaded twice
- minor and related to above: since the cache directory is already named `~/.cache/pyserini/indexes/`, would it make sense to make the environment variable `PYSERINI_CACHE` to match? then it'd be consistent across anserini/pyserini

maybe only check for the `indexes/` directory in `cwd` if the env variable/system property isn't set? so the user's specification can always take precedence?

src/main/java/io/anserini/util/CacheDirectoryResolver.java
lintool: ugh, slop. thanks for catching.

docs/prebuilt-indexes.md
lintool: slop... :(
will de-slop-ify.

lintool:
okay, update:

---

## Cache Resolution Rules

### Base Cache

1. If `pyserini.cache` is set, use that as the base.
2. Otherwise, if `PYSERINI_CACHE` is set, use that as the base.
3. Otherwise, if `<cwd>/.cache/` exists, use `<cwd>/.cache/pyserini`.
4. Otherwise, use `~/.cache/pyserini`.

### Indexes

Use `<base>/indexes`.

### Topics and Qrels

Use `<base>/topics-and-qrels`.

### Encoders

Use `<base>/encoders`.

All resolver methods create the returned directory before returning it.

---

Simplified.

Scenarios -
+ If I do a normal dev checkout, I'll use `~/.cache/pyserini`; Anserini and Pyserini can share.
+ If I do a fatjar download, it'll use `~/.cache/pyserini` - but I can create `.cache/` locally as an explicit signal that I want everything to be in `cwd`.
+ If I set env variables, it'll just use those.

---

@lilyjge I think I've got everything covered?

src/main/java/io/anserini/util/CacheDirectoryResolver.java
```diff
@@ -0,0 +1,110 @@
...
+ * is {@code <cwd>/.cache/pyserini}. Indexes are stored under {@code <base>/indexes}. Topics and qrels are stored
+ * together under {@code <base>/topics-and-qrels}. Encoders are stored under {@code <base>/encoders}. All public
+ * accessors create the resolved directory before returning it.
+ */
+public final class CacheDirectoryResolver {
+  public static final String CACHE_PROPERTY = "pyserini.cache";
+  public static final String CACHE_ENV = "PYSERINI_CACHE";
+
+  private static final String LOCAL_CACHE_DIR = ".cache";
+  private static final String DEFAULT_CACHE_PARENT = ".cache";
```
lilyjge, line 37: just confirming that it's intentional for these two to be separate variables?

lilyjge (approved):
Yes, I'm happy with this!

src/main/java/io/anserini/util/CacheDirectoryResolver.java
lintool: yupp, in case they diverge in the future.

PR #3255, opened 2026-05-06
[codex] Capture trec_eval stderr in wrapper

lintool:
## Summary

- Capture the native `trec_eval` process stderr in `TrecEval.runAndGetOutput()` instead of letting it inherit the forked test JVM stderr stream.
- Preserve the existing stdout parsing behavior for successful metric extraction while preventing native stderr bytes from bypassing Surefire's framed fork protocol.
- Include captured stderr in the thrown exception when `trec_eval` exits nonzero, so failure diagnostics are still available without writing directly to the inherited stream.
- Enable native access for Surefire and `bin/run.sh` JVM launches to suppress Java FFM native-access warnings that can be emitted outside normal Java stream redirection.
- Add shared stdin redirect/restore helpers for tests, and update `GetDocumentTest` and `SearchTest` so interactive CLI cases do not leave `System.in` replaced or accidentally read Surefire's control stream.

## Context

`ReproduceFromDocumentCollectionTest` calls `TrecEval.runAndGetOutput()` as part of CACM regression verification. Under Surefire's forked JVM protocol, native subprocess output or direct reads/writes on inherited stdio can corrupt the control stream and surface as:

```text
[SUREFIRE] std/in stream corrupted
```

The wrapper already reads `trec_eval` stdout through a pipe. This change applies the same containment to stderr by draining it from Java and reporting it only through the Java exception path when needed.

Follow-up CI runs showed the same Surefire corruption later in the CLI tests. The interactive `GetDocumentTest` and `SearchTest` cases were replacing or reading from `System.in`; the tests now use explicit byte-array stdin fixtures and restore the original stream in teardown.

## Validation

- `mvn -Dtest=TrecEvalTest,ReproduceFromDocumentCollectionTest test`
- `mvn '-Dtest=io.anserini.reproduce.*Test' test`
- `mvn -Dtest=GetDocumentTest,ExtractQueriesAndDocumentsFromTrecRunTest,SearchTest,TrecEvalTest,ReproduceFromDocumentCollectionTest test`
- GitHub Actions `Java CI with Maven / build` passed on commit `61638596` after the stdin restoration fix.

### DIFF
--- a/src/main/java/io/anserini/eval/TrecEval.java
+++ b/src/main/java/io/anserini/eval/TrecEval.java
@@ -167,8 +167,7 @@ public class TrecEval {
     if (exit != 0) {
       String stderrMessage = stderr.toString(StandardCharsets.UTF_8).trim();
       if (!stderrMessage.isEmpty()) {
-        throw new RuntimeException(String.format("trec_eval ended with non-zero exit code (%d): %s",
-            exit, stderrMessage));
+        throw new RuntimeException(String.format("trec_eval ended with non-zero exit code (%d): %s", exit, stderrMessage));
       }
       throw new RuntimeException(String.format("trec_eval ended with non-zero exit code (%d)", exit));
     }
--- a/src/main/java/io/anserini/eval/TrecEval.java
+++ b/src/main/java/io/anserini/eval/TrecEval.java
@@ -17,13 +17,14 @@
 package io.anserini.eval;
 
 import java.io.BufferedOutputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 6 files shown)


src/main/java/io/anserini/eval/TrecEval.java
```diff
@@ -149,12 +158,18 @@ public String[][] runAndGetOutput(String[] args) {
...
       exit = p.exitValue();
     } catch (Exception e) {
       throw new RuntimeException(e);
     }
 
     if (exit != 0) {
+      String stderrMessage = stderr.toString(StandardCharsets.UTF_8).trim();
+      if (!stderrMessage.isEmpty()) {
+        throw new RuntimeException(String.format("trec_eval ended with non-zero exit code (%d): %s", exit,
+            stderrMessage));
```
lintool: stderrMessage on previous line

lintool:
@codex review

PR #3279, opened 2026-05-20
[codex] Make reproduction output locale-stable

lintool:
## Summary

- Make reproduction package formatting locale-stable with `Locale.ROOT`.
- Prevent randomized locale runs from emitting localized digits/decimal separators in reproduction logs.
- Add regression coverage for `ReproduceFromPrebuiltIndexesTest` using `ar-LB`.

## Validation

- `mvn -Dtest=ReproduceFromPrebuiltIndexesTest test`
- `mvn -Dtest=ReproduceFromPrebuiltIndexesTest#testCacmEndToEnd -Dtests.seed=C3765A39C8A9E9A5 -Dtests.locale=ar-LB -Dtests.timezone=Africa/Asmara -Dtests.asserts=true -Dtests.file.encoding=UTF-8 test`
- `mvn -Dtest=ReproduceFromPrebuiltIndexesTest,ReproduceFromDocumentCollectionTest,SummarizeLogsFromDocumentCollectionTest,SummarizeLogsFromPrebuiltIndexesTest,RunJavaReproductionCommandsTest test`
- `mvn -Dtest=RunJavaReproductionCommandsTest test`

### DIFF
--- a/src/main/java/io/anserini/reproduce/ReproduceFromDocumentCollection.java
+++ b/src/main/java/io/anserini/reproduce/ReproduceFromDocumentCollection.java
@@ -173,7 +173,7 @@ public class ReproduceFromDocumentCollection {
     try {
       parser.parseArgument(args);
     } catch (CmdLineException exception) {
-      System.err.println(String.format("Error: %s", exception.getMessage()));
+      System.err.println(String.format(Locale.ROOT, "Error: %s", exception.getMessage()));
       CliUtils.printUsage(parser, ReproduceFromDocumentCollection.class, argsOrdering);
 
       return;
@@ -197,7 +197,7 @@ public class ReproduceFromDocumentCollection {
     }
 
     if (parsedArgs.show) {
-      String resourceName = String.format("%s/%s.yaml", CONFIG_DIRECTORY, parsedArgs.config);
+      String resourceName = String.format(Locale.ROOT, "%s/%s.yaml", CONFIG_DIRECTORY, parsedArgs.config);
       try (InputStream yamlStream = ReproductionUtils.loadResourceStream(resourceName, ReproduceFromDocumentCollection.class)) {
         System.out.print(new String(yamlStream.readAllBytes(), StandardCharsets.UTF_8));
       }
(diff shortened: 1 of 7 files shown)


lintool:
I asked Codex if it was good practice to actually change the prod code, as opposed to make the tests more forgiving. It said:

```
› is it good practice to actual change the prod code in src/main/java/io/anserini/reproduce/ReproduceFromPrebuiltIndexes.java? or just make the test case more robust? don't write code, just discuss.


• Yes, changing the production code is the better fix here.

  The failure shows that ReproduceFromPrebuiltIndexes emits locale-dependent numeric output: ١ total, ٠٫٣١٢٣, ٠٠:٠٠:٠١, etc. That is not just a brittle test expectation. Reproduction logs are a user-facing artifact and are likely consumed by humans, docs, scripts, summaries, or CI logs. For a reproducibility tool,
  those logs should be stable across machines, locales, and randomized test environments.

  Making only the test “more robust” by accepting Arabic-Indic digits would hide the problem. It would say “this output variation is acceptable,” but I do not think it is. Reproduction output is closer to a CLI/reporting contract than localized UI text. Stable ASCII decimals are the right default for scientific
  metrics, command-line logs, generated summaries, and regression output.

  The production-code change is also low-risk because it only affects formatting, not retrieval, scoring, evaluation, or config parsing. Using Locale.ROOT/Locale.US around String.format/printf is the standard Java pattern when output is meant to be machine-stable rather than localized.

  The test should still be made more robust, but as a regression test for the desired behavior: under ar-LB, the tool should still print MAP: 0.3123, P30: 0.1942, 1 total, and ASCII duration digits. So the test becomes stricter in the right way: it verifies locale invariance instead of becoming permissive about locale
  drift.

  One nuance: if there were a product requirement that CLI output should localize for end users, then the production change would be questionable. But for Anserini reproduction logs, that would be surprising and probably harmful.
```

And later:

```
 Conceptually, reproduction code should behave like protocol/log generation, not localized UI. So the broader principle is sound: reproduction artifacts should be byte-stable across locales unless localization is explicitly intended.
```

sahel-sh (approved):

PR #3284, opened 2026-05-23
[codex] Add uniCOIL ONNX reproduction configs

lintool:
## Summary

- Add ONNX-backed uniCOIL reproduction configs for MS MARCO v1/v2 and TREC DL19-DL23 passage/doc-segmented tasks.
- Use plain topic files with `-encoder UniCoil` while preserving existing reported metric values from the cached configs.

## Validation

- `git diff --cached --check`
- Not run; config-only change.


lintool:
This is ready for review. Just the YAML files for now. I will circle back and add documentation pages, etc. once other PRs in flight have landed.

ronakice (approved):
Kinda sad we never did TREC RAG 24/25 with unicoil!

lintool:
> Kinda sad we never did TREC RAG 24/25 with unicoil!

Superseded by SPLADE family... but you're right, we should throw learned sparse at the corpus.

PR #2174, opened 2023-08-28
Use `KoreanAnalyzer` for Korean language (ko)

sudokim:
This PR enables the use of `KoreanAnalyzer`, an analyzer specialized for Korean.

The previous `CJKAnalyzer` only splits sequences into bi-grams, while `KoreanAnalyzer` splits a sentence into morphemes.

[LUCENE-8231](https://issues.apache.org/jira/browse/LUCENE-8231)

PR #3296, opened 2026-06-04
Update states from ko analyzer update

lintool:
As a result of merging #2174 and `ko` analyzer update, we need to update Mr.TyDi and MIRACL scores.


lilyjge (approved):
seems like it's a lot more effective, nice!

PR #3311, opened 2026-06-16
Onboarding: BM25 fallback + setup changes

Syum-Berihu:
### What was done in this PR:
* **Fixed a Bug in SearchCollection.java**: Added an `else` block so the code does not crash if someone forgets to type a scoring model like `-bm25`.
* **Added a Warning Message**: The program now prints a clear `[WARNING]` message on the screen telling the user that it is automatically defaulting to BM25.
* **Added a .gitignore File**: Set up rules to hide large local index folders, sample data, and temporary results from cluttering GitHub.
* **Onboarding Setup**: Completed local training experiments to make sure both Anserini (Java) and Pyserini (Python) work correctly.

### Local Testing Proof:
I compiled the code with Maven and tested both tools. The search results and scores match up perfectly:

* **Anserini Score**: 1 Q0 1 1 0.753400 Anserini
* **Pyserini Score**: 1 Q0 1 1 0.753400 Anserini

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,4 +82,18 @@ tools/topics-and-qrels/qrels.covid-round2-cumulative.txt
 /src/main/frontend/*.tsbuildinfo
 /src/main/frontend/next-env.d.ts
 /src/main/python/onnx/models/*
+/src/main/frontend/next-env.d.ts
+/src/main/python/onnx/models/*
 
+# local experiments (safe to ignore)
+anserini-data/
+runs/
+logs/
+indexes/
+
+# generated outputs
+results*.txt
+topics*.txt
+
+# temporary scripts
+test_*.py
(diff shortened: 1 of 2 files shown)


lintool:
hi @Syum-Berihu - thanks for your PR, but I don't think it's necessary to set a default in this case.

Syum-Berihu:
Thank you for the feedback and for reviewing the code! 

Just to clarify my approach, the logic handles a fallback scenario:
1. If a user enters another specific scoring model, it works correctly.
2. If no scoring model is provided at all, instead of crashing or failing silently, it falls back to the BM25 scoring model as a default and prints a warning message informing the user that BM25 is being used.


If setting a default is still not preferred here, I completely understand and will close things out on my end!

PR #3334, opened 2026-07-06
Fix sharded indexing dropping segments with negative hashCode

yashs33244:
`getSegmentPaths(shardCount, currShard)` filtered with `x.toString().hashCode() % shardCount == currShard`. Java `%` keeps the dividend's sign and `String.hashCode()` is negative for roughly half of inputs, so the result can be negative and match no `currShard` in `[0, shardCount-1]`. Those segments are then indexed by no shard and silently dropped (a quick check drops ~half of segments on a random set).

Switched to `Math.floorMod(...)` in `DocumentCollection` and in the `C4Collection` override (whose `NumberFormatException` fallback also returns a raw `hashCode()`). Added a `testShardedPartition` regression test asserting the shards partition the collection: every segment lands in exactly one shard (complete and disjoint). It runs for all collection subclasses.


lintool:
@codex review

lintool (approved):

PR #3338, opened 2026-07-09
Add BM25 reproduction log entry

abubinfahd:
- bin/run.sh – Changed the JVM heap from a fixed -Xmx192G to -Xmx${ANSERINI_XMX:-10G} (defaults to 10G, but can be overridden). This lets the fat JAR run on a typical machine without requiring 192 GB of RAM.

- PrebuiltIndexHandler.java – Expanded catch (Exception e) to catch (Exception | LinkageError e). When GitHub returns an HTTP 429 during the prebuilt-index metadata lookup, the error can surface as a LinkageError. This change lets Anserini fall back to the local index instead of exiting.


lintool:
(1) Please revert changes to `bin/run.sh`
(2) Please continue along the onboarding path until you reach the end of the exercises, and then we'll take a look all at once.

lintool:
Having seen no activity, closing.

PR #3348, opened 2026-07-14
[codex] Add local metadata for registered topics

lintool:
## Summary

- Add local topic metadata and aliases for registered topics.
- Refactor `Topics` into a metadata-backed registry loaded from JSON resources.
- Replace `Topics.resolve(...)` with explicit `Topics.load(...)` and `TopicReader.load(...)` entry points.
- Keep local-file topic loading explicit: local readable paths use `TopicReader.load(file, topicReader)`, while registry keys use `Topics.load(key)`.
- Update callers and tests to use canonical topic keys and remove legacy enum-style topic symbols and wrappers.
- Remove unused topic-loading helpers.

## Validation

- `mvn -Dtest=TopicsTest,TopicReaderTest,TopicsRegistryTest test`
- `mvn -Dtest=SearchCollectionTest test`
- `git diff --check`

### DIFF
--- a/src/main/java/io/anserini/search/SearchHnswDenseVectors.java
+++ b/src/main/java/io/anserini/search/SearchHnswDenseVectors.java
@@ -97,7 +97,7 @@ public final class SearchHnswDenseVectors<K extends Comparable<K>> implements Ru
         if (ref==null) {
           throw new IllegalArgumentException(String.format("\"%s\" does not refer to valid topics.", topicsFilePath));
         } else {
-          topics.putAll(TopicReader.getTopics(ref));
+          topics.putAll(TopicReader.load(ref));
         }
       } else {
         try {
--- a/src/main/java/io/anserini/search/SearchHnswDenseVectors.java
+++ b/src/main/java/io/anserini/search/SearchHnswDenseVectors.java
@@ -93,7 +93,7 @@ public final class SearchHnswDenseVectors<K extends Comparable<K>> implements Ru
     for (String topicsFile : args.topics) {
       Path topicsFilePath = Paths.get(topicsFile);
       if (!Files.exists(topicsFilePath) || !Files.isRegularFile(topicsFilePath) || !Files.isReadable(topicsFilePath)) {
-        Topics ref = Topics.getByName(topicsFile);
+        Topics ref = Topics.get(topicsFile);
         if (ref==null) {
           throw new IllegalArgumentException(String.format("\"%s\" does not refer to valid topics.", topicsFilePath));
         } else {
--- a/src/main/java/io/anserini/search/SearchHnswDenseVectors.java
+++ b/src/main/java/io/anserini/search/SearchHnswDenseVectors.java
@@ -91,25 +91,11 @@ public final class SearchHnswDenseVectors<K extends Comparable<K>> implements Ru
     // exceptions together as an unchecked exception to make initialization and error reporting clearer.
     SortedMap<K, Map<String, String>> topics = new TreeMap<>();
     for (String topicsFile : args.topics) {
-      Path topicsFilePath = Paths.get(topicsFile);
-      if (!Files.exists(topicsFilePath) || !Files.isRegularFile(topicsFilePath) || !Files.isReadable(topicsFilePath)) {
-        Topics ref = Topics.get(topicsFile);
-        if (ref==null) {
-          throw new IllegalArgumentException(String.format("\"%s\" does not refer to valid topics.", topicsFilePath));
-        } else {
-          topics.putAll(TopicReader.load(ref));
-        }
+      Path topicsPath = Paths.get(topicsFile);
+      if (Files.exists(topicsPath) && Files.isRegularFile(topicsPath) && Files.isReadable(topicsPath)) {
+        topics.putAll(TopicReader.load(topicsFile, args.topicReader));
       } else {
-        try {
-          @SuppressWarnings("unchecked")
-          TopicReader<K> tr = (TopicReader<K>) Class
-              .forName(String.format("io.anserini.search.topicreader.%sTopicReader", args.topicReader))
-              .getConstructor(Path.class).newInstance(topicsFilePath);
-
-          topics.putAll(tr.read());
-        } catch (Exception e) {
-          throw new IllegalArgumentException(String.format("Unable to load topic reader \"%s\".", args.topicReader));
-        }
+        topics.putAll(Topics.load(topicsFile));
       }
     }
 
--- a/src/main/java/io/anserini/search/SearchHnswDenseVectors.java
+++ b/src/main/java/io/anserini/search/SearchHnswDenseVectors.java
@@ -17,15 +17,11 @@
 package io.anserini.search;
 
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.SortedMap;
-import java.util.TreeMap;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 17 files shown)


lintool:
Initial refactor pass by Codex. Will now de-slopify.

lintool:
@codex review

lintool:
Wrapping up reproductions, working fine - ready for review.

```
# Invocation to run on batch01 on 102, batch02 on 104, and batch03 on tuna:
nohup bin/run.sh io.anserini.reproduce.RunJavaReproductionCommands --config from-document-collection.batch01 --load 32 --max 3 --sleep 60 >& from-document-collection.batch01.log &
nohup bin/run.sh io.anserini.reproduce.RunJavaReproductionCommands --config from-document-collection.batch02 --load 32 --max 3 --sleep 60 >& from-document-collection.batch02.log &
nohup bin/run.sh io.anserini.reproduce.RunJavaReproductionCommands --config from-document-collection.batch03 --load 44 --max 5 --sleep 60 >& from-document-collection.batch03.log &
```

lilyjge (commented):
would it make more sense for `src/main/resources/topics-and-qrels/_local_metadata_topics.json` and `src/main/resources/topics-and-qrels/_local_metadata_topics_aliases.json` to go in the tools repo? instead of an explicit topic reader class it could be a more vague enum topic type that get's mapped to the appropriate reader class in anserini/pyserini. the path and aliases could definitely be reused across pyserini/anserini as well

lintool:
hi @lilyjge thanks for your review - responses:

> would it make more sense for `src/main/resources/topics-and-qrels/_local_metadata_topics.json` and `src/main/resources/topics-and-qrels/_local_metadata_topics_aliases.json` to go in the tools repo? 

Yupp, the plan is for the metadata files to go in https://github.com/castorini/anserini-tools/tree/master/topics-and-qrels eventually.

I'm currently using `src/main/resources/topics-and-qrels/` as a staging area to make sure everything is okay before moving over.

This is to avoid churn, having to repeated touch files in another repo.

> instead of an explicit topic reader class it could be a more vague enum topic type that get's mapped to the appropriate reader class in anserini/pyserini. the path and aliases could definitely be reused across pyserini/anserini as well

My current plan is to decouple Anserini/Pyserini and have both read the metadata separately into their own internal classes. This decoupling allows Anserini/Pyserini to advance without inter-dependencies.

That is, if Pyserini depends on Anserini enums, then moving metadata to files doesn't shorten the dependency chain... you'd still have to modify Anserini, publish fatjar, and then modify Pyserini.

Under the current plan I'm in the process of implementing, we would just modify the json metadata, and Anserini/Pyserini can read the new info (prebuilt indexes, topics, and qrels) independently. The downside here is that there will be redundant code reading the metadata (in Java and Python), but I think this is a worthwhile tradeoff (to decouple Pyserini from Anserini and shorten the dependency chain).

Happy to discuss...

lilyjge:
this resolves my primary concern: `using src/main/resources/topics-and-qrels/ as a staging area to make sure everything is okay before moving over.`

yes, I agree with this!
`decouple Anserini/Pyserini and have both read the metadata separately into their own internal classes. This decoupling allows Anserini/Pyserini to advance without inter-dependencies.` 

what I meant about an enum type for the queries is that instead of specifying the anserini reader class, to abstract a level higher so it can be shared across anserini/pyserini, eg `reader_class: tsv` instead of `"reader_class": "io.anserini.search.topicreader.TrecTopicReader"` but we can iterate on that

design wise I think this is fine! I will take a pass through the code then

lintool:
> eg `reader_class: tsv` instead of `"reader_class": "io.anserini.search.topicreader.TrecTopicReader"` but we can iterate on that

Ah, I see.

However, currently, pyserini needs Anserini TopicReaders to read the topics: https://github.com/castorini/pyserini/blob/master/pyserini/search/_base.py#L122

So we need to rethink a bit more...

Yes, let's iterate on this further on another PR?

src/main/java/io/anserini/search/topicreader/Topics.java
```diff
@@ -16,656 +16,187 @@
...
+
+  private static void addLookupEntry(Map<String, Topics> lookup, String name, Topics topic) {
+    Topics existing = lookup.get(name);
+    if (existing != null && existing != topic) {
+      throw new IllegalStateException("Topic name maps to conflicting topics: " + name);
     }
+    lookup.put(name, topic);
   }
 
-  public static <K> SortedMap<K, Map<String, String>> resolve(String[] topicsArray, String topicReader) {
```
lilyjge, line 662: why did we remove this helper method? it was called in multiple places that now have chunks of code

src/main/java/io/anserini/search/SearchHnswDenseVectors.java
```diff
@@ -91,25 +91,11 @@ public SearchHnswDenseVectors(Args args) throws IOException {
...
     for (String topicsFile : args.topics) {
-      Path topicsFilePath = Paths.get(topicsFile);
-      if (!Files.exists(topicsFilePath) || !Files.isRegularFile(topicsFilePath) || !Files.isReadable(topicsFilePath)) {
-        Topics ref = Topics.getByName(topicsFile);
-        if (ref==null) {
-          throw new IllegalArgumentException(String.format("\"%s\" does not refer to valid topics.", topicsFilePath));
-        } else {
-          topics.putAll(TopicReader.getTopics(ref));
-        }
+      Path topicsPath = Paths.get(topicsFile);
```
lilyjge: for future note, the fact that this section has basically the same diff as `src/main/java/io/anserini/search/SearchFlatDenseVectors.java` makes me think there can be more abstraction to reduce duplication

src/test/java/io/anserini/search/topicreader/TopicReaderTest.java
```diff
@@ -813,47 +795,7 @@ public void testTREC20DL() throws IOException {
     assertEquals("why did the ancient egyptians call their land ke ##met , or black land ?", topics.get(topics.lastKey()).get("title"));
     assertEquals("who is aziz hash ##im", topics.get(1030303).get("title"));
 
-    topics = TopicReader.getTopics(Topics.TREC2020_DL_UNICOIL);
```
lilyjge, line 816: why did these tests get removed instead of replaced? similarly below

src/main/java/io/anserini/search/topicreader/Topics.java
lintool: It survived the initial metadata refactor, but I removed it in the follow-up API cleanup (`9498aafdd`). The intent was to separate registered-topic loading (`Topics.load(...)`) from local-file loading (`TopicReader.load(...)`), since `Topics` was becoming registry-only. In doing that, I moved the registered-vs-local dispatch and multi-file merge into the callers. So the functionality did not disappear, but the shared abstraction did—and that produced the duplicated chunks you noticed. The array overload itself had one direct production caller (`SearchCollection`), while similar loading logic existed in the other callers. You’re right that this should remain factored behind a shared helper, even if that helper no longer belongs on `Topics`.

src/test/java/io/anserini/search/topicreader/TopicReaderTest.java
lintool: Ah, we decided not to support cached queries, per https://github.com/castorini/anserini/discussions/3268

Support cached queries creates an extra code path (to maintain, to test, etc.) without any particular advantages...

src/main/java/io/anserini/search/topicreader/Topics.java
lintool: Restored the helper method - reduces code dup.

src/main/java/io/anserini/search/SearchHnswDenseVectors.java
lintool: Restored the helper method - reduces code dup.

lilyjge (approved):

PR #3354, opened 2026-07-23
Upgrade to Lucene 10.5.0

danmuzi:
Lucene 10.5.0 was released on June 25, 2026.
(https://lucene.apache.org/core/corenews.html#apache-lucenetm-1050-available)

The Lucene 10.5.0 upgrade requires only one API change in the current Anserini codebase:
the return type of `KnnVectorsWriter#mergeOneField` changed from `void` to `IORunnable` (apache/lucene#15732).

The reason for this change is that Lucene now performs vector merges in two phases.

In phase 1, `mergeOneField` performs the initial merge work—typically writing the flat vectors—and may return an `IORunnable` for deferred work.
In phase 2, Lucene executes the returned runnables to build expensive auxiliary structures (e.g., HNSW graphs) on top of the flat vectors written in phase 1.

This avoids writing the flat vectors twice during HNSW merges.
A writer with no deferred work simply returns `null`.

This PR updates the four delegating `KnnVectorsWriter` wrappers in`io.anserini.index.codecs` to match the new signature; they just return the delegate's result.
No other changes are needed. The default codec in 10.5.0 is still `Lucene104`, so the index format is unchanged. All existing tests pass without modification.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
   </developers>
 
   <properties>
-    <lucene.version>10.4.0</lucene.version>
+    <lucene.version>10.5.0</lucene.version>
     <jackson.version>2.22.1</jackson.version>
     <java.version>21</java.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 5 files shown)


lintool:
Hi @danmuzi thanks so much for the contribution. I've verified all regressions. Will merge.

lintool (approved):

PR #3364, opened 2026-08-05

mahimairaja:
Add reproduction log entries for the Foundations of Retrieval onboarding path (lessons 1-3): `start-here.md`, BM25 baseline, and dense retrieval on MS MARCO passage ranking.

## Setup

Runs were performed in a Linux container rather than on a laptop. The BGE HNSW step needs roughly 55 GB peak disk (the 24 GB `.tar.gz` is gunzipped to a `.tar` alongside it, then untarred alongside that, before anything is deleted) and my machine had 22 GB free.

- OS: Ubuntu 22.04 (container), CPU only, 8 vCPU / 32 GB RAM
- Java: Eclipse Temurin JDK 21
- Maven: 3.9.x, built with `bin/qbuild.sh`
- Commit: `5b3e896`

## Results

All three lessons reproduced exactly against the documented values.

**start-here.md**

- `collection.tsv`: 8,841,823 lines
- `collection_jsonl`: 9 files totalling 8,841,823 lines (last file 841,823)
- `queries.dev.small.tsv`: 6,980 lines
- `qrels.dev.small.tsv`: 7,437 lines

**experiments-msmarco-passage.md** (BM25 baseline)

- MRR@10 = `0.18741227770955546` (official MS MARCO eval script) — exact match
- `map = 0.1957`, `recall_1000 = 0.8573` (trec_eval) — exact match
- trec_eval MRR@10 = `0.1874` — exact match

**experiments-msmarco-passage2.md** (dense retrieval)

- Prebuilt BM25 index: MRR@10 = `0.1875` — exact match
- BGE-base HNSW dense retrieval: MRR@10 = `0.3521` — exact match

No issues encountered with the documented steps.

## Additional edits

`experiments-msmarco-passage2.md` pointed readers to the `IndexInfo` class for the list of prebuilt indexes. That class was removed in 4fbd16ba ("Remove IndexInfo: move all prebuilt index metadata to PrebuiltIndex"), so the link currently returns a 404. Updated it to point at `PrebuiltIndex`, and added the `PrebuiltIndexRegistry` CLI invocation since that is how the list is enumerated at runtime.

Also restored the missing trailing newline on the three edited files, so that subsequent reproduction log entries show up as a clean one-line diff instead of modifying the previous contributor's line.

### DIFF
--- a/docs/experiments-msmarco-passage.md
+++ b/docs/experiments-msmarco-passage.md
@@ -700,3 +700,4 @@ The BM25 run with default parameters `k1=0.9`, `b=0.4` roughly corresponds to th
 + Results reproduced by [@Hamza-Nadif](https://github.com/Hamza-Nadif) on 2026-07-28 (commit [`3a52e64`](https://github.com/castorini/anserini/commit/3a52e64fa331ade3de312b47b3674117aad399c4))
 + Results reproduced by [@niloy-biswas](https://github.com/niloy-biswas) on 2026-07-30 (commit [`02f25de`](https://github.com/castorini/anserini/commit/02f25dea042ee651085340d60ffbe9bc9820169b))
 + Results reproduced by [@dawoodkhandev](https://github.com/dawoodkhandev) on 2026-08-02 (commit [`ff848cb`](https://github.com/castorini/anserini/commit/ff848cb5872b42ee5305dbb1e27d72d2602aeade))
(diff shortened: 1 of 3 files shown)


lintool (approved):
Thanks for catching the `IndexInfo` issue!

mahimairaja:
It's my pleasure professor

lintool (approved):

PR #3371, opened 2026-08-09
Update scores based on upgrade to Lucene 10.5.0 in #3354

lintool:
Minor score tolerance tweaks due to Lucene 10.5.0 upgrade.


lintool:
Merging directly with CR.

PR #3384, opened 2026-08-16
[codex] Use canonical topic IDs in reproduction configs

lintool:
## Summary

- replace filename-style topic IDs with canonical Topics registry keys in document-collection reproduction configs
- regenerate the corresponding reproduction documentation
- preserve canonical `.wp` IDs and leave unresolved IDs unchanged

## Validation

- `mvn -Dtest=ReproduceFromDocumentCollectionTest test`
- `git diff --check`
- Parsed all 625 reproduction YAML configs and verified all 200 changed IDs against the pinned registry


lintool:
Fixes broken from-document regressions as part of recent commits to remove tools/ submodule.

PR #3389, opened 2026-08-19
[codex] Use index names in document collection run files

lintool:
## Summary

- construct document-collection run filenames from the `index_path` basename instead of `corpus`
- label model and topic components as `model-<name>` and `topics-<id>`
- standardize 110 legacy `lucene-index` config paths as `lucene-inverted`
- fix the `msmacro-doc` corpus typo in `dl20-doc.yaml`
- regenerate document-collection reproduction documentation

## Motivation

Multiple index variants can share the same corpus, model, and topics, causing run-file name collisions. Using the index basename distinguishes those runs and makes filenames reflect the configuration that produced them.

## Validation

- `mvn -Dtest=ReproduceFromDocumentCollectionTest,GenerateReproductionDocsFromDocumentCollectionTest test`
- `git diff --check`

All 14 focused tests passed.


lintool:
There's a name clash that causes errors (i.e., race condition). For example, `dl19-doc-segmented` and `dl19-doc-segmented.wp-ca` both write the same runfile, since we key on "corpus":

```bash
$ grep corpus src/main/resources/reproduce/from-document-collection/configs/dl19-doc-segmented.yaml 
corpus: msmarco-doc-segmented
corpus_path: collections/msmarco/msmarco-doc-segmented/
$ grep corpus src/main/resources/reproduce/from-document-collection/configs/dl19-doc-segmented.wp-ca.yaml 
corpus: msmarco-doc-segmented
corpus_path: collections/msmarco/msmarco-doc-segmented/
```

This PR fixes the issue.

lintool:
Fixes broken from-document regressions as part of recent commits to remove `tools/` submodule.

Issue #3403, opened 2026-08-31
start-here.md and experiments-msmarco-passage.md still reference the removed tools/ submodule


Ben-geo:
Working through the Foundations of Retrieval onboarding path, lessons 1-2. Both docs/start-here.md and docs/experiments-msmarco-passage.md instruct:

git submodule update --init --recursive

and then call python tools/scripts/msmarco/convert_collection_to_jsonl.py (also filter_queries.py, msmarco_passage_eval.py, convert_msmarco_to_trec_*.py).

The tools/ submodule was removed in #3382 , so there's no submodule to init and tools/scripts/ doesn't exist. I worked around it by git clone https://github.com/castorini/eval.git tools, which makes every tools/scripts/msmarco/... path resolve as written.

Happy to send a PR once you confirm the preferred fix (clone eval into tools/ in the docs vs. vendoring the scripts vs. something else).

lintool:
Yes, thanks for noticing this.

Please send a PR for:

(1) Remove references to `git submodule update --init --recursive`
(2) Make copies of needed scripts in `src/main/python/msmarco/`

PR #3404, opened 2026-08-31
Add to Onboarding Reproduction Logs

Ben-geo:
Setup: Windows 11 + WSL2 (Ubuntu 24.04), OpenJDK 21, Python 3.13. Fatjar
built with bin/qbuild.sh.

Results:
- BM25 msmarco_passage_eval.py: MRR@10 = 0.18741227770955546
- BM25 trec_eval: map = 0.1957, recall@1000 = 0.8573
- Prebuilt-index BM25: MRR@10 = 0.1875
- BGE-base dense (HNSW): MRR@10 = 0.3521

Issues / notes:
- start-here.md and experiments-msmarco-passage.md still say to run
  `git submodule update --init --recursive` and call tools/scripts/msmarco/*.py,
  but the tools/ submodule was removed in #3382 . Cloned castorini/eval into
  tools/ to proceed. - mentioned this in issue #3403 
- Had to lower -Xmx in bin/run.sh; the 192G default won't start on a laptop
  (the indexing guide already anticipates this).
- bin/trec_eval has no trec_eval-win-amd64 binary, so evaluation only runs
  under WSL/Linux, not native Windows.
- Prebuilt-index metadata resolution against api.github.com intermittently
  times out from the JVM even though curl to the same host succeeds; worked
  around with -Djava.net.preferIPv4Stack=true and an /etc/hosts pin.
  
  Future : 
  - Happy to raise a pr for #3403 Please let me know if I missed something and this is is already resolved somehow


lintool:
Please resolve conflicts...

Ben-geo:
resolved conflicts!

lintool (approved):

PR #3426, opened 2026-09-09
[codex] Unify reproduction score comparison rules

### LINKED ISSUE #3425: Centralize reproduction score comparison and unify OK/OKish/FAIL rules
### Problem

The reproduction runners independently classify observed scores as `OK`, `OKish`, or `FAIL`, using different rules. `ReproductionUtils.Constants` centralizes the display labels, but the comparison logic remains in `ReproduceFromPrebuiltIndexes` and `ReproduceFromDocumentCollection`.

### Current behavior

**Prebuilt indexes** apply these checks in order, with `delta = abs(observed - expected)`:

1. Any improvement over expected → `OKish`.
2. `delta < 0.00001` → `OK`.
3. `delta < 0.0002` → `OKish`.
4. Otherwise → `FAIL`.

**Document collections** first round expected and observed scores to the configured metric precision, then apply:

1. Numerical equality (relative tolerance `1e-9`), any improvement, or a flat/HNSW result within configured tolerance → `OK`.
2. A flat/HNSW result within `1.5 ×` configured tolerance → `OKish`.
3. Otherwise → `FAIL`.

Configured tolerance defaults to zero and currently applies only to flat/HNSW models.

Consequently, improvements receive different statuses across workflows, and small regressions can also be classified differently.

### Desired behavior

Centralize classification in `ReproductionUtils`, returning a shared `Status` enum (`OK`, `OKISH`, `FAIL`). Both runners should use the same rules, while retaining responsibility for reading configuration and formatting output.

Let:

- `delta = abs(observed - expected)`
- `tolerance` be the configured tolerance, defaulting to `0` when absent
- `numericalTolerance` be a small, shared allowance for floating-point comparisons

Apply these checks **in order**:

1. **OK:** `delta ≤ max(tolerance, numericalTolerance)`.
2. **OKish:** If not already `OK`, any of:
   - `delta ≤ 1.5 × tolerance`
   - `observed > expected`
   - `delta < 0.0002`
3. **FAIL:** Otherwise.

Absent and explicitly zero tolerances must behave identically. Configured tolerance should apply uniformly, without flat/HNSW-specific classification branches.

Improvements within the `OK` threshold remain `OK`; larger improvements become `OKish`. The `< 0.0002` fallback applies even when a configured tolerance exists.

### Acceptance criteria

- Both runners delegate classification to the shared method.
- The numerical tolerance is explicitly defined and documented.
- Existing metric-precision rounding and output formats are preserved.
- Tests cover equality, floating-point differences, absent/zero tolerance equivalence, improvements, configured tolerance boundaries, the `1.5 ×` boundary, and the strict `< 0.0002` fallback.

lintool:
## Summary

- Centralize score classification in `ReproductionUtils`, returning shared `OK`, `OKISH`, and `FAIL` statuses with a documented absolute numerical tolerance of `1e-9`.
- Apply the shared rules in both runners, support optional per-metric prebuilt tolerances, and apply configured tolerances across all model types while preserving rounding and output formats.
- Add tests for comparison boundaries, improvements, absent/zero tolerances, rounding, and runner output.
- Isolate captured test logs from console output and make numeric YAML fixtures locale-independent, with regression coverage for localized digits.

Fixes #3425.

## Validation

- `mvn -Dtest=ReproduceFromDocumentCollectionTest -Dtests.seed=D5DC109D3569F15D -Dtests.locale=mzn-Arab-IR test` — 18 tests passed; no failures, errors, or skips.
- `git diff --check` — passed.

The full suite previously passed 1,171 tests; it has not been rerun after this locale-fixture update.

### DIFF
--- a/src/main/java/io/anserini/reproduce/ReproductionUtils.java
+++ b/src/main/java/io/anserini/reproduce/ReproductionUtils.java
@@ -46,6 +46,31 @@ public final class ReproductionUtils {
 
   private ReproductionUtils() {}
 
+  public enum Status {
+    OK, OKISH, FAIL
+  }
+
+  /** Absolute allowance for floating-point noise, independent of metric scale or display precision. */
+  public static final double NUMERICAL_TOLERANCE = 1e-9;
+
+  /**
+   * Classifies scores after any metric-specific rounding performed by the caller.
+   * Callers should pass zero when no tolerance is configured.
+   * Differences within the larger of the configured and numerical tolerances are OK.
+   * Otherwise, differences within 1.5 times the configured tolerance, improvements,
+   * and differences strictly below 0.0002 are OKish; all other results fail.
+   */
+  public static Status compareScores(double expected, double observed, double tolerance) {
+    double delta = Math.abs(observed - expected);
+    if (delta <= Math.max(tolerance, NUMERICAL_TOLERANCE)) {
+      return Status.OK;
+    }
+    if (delta <= 1.5 * tolerance || observed > expected || delta < 0.0002) {
+      return Status.OKISH;
+    }
+    return Status.FAIL;
+  }
+
   public static final class Constants {
     // ANSI escape code for red text
     private static final String RED = "\u001B[91m";
--- a/src/main/java/io/anserini/reproduce/ReproductionUtils.java
+++ b/src/main/java/io/anserini/reproduce/ReproductionUtils.java
@@ -36,6 +36,8 @@ import java.util.Locale;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 
+import javax.annotation.Nullable;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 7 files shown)


src/main/java/io/anserini/reproduce/ReproductionUtils.java
```diff
@@ -46,6 +46,31 @@ public final class ReproductionUtils {
...
+   * Differences within the larger of the configured and numerical tolerances are OK.
+   * Otherwise, differences within 1.5 times the configured tolerance, improvements,
+   * and differences strictly below 0.0002 are OKish; all other results fail.
+   */
+  public static Status compareScores(double expected, double observed, double tolerance) {
+    double delta = Math.abs(observed - expected);
+    if (delta <= Math.max(tolerance, NUMERICAL_TOLERANCE)) {
+      return Status.OK;
+    }
+    if (delta <= 1.5 * tolerance || observed > expected || delta < 0.0002) {
```
lilyjge: does it make sense to always have both delta <= 1.5 * tolerance and delta < 0.0002? now there's no 'real' way to have 0 tolerance and have something strictly fail even with a delta of 0.0001

lintool:
Adding more detailed explanations:

1. The tolerance is primarily for HNSW indexes with from-collection indexes, which is non-deterministic each time you build the index.
  - For those, I set a tolerance, and if it's within bounds (based on my observations), program spits out OK.
  - If it's outside the bounds, but not to "badly", the program spits out OKish, and I verify by hand.
2. The `delta < 0.0002` condition is primarily for pre-built indexes, where the differences should be very small - although they are still present due to OS-specific implementations.

So (1) and (2) are usually mutually exclusive - but it still makes sense to centralize the logic.

src/main/java/io/anserini/reproduce/ReproductionUtils.java
lintool: See more detailed explanation below.

> strictly fail even with a delta of 0.0001

Let's say you set the tolerance to 0.0001:
+ OK if delta is within the tolerance.
+ OKish if delta is within 0.00015.

I think I see your concern, which is a edge case... the last branch should be `delta < 0.0002` AND tolerance not set?

lintool:
Updated `ReproductionUtils.compareScores`:

- Annotated `tolerance` with `@Nullable` to make the absent-tolerance contract explicit.
- Rewrote the Javadoc as a bulleted decision tree and documented all parameters and the return value.
- Extracted the fallback threshold into `DEFAULT_OKISH_TOLERANCE = 0.0002`.

Classification behavior is unchanged by this follow-up.

Validation:

- `mvn -Dtest=ReproductionUtilsTest test` — 9 tests passed, no failures or errors.

lilyjge (approved):
I don't love how `score-comparison.yaml` is used but it's fine if it works, the naming of the conditions is not the most intuitive, but it's not a big deal

PR #3428, opened 2026-09-10

tanvirsarao:
Completed the Anserini exercises in the Foundations of Retrieval onboarding path on an Apple Silicon Mac with Java 21.

Successfully reproduced the following items:
- MS MARCO passage collection setup and preprocessing
- BM25 baseline
     ->  MRR@10: 0.1874
- Prebuilt BM25 baseline 
     ->  MRR@10: 0.1875
- BGE-base dense retrieval baseline 
     ->  MRR@10: 0.3521

Some interesting notes/findings:
- It was interesting to see the large (almost 2x!) improvement from BM25 to BGE-base dense retrieval. BM25 relies on lexical term matching, while the dense model can retrieve passages based on learned semantic similarity even when the query and passage use different wording. I found this to be quite interesting and worth noting!
- Working through the collection setup made the retrieval pipeline much more concrete: raw passages are transformed and indexed first, then queries are run against that representation to produce ranked results that can be evaluated against qrels.
- Looking at individual queries alongside MRR helped connect the aggregate metric to actual ranking behavior. A relevant passage at rank 1 contributes a reciprocal rank of 1, while pushing the first relevant result further down the ranking quickly reduces its contribution.


lintool:
I think there should be 3 lessons?

lintool (approved):

